PR #28, opened 2025-03-24
27 inject agent definitions from outside the project

vincent-audibert-whoz:


libs/model/agent-definition.ts
```diff
@@ -16,6 +16,19 @@ export enum ModelSize {
  *   - wrapping a full complex integration (JIRA ?)
  */
 export interface AgentDefinition extends WithDocs {
+  /**
+   * Temporary properties used internally by the AgentService.
+   * These properties are not part of the agent definition schema and should not be included in YAML files.
+   */
+  _tmp?: {
```
dpalita-whoz, line 23: a bit strange, should live in another entity

dpalita-whoz (dismissed):

libs/model/agent-definition.ts
vincent-audibert-whoz: agreed, information kept inside transiently in the `agent.service.ts` for agents creation.

thomasmartin-whoz (approved):

PR #45, opened 2025-04-09
Prompt history

dpalita-whoz:
Add a simple prompt history management (not persisted yet, but it should be trivial to put if in ~/.coday/user/.../prompt_history.txt).
Trigger it with up/down arrows from first/last lines of input to recall your previous or next prompts in the current session.
Also added multiline input management in the terminal (Alt+Enter as in claude code, but we may want to align with the web and have cmd+enter submit...)

### DIFF
--- a/libs/handler/openai.client.ts
+++ b/libs/handler/openai.client.ts
@@ -13,7 +13,7 @@ import { RunSubmitToolOutputsParams } from 'openai/resources/beta/threads/runs/r
 const OpenaiModels = {
   [ModelSize.BIG]: {
     name: 'gpt-4.1-2025-04-14',
-    contextWindow: 128000,
+    contextWindow: 1000000,
     price: {
       inputMTokens: 2,
       cacheRead: 0.5,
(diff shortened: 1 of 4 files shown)


vincent-audibert-whoz:
I had a few issues:
- validation of the prompt input not taken on iterm or regular macos terminal not working with any combination of mod + enter. When playing with it to try to change this behavior, I observe wild results (like shift never detected).
- had some issues with backspace messing with the written text.

It happens to work for me in the IJ terminal but not elsewhere 😞

vincent-audibert-whoz (dismissed):
is ok, let's hope node internals do not change too much 😅

libs/handler/openai.client.ts
```diff
@@ -12,12 +12,12 @@ import { RunSubmitToolOutputsParams } from 'openai/resources/beta/threads/runs/r
 
 const OpenaiModels = {
   [ModelSize.BIG]: {
-    name: 'gpt-4o',
+    name: 'gpt-4.1-2025-04-14',
     contextWindow: 128000,
```
vincent-audibert-whoz, line 16:
```suggestion
    contextWindow: 1000000,
```

vincent-audibert-whoz (approved):

vincent-audibert-whoz (approved):

PR #64, opened 2025-04-28
Fix issue #63: Agent selection with newline

### LINKED ISSUE #63: Agent wrong selection
The agent name parsing does not work if immediately followed by a line return.

vincent-audibert-whoz:
## Description
This PR fixes issue #63 where agent selection fails when the agent name is followed by a newline character instead of a space.

## Changes
- Improved the agent name parsing in `AiHandler.ts` to use a regex that handles any whitespace character (spaces, tabs, newlines)
- Added better error handling when an agent name is specified but not found
- Improved feedback to users about which agent is being used and why
- Added comments to clarify the behavior in different scenarios

## Testing
Tested with the following scenarios:
- `@AgentName some instructions` - Correctly selects the agent
- `@AgentName\nsome instructions` - Now correctly selects the agent (was failing before)
- `@ some instructions` - Correctly uses the default agent selection logic
- `@nonexistentAgent some instructions` - Shows a warning and uses the default agent

Fixes #63

### DIFF
--- a/libs/handler/openai/ai.handler.ts
+++ b/libs/handler/openai/ai.handler.ts
@@ -19,38 +19,51 @@ export class AiHandler extends CommandHandler implements Killable {
   }
 
   async handle(command: string, context: CommandContext): Promise<CommandContext> {
-    /* The command can contain a space after the commandWord that is important to detect if it is an agent name or not.
-     * The getSubCommand of CommandHandler does trim it, misplacing first word as agent name.
-     * ex:
-     *   - "@ foo bar" => no agent name, user prompt is "foo bar"
-     *   - "@foo bar" => agent name is "foo", user prompt is "bar"
+    /* The command should start with @ followed by an agent name, optionally followed by the actual command
+     * Examples:
+     *   - "@" => invalid, no agent name, use default agent
+     *   - "@ " => invalid, no agent name, use default agent
+     *   - "@AgentName" => valid, use AgentName agent with no further command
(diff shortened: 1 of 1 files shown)


vincent-audibert-whoz:
### Technical Details of the Fix

The root cause of issue #63 was that the agent name parsing logic in `AiHandler.ts` was only looking for space characters to delimit the agent name, not handling newlines or other whitespace characters.

#### Before:
```typescript
const firstSpaceIndex = cmd.indexOf(' ')
const potentialAgentName = firstSpaceIndex < 0 ? cmd : cmd.slice(0, firstSpaceIndex)
const restOfCommand = firstSpaceIndex < 0 ? '' : cmd.slice(firstSpaceIndex + 1)
```

This code only checked for a space character (` `), so if a user typed `@AgentName\nsome instructions`, it would try to find an agent named `AgentName\nsome` which obviously doesn't exist.

#### After:
```typescript
const match = command.match(/^@(\S*)(?:\s+(.*))?$/)
const agentName = match[1] // First group: the agent name (could be empty)
const restOfCommand = match[2] || '' // Second group: the rest of the command
```

The new regex-based approach correctly handles any whitespace character (spaces, tabs, newlines) as a delimiter between the agent name and the rest of the command.

I also improved the error handling to provide better feedback when an agent isn't found, and to correctly fall back to the default agent selection logic when needed.

This change makes the agent selection more robust and user-friendly, fixing the issue reported in #63.

vincentpalita-whoz (approved):

PR #123, opened 2025-06-16
feat: separate token info from agent and price

vincent-audibert-whoz:
Token info put at debug level (masked by default).


vincent-audibert-whoz:
Cancel, already merged.

PR #165, opened 2025-09-03
jira-tool: by default all comment are internal. It is based on user t…

charles-monot-whoz:
…oken and if internal note ar enot allowed the comment will be published as standard comment

### DIFF
--- a/libs/integration/jira/add-jira-comment.ts
+++ b/libs/integration/jira/add-jira-comment.ts
@@ -9,7 +9,8 @@ export async function addJiraComment(
   jiraBaseUrl: string,
   jiraApiToken: string,
   jiraUsername: string,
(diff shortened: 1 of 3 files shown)


libs/integration/jira/add-jira-comment.ts
```diff
@@ -9,7 +9,8 @@ export async function addJiraComment(
   jiraBaseUrl: string,
   jiraApiToken: string,
   jiraUsername: string,
-  interactor: Interactor
+  interactor: Interactor,
+  internal: boolean = true
```
charles-monot-whoz, line 13: for safety reason jira comment are internal by default (prevent agent to add comment visible by customers). If the user token doesn't support internal note, the comment will be created anyway without internal scope.

PR #464, opened 2026-02-09
fix: file dates in exchange area

vincent-audibert-whoz:


apps/client/src/app/components/file-exchange-drawer/file-exchange-drawer.component.ts
```diff
@@ -134,23 +128,35 @@ export class FileExchangeDrawerComponent {
   }
 
   /**
-   * Format date in relative time
+   * Format date with day and time (HH:MM:SS)
    */
   formatDate(date: Date): string {
+    // Check if date is valid
+    if (!date || isNaN(date.getTime())) {
```
vincent-palita-whoz: Would be great to be in utils and maybe using date-fns instead of having it tied to a component?

vincent-palita-whoz (approved):

PR #475, opened 2026-02-17
fix: auto-create prompts directory when saving prompt with project source

vincent-audibert-whoz:
- Modified PromptService.getPromptsDir() to create directory if it doesn't exist
- Works for both 'local' and 'project' sources
- Simplified create() method by removing redundant mkdirSync
- Fixes 500 error when creating prompt in project without /prompts/ directory

### DIFF
--- a/libs/service/src/lib/prompt.service.ts
+++ b/libs/service/src/lib/prompt.service.ts
@@ -67,7 +67,7 @@ export class PromptService {
   /**
    * Get prompt file path
    */
-  private async getPromptFilePath(projectName: string, id: string, source: PromptSource): Promise<string> {
+  private async getOrCreatePromptFilePath(projectName: string, id: string, source: PromptSource): Promise<string> {
     const dir = await this.getOrCreatePromptsDir(projectName, source)
     return path.join(dir, `${id}.yml`)
   }
@@ -78,7 +78,7 @@ export class PromptService {
    */
   private async findPromptSource(projectName: string, id: string): Promise<PromptSource | null> {
     // Check local first (most common)
-    const localPath = await this.getPromptFilePath(projectName, id, 'local')
+    const localPath = await this.getOrCreatePromptFilePath(projectName, id, 'local')
     if (existsSync(localPath)) {
       return 'local'
     }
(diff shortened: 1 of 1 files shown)


libs/service/src/lib/prompt.service.ts
```diff
@@ -49,15 +52,23 @@ export class PromptService {
...
+      promptsDir = path.join(this.projectPath, codayFolder, 'prompts')
+    }
+
+    // Create directory if it doesn't exist
+    if (!existsSync(promptsDir)) {
+      mkdirSync(promptsDir, { recursive: true })
+      console.log(`[PROMPT] Created prompts directory: ${promptsDir}`)
     }
+
+    return promptsDir
```
vincent-couturier-whoz: ❤️

la-celton-whoz (approved):

libs/service/src/lib/prompt.service.ts
```diff
@@ -188,9 +199,7 @@ export class PromptService {
         parameterFormat: this.getParameterFormat(prompt.commands), // Compute at creation
       }
 
-      const promptsDir = await this.getPromptsDir(projectName, source)
-      mkdirSync(promptsDir, { recursive: true })
-
+      // getPromptFilePath will call getPromptsDir which creates the directory if needed
```
vincent-couturier-whoz: suggestion: we could improve the code by removing this comment and renaming getPromptFilePath in getPromptFilePathAndCreatePromptsDirIfNeeded

vincent-couturier-whoz (approved):

vincent-audibert-whoz (commented):
@Reviewer: This PR fixes the 500 error when creating prompts in projects without a prompts directory, which is good. However, the solution introduces a design issue where read operations now have directory creation side effects. Methods like findPromptSource and list will create directories just by checking if prompts exist, which violates the principle of least surprise. The naming getOrCreatePromptFilePath is also misleading since that method doesn't create anything itself, only delegates to getOrCreatePromptsDir. Consider either separating read and write concerns (keep getPromptsDir for reads, add ensurePromptsDir for writes) or accept the current behavior but rename to clarify the side effect (e.g., resolvePromptsDir).

Answer: the side-effect of creating the folders if missing is wanted. As we are still dealing with a file-based storage that is a medium-term temporary solution, such details, though important, will be ignored.

PR #519, opened 2026-02-21
Feat/slack two ways

vincent-palita-whoz:


apps/server/src/lib/canal-bridge.ts
```diff
@@ -0,0 +1,198 @@
...
+  getExistingThread(threadId: string, projectName: string, username: string): CanalThreadHandle {
+    return this.ensureInstance(threadId, projectName, username)
+  }
+
+  /**
+   * Send a message into a running Coday thread.
+   * If the thread has an active InviteEvent (waiting for input), responds to it.
+   * Otherwise sends a bare AnswerEvent as a fallback.
+   */
+  sendMessage(threadId: string, message: string): void {
```
vincent-couturier-whoz: Suggestion : it could be simplified by using async paradigm

PR #521, opened 2026-02-24
feat: #520 propagate stop signal and handle interrupted delegations

### LINKED ISSUE #520: Stop button not reactive enough
The stop button is not reactive enoug when:
- agent is outputting text: it should stop right away in the middle of the TextChunkEvents (see implementations of `ai.client.ts`)
- a delegation in in progress, the sub-thread does not receive the stop/kill command (see `delegate.tools.ts`)

vincent-audibert-whoz:
Closes #520

### Stop during text streaming (`anthropic.client.ts`)
In `streamApiCall`, the `stream.on('text', ...)` callback now checks `shouldProceed(thread)` before emitting each chunk. If the thread is stopped, `stream.abort()` is called immediately to cut the Anthropic stream.

OpenAI was already handled correctly via the `for await` loop which checks `shouldProceed(thread)` on every chunk.

### Stop propagation to delegated sub-threads (`delegate.function.ts`)
A 1s interval monitors the parent thread's `runStatus` during delegation. When it switches to `STOPPED`, the forked thread is stopped too, triggering the sub-agent's shutdown via `shouldProceed()`.

The interrupted delegation now returns a clear message to the parent agent (`"Delegation was interrupted before completion. No result available."`) rather than throwing an `EmptyError`, using `forkedThread.runStatus` as the source of truth to distinguish interruption from a normal empty completion.

### Silent stop in `ai.handler.ts`
`lastValueFrom(events)` now uses `{ defaultValue: undefined }` so that a voluntary stop no longer surfaces as `"Could not run agent X : no elements in sequence"`.

### DIFF
--- a/libs/integrations/ai/src/lib/delegate.function.ts
+++ b/libs/integrations/ai/src/lib/delegate.function.ts
@@ -57,14 +57,17 @@ The parent conversation (all previous messages) is there for context, but your c
         result = await lastValueFrom(delegatedEvents, { defaultValue: undefined })
       } finally {
         clearInterval(stopPropagationInterval)
+        // Always restore stackDepth to avoid corrupting the depth guard on exception
+        context.stackDepth++
       }
-
-      context.stackDepth++
-      context.aiThread!.merge(forkedThread)
 
       if (forkedThread.runStatus === RunStatus.STOPPED) {
         return 'Delegation was interrupted before completion. No result available.'
       }
+
+      // Only merge the forked thread if delegation completed successfully
+      context.aiThread!.merge(forkedThread)
+
       if (!result) {
         return 'Delegation completed but produced no message.'
       }


libs/integrations/ai/src/lib/delegate.function.ts
```diff
@@ -43,11 +43,31 @@ The parent conversation (all previous messages) is there for context, but your c
...
+        clearInterval(stopPropagationInterval)
+      }
 
       context.stackDepth++
       context.aiThread!.merge(forkedThread)
+
+      if (forkedThread.runStatus === RunStatus.STOPPED) {
+        return 'Delegation was interrupted before completion. No result available.'
+      }
+      if (!result) {
```
vincent-audibert-whoz: **Bug : merge du thread interrompu avant le check de stop**

`context.stackDepth++` et `context.aiThread!.merge(forkedThread)` sont exécutés inconditionnellement, avant le check `forkedThread.runStatus === RunStatus.STOPPED`. En cas d'interruption, un état partiel du sous-thread est donc mergé dans le thread parent. C'est probablement indésirable — le merge devrait être conditionnel au succès de la délégation, ou au minimum placé après les guards de runStatus.

romain-petit-whoz (approved):

Issue #444, opened 2026-01-23
feat: Generic HTTP integration with OAuth2 support


vincent-audibert-whoz:
## Context

Need to integrate PSO (behind Google SSO) without creating a custom integration. This opens the door for declarative HTTP integrations for internal/custom APIs.

### 1. Integration Schema Definition

Enable declarative HTTP integration in `coday.yaml`:

```yaml
integration:
  PSO:
    type: http  # New type
    baseUrl: "https://pso.whoz.com/api"
    auth:
      type: oauth2
      provider: google-sso
      # OR for simple auth:
      # type: credentials
      # username: "${env:PSO_USER}"
      # apiKey: "${env:PSO_TOKEN}"
    
    endpoints:
      - name: searchConcepts
        method: POST
        path: "/search"
        params:
          - name: term
            type: string
            required: true
          - name: languages
            type: array
            required: false
          - name: conceptType
            type: string
            required: false
        description: "Search for concepts by term"
      
      - name: getConcept
        method: GET
        path: "/concepts/{conceptId}"
        params:
          - name: conceptId
            type: string
            required: true
            location: path
        description: "Get concept details by ID"
      
      - name: getConceptChildren
        method: GET
        path: "/hierarchy/children/{conceptId}"
        params:
          - name: conceptId
            type: string
            required: true
            location: path
```

### 2. Runtime Tool Generation

Generate tools dynamically from endpoint definitions:
- Tool name: `{integrationName}.{endpointName}` (e.g., `PSO.searchConcepts`)
- Parameters from schema
- Auto-generate tool descriptions from endpoint metadata

### 3. Authentication Support

**Phase 1: Simple credentials** (like existing integrations)
- Username + API key in headers
- Stored in user config

**Phase 2: OAuth2** (like Basecamp integration)
- Token storage in user config
- Token refresh flow
- Support standard OAuth2 + Google SSO variant

### Intermediate Step (Recommended)

Create hybrid integration pattern:
```typescript
// libs/integration/pso.tools.ts
export class PsoIntegration extends GenericHttpIntegration {
  constructor(config: HttpIntegrationConfig) {
    super(config)
    // Override specific behavior if needed
  }
}
```

Allows iteration on schema/interfaces without full commitment to final design.

### Existing OAuth2 Reference

Basecamp integration already has non-standard OAuth2 implementation with token storage in user config. Reuse patterns where applicable.

## Scope

**In scope:**
- Schema definition for HTTP endpoints
[...]

PR #526, opened 2026-02-25
feat: #444 generic HTTP integration with OAuth2 and declarative endpoints

vincent-audibert-whoz:
Relates to #444

## What

New `@coday/integrations-http` lib implementing a generic HTTP integration with OAuth2 support. Tools are generated dynamically from declarative endpoint configuration — no code required for new integrations.

## How it works

Configuration in `coday.yaml` (or user config as JSON):

```yaml
integration:
  MY_CALENDAR:
    type: HTTP
    http:
      baseUrl: "https://www.googleapis.com/calendar/v3"
      endpoints:
        - name: getEvents
          method: GET
          path: "/calendars/{calendarId}/events"
          description: "List events from a Google Calendar"
          responseFormat: yaml       # optional, ~30% token savings
          ignorePaths:               # optional response filtering
            - "items.*.conferenceData"
            - "items.*.htmlLink"
            - "items.*.etag"
          params:
            - name: calendarId
              type: string
              required: true
              location: path
            - name: timeMin
              type: string
              location: query
    oauth2:
      client_id: "..."
      client_secret: "..."
      redirect_uri: "http://localhost:3001/oauth/callback"
      authorization_endpoint: "https://accounts.google.com/o/oauth2/v2/auth"
      token_endpoint: "https://oauth2.googleapis.com/token"
      scope:
        - "https://www.googleapis.com/auth/calendar.readonly"
```

Each endpoint becomes one AI tool named `${integrationName}__${endpointName}`.

## Key design decisions

- **OAuth credentials in `oauth2.*`** (not `username`/`apiKey` like legacy integrations)
- **`scope` accepts `string | string[]`** — joined with spaces per OAuth2 spec
- **`http.baseUrl`** in a dedicated `http` block in `IntegrationConfig`
- **Response filtering** via `keepPaths` / `ignorePaths` (dot-notation, `*` wildcard) — `keepPaths` takes precedence
- **`responseFormat: yaml`** for more compact LLM output
- **`access_type=offline` + `prompt=consent`** in OAuth flow to always get a `refresh_token`
- Token refresh is automatic; full re-auth triggered only if no `refresh_token` available

## Files

- `libs/integrations/http/` — new lib (`generic-oauth.ts`, `http.tools.ts`)
- `libs/model/src/lib/integration-config.ts` — added `HttpConfig`, `HttpEndpointConfig`, `HttpParamConfig`, `scope: string | string[]`
- `libs/agent/src/lib/toolbox.ts` — registered `HTTP` type in factory registry
- `tsconfig.base.json` — added `@coday/integrations-http` path alias

## What's not in scope (follow-up)

- Configuration UI for HTTP integrations
- Non-OAuth auth (API key in header, bearer token)
- Response transformation beyond filtering

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
+++ b/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
@@ -9,10 +9,10 @@
         @if (item.kind === 'message') {
           <div
             class="case-chat__message"
-            [class.case-chat__message--user]="item.event.actor?.role === 'USER'"
-            [class.case-chat__message--agent]="item.event.actor?.role === 'AGENT'"
+            [class.case-chat__message--user]="item.event.actor.role === 'USER'"
+            [class.case-chat__message--agent]="item.event.actor.role === 'AGENT'"
           >
-            <span class="case-chat__message-author">{{ item.event.actor?.displayName }}</span>
+            <span class="case-chat__message-author">{{ item.event.actor.displayName }}</span>
             <p class="case-chat__message-text">{{ extractText(item.event) }}</p>
           </div>
         } @else {


libs/integrations/http/src/lib/generic-oauth.ts
```diff
@@ -0,0 +1,325 @@
...
+        )
+      }
+    }
+    return this.tokenData!.accessToken
+  }
+
+  /**
+   * Initiates the OAuth2 Authorization Code + PKCE flow.
+   * Emits an OAuthRequestEvent for the frontend to open the auth URL.
+   * Returns a Promise resolved by handleCallback() when the user completes auth.
```
vincent-audibert-whoz: **Design concern: No concurrent-call guard**

```typescript
async authenticate(): Promise<TokenData> {
  // no check for already-pending flow
  this.pendingState = state
  this.pendingCodeVerifier = codeVerifier
  this.pendingResolve = resolve
  this.pendingReject = reject
```

If `authenticate()` is called a second time while a flow is already in progress (user triggers the tool twice before completing the browser auth), all pending fields are overwritten. The first flow's Promise leaks permanently; only the second flow will ever resolve.

libs/integrations/http/src/lib/http.tools.ts
```diff
@@ -0,0 +1,329 @@
...
+export class HttpTools extends AssistantToolFactory {
+  static readonly TYPE = 'HTTP' as const
+
+  private oauth: GenericOAuth | null = null
+
+  constructor(
+    interactor: Interactor,
+    private readonly userService: UserService,
+    instanceName: string,
+    config?: IntegrationConfig
```
vincent-audibert-whoz: **Bug: `GenericOAuth` instance replaced on every `buildTools()` call**

```typescript
this.oauth = new GenericOAuth(...)
```

`GenericOAuth` caches the access token in `this.tokenData`. Every `buildTools()` invocation discards any previously acquired in-memory token state. If `buildTools()` is called more than once per session (agent re-init, context refresh), the user is forced through the full OAuth flow again even though a valid token may exist in storage. Tokens in storage survive, but any in-flight pending flow is silently dropped.

PR #541, opened 2026-03-09
feat: WZ-28667: single responsability for case runtime

thomasmartin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/Case.kt
```diff
@@ -1,401 +1,24 @@
...
-            logger.info { "[Case $id] AgentSelectedEvent emitted for $defaultAgentName" }
-        } else {
-            logger.error { "[Case $id] No default agent configured, stopping case" }
-            stopRequested.set(true)
-        }
-    }
-
-    companion object : KLogging()
-}
+    val status: CaseStatus = CaseStatus.PENDING,
```
vincent-couturier-whoz: Suggestion: name or titre should be useful to be able to what the case deals with.

Issue #567, opened 2026-03-16
feat: invite a user to a thread — sharing UI and API


vincent-audibert-whoz:
## Goal

Allow thread owners to explicitly invite other users to a thread. Any user in the thread's 
user list has full access to participate in the conversation.

> Access granularity (read vs write) is out of scope for now — being in the list means full access.

## Intended changes

**Backend**
- Expand the current PUT REST endpoint to update a thread with its users list
- Validate that the invited username exists before adding

**Frontend**
- Add a sharing panel or button on the thread detail view (visible to the thread owner)
- Input to enter a username to invite with autocomplete, with confirmation
- Display the current participants list with option to remove each
- Shared threads are visually distinct in the thread list (e.g. a small indicator showing it is shared)

## Expected outcome

- A thread owner can add a specific colleague to the thread's user list
- The invited user sees the thread in their thread list and has full access to it
- The owner can remove a user from the list at any time
- Non-owners cannot manage participants

## Notes

- Requires #563 (access model) — the `users` field must exist before this story can be implemented
- A separate CRUD REST to manage users (admin scope) is deferred to a later issue

PR #578, opened 2026-03-17
feat: #567 thread sharing UI and API

charles-monot-whoz:


apps/client/src/app/components/thread-share/thread-share.component.scss
```diff
@@ -0,0 +1,129 @@
...
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+}
+
+.username-field {
+  flex: 1;
+
+  // Reduce default Material form field height
+  ::ng-deep .mat-mdc-form-field-infix {
```
charles-monot-whoz: avoid usage of ng-deep

PR #601, opened 2026-03-19
Feat/599 free message posting

charles-monot-whoz:


libs/core/src/lib/core.ts
```diff
@@ -61,6 +65,54 @@ export class Coday {
       this.services.project,
       this.services.logger
     )
+
+    // Track pending invite events so addUserMessage can answer them.
+    // Skip events during replay to avoid stale historical events polluting state.
+    this.interactor.events
+      .pipe(
+        filter((e) => e instanceof InviteEvent),
+        filter(() => !this.isReplaying)
```
mathieu-dearmey: could be combined in single filter

mathieu-dearmey (approved):

libs/core/src/lib/core.ts
```diff
@@ -61,6 +65,54 @@ export class Coday {
...
+      this.interactor.sendEvent(answerEvent)
+    } else {
+      // Agent is running — queue the message for the next initCommand() iteration.
+      // Emit a MessageEvent immediately so the UI shows the message without delay.
+      const messageEvent = new MessageEvent({
+        role: 'user',
+        content: [{ type: 'text', content: message }],
+        name: username,
+      })
+      this.interactor.sendEvent(messageEvent)
```
cetautomatix-coday: It does not generate duplicated messages when the user message is indeed handled by the agent ?

Issue #585, opened 2026-03-17
Agents: define a way of working with worktrees


vincent-audibert-whoz:
What agent should create a worktree ?
=> PM to initiate the work on an issue / story / bug ?

How an agent gets information on the progress of a "worktree" (meaning the worktree's purpose) ?
=> THREAD integration accesses also other worktree threads ? Other projects (through configuration of THREAD integration) ?

When does an agent removes a worktree ?
=> PM when the issue/PR is seen closed ? Needs a specific prompt to list worktrees and check on github the status.

How an agent runs a task on a worktree ?
Through a distinct tool, part of the create tool ?
Need to connect to the PromptExecutionService to trigger that on a "distinct" project.
How is that really distinct from the delegation (ie should delegation cover the cross-project tasks) ? Should delegation be able to be sync (currently) or async (as with webhooks without the awaitFinalAnswer) ?

PR #611, opened 2026-03-21
feat: #585 async delegation support and worktree setup fix

vincent-audibert-whoz:
Easy support of async delegation initiation.

Small worktree setup fix.


cetautomatix-coday:
## Review — feat: #585 async delegation support and worktree setup fix

The PR title mentions a "worktree setup fix" but the diff contains no changes to `git-worktree.tools.ts` or `project.service.ts` — those files are identical to master. The actual scope is limited to async delegation in `delegate.function.ts` and the matching schema addition in `delegate.tools.ts`. The PR description and title should be corrected to reflect the actual scope.

**Async delegation logic** has several correctness concerns worth addressing.

### `stackDepth` not decremented for async delegations

In the new code, `asyncDelegations` are launched without touching `context.stackDepth`, while `syncDelegations` correctly bracket with `--`/`++`. This means an agent can fire unlimited async delegations regardless of the configured depth limit. If the intent is that async delegations are truly unconstrained (fire-and-forget), this should be documented explicitly with a comment explaining the design decision; otherwise it's a silent guard bypass.

### `subThread.id` mutation after `fork()`

In `setupAndRunAsync`:

```typescript
subThread = parentThread.fork(agentName, task, parentEventId)
subThread.id = delegation.threadId  // direct mutation after construction
```

`fork()` creates the thread with `crypto.randomUUID()` and sets internal state tied to that ID. Mutating `.id` after the fact means the `AiThread` instance carries a different ID than the one used in its construction context. Since `AiThread` is not designed to have its `id` changed post-construction (it's a public mutable field, but nothing in the class contract says it's safe to reassign), this is fragile. The proper fix is to pass the pre-generated ID into `fork()` or to have `fork()` accept an optional ID parameter.

### Stop propagation interval leaks on agent-not-found path (async)

In `setupAndRunAsync`, the `stopPropagationInterval` is started **before** the `agentFind` call but the early-return on agent-not-found path (`if (!agent) { return }`) does not call `clearInterval`. The interval will leak for the lifetime of the process.

```typescript
// stopPropagationInterval created here
const agent = await agentFind(agentName, context)
if (!agent) {
  interactor.error(...)
  return  // ← interval never cleared
}
```

The same pattern exists in `runSingleDelegation` in master and is reproduced in the new function — both paths should be fixed.

### `crypto.randomUUID()` called without import guard

`crypto.randomUUID()` is used in `setupAndRunAsync` without an explicit import (`import * as crypto from 'node:crypto'` or similar). This relies on the global `crypto` object (available in Node 19+ and browsers). If the runtime is Node 18 (which the project supports per `package.json`), `globalThis.crypto.randomUUID` is available but using the explicit `node:crypto` import (already used in `project.service.ts`) is more consistent and explicit. Not a blocker but worth aligning.

### Branch naming deviation

The branch `feat/delegation-async-threadid` does not follow the required convention `feature/username/issue-XXXX-...` from `DEV_WORKFLOW.md`. Minor, but worth noting for consistency.

Issue #566, opened 2026-03-16
feat: per-request user context — agent runs under the identity of the requesting user


vincent-audibert-whoz:
## Goal

Decouple the agent execution context from the thread owner's identity so that each agent run uses the credentials, configuration, and memories of the user who sent the triggering message.

## Context

Today a `ThreadCodayInstance` is created with a fixed `username` (the thread owner). The `AiClientProvider`, `UserService`, and agent instructions are all initialized from that single identity and never change. In a multi-user thread, a message from User B should run the agent with User B's API keys and bio, not User A's.

The chosen rule: **the agent hierarchy runs with the credentials of whoever last injected input**. If User B sends a message and the agent asks a question that User C answers, the follow-up runs under User C's context.

## Intended changes

- Carry the sender's username on each incoming `AnswerEvent` (set by the server from the authenticated request)
- `ThreadCodayInstance` tracks an `activeUsername` that is updated on each incoming message before the agent run begins
- `AiClientProvider` resolves AI provider credentials against `activeUsername` at call time rather than at construction time — it becomes user-context-aware rather than user-bound
- Agent instructions (bio, memories) are composed using the active user's `UserService` context at the moment the agent is invoked, not at instance creation
- The agent cache in `AgentService` is keyed by `(agentName, username)` or invalidated on user switch, to prevent one user's agent state leaking into another's run
- Usage logging and cost attribution use `activeUsername` so costs are correctly assigned to the user who triggered the run

## Expected outcome

- In single-user threads, behavior is identical to today
- In multi-user threads, each user's messages run agents with their own API keys, bio, and memory context
- Cost attribution in logs correctly reflects the triggering user
- The thread owner's credentials are never used for another user's agent run

## Notes

- Requires #563 (access model) as the structural foundation
- This is the deepest architectural change in the multi-user epic; all other stories are simpler once this is in place
- `AiClientProvider` moving from user-bound to user-context-aware is the key design decision

PR #613, opened 2026-03-22
feat: #566 per request user context

vincent-audibert-whoz:


libs/handlers/openai/src/lib/ai.handler.ts
```diff
@@ -34,6 +43,10 @@ export class AiHandler extends CommandHandler implements Killable {
       return context
     }
 
+    const selectedEvent = new AgentSelectedEvent({ agentName: selectedAgent.name })
+    context.aiThread?.addOrchestrationEvent(selectedEvent)
+    this.interactor.sendEvent(selectedEvent)
```
vincent-audibert-whoz: Check if `interactor.sendEvent` is not doubling the calls and event sent.

libs/model/src/lib/ai-thread.ts
```diff
@@ -419,6 +425,14 @@ export class AiThread {
     this.modifiedDate = new Date().toISOString()
   }
 
+  /**
+   * Add an orchestration event to the thread history.
+   * Used for agent lifecycle tracking (selected, running, finished).
+   */
+  addOrchestrationEvent(event: AgentSelectedEvent | AgentRunningEvent | AgentFinishedEvent): void {
```
vincent-audibert-whoz: Method a bit too specific, no other more generic method ?

PR #618, opened 2026-03-23
fix: #561 add inline OAuth authorization panel to avoid popup blocking

### LINKED ISSUE #561: bugfix: OAuth popup blocked by Chrome — add inline auth panel in conversation flow
## Problem

The current OAuth flow calls `window.open()` programmatically (triggered by an SSE event, not a direct user gesture). Chrome blocks this by default, and its UX for the blocked popup is broken:

> "Clicking 'Always allow' makes the notification disappear and the popup never opens. The only recovery was to restart Coday."

The iframe alternative is not viable: OAuth providers (Google, Basecamp...) explicitly forbid being loaded in iframes via `X-Frame-Options: DENY` / `Content-Security-Policy: frame-ancestors 'none'` — a deliberate anti-clickjacking measure.

**Root cause:** `window.open()` in `oauth.service.ts` is triggered by an SSE event, not a direct user click. Browsers only allow popups opened from a direct user gesture.

**Affected code:**
- `apps/client/src/app/core/services/oauth.service.ts` — calls `window.open()` on SSE event, no fallback when `popup` is `null` (blocked)
- `apps/client/src/app/components/oauth-callback/oauth-callback.component.ts` — relies on `window.opener.postMessage()`, works fine once the popup actually opens

## Proposed Solution

**Add an inline OAuth authorization panel in the conversation flow**, following the existing `choice-select` pattern (which already replaces the textarea when user input is needed).

When an `OAuthRequestEvent` is received:
1. Display an `oauth-request` panel in the input area (similar to `choice-select`) showing the integration name and an "Authorize" button
2. The user clicks the button → `window.open()` is called **from a direct user gesture** → no browser blocking
3. The existing popup + `postMessage` flow continues unchanged

**Why this works:**
- Browsers only block popups triggered programmatically. A popup opened from a `click` event handler is always allowed.
- No change to the `OAuthCallbackComponent` or the token exchange logic
- Follows the established pattern of the project (the `choice-select` component already does exactly this kind of "interrupt the conversation to get user input")

## Implementation sketch

```
thread.component.html
  @if (pendingOAuthRequest) {
    <app-oauth-request-panel
      [integrationName]="pendingOAuthRequest.integrationName"
      [authUrl]="pendingOAuthRequest.authUrl"
      (authorized)="onOAuthAuthorized()"
      (cancelled)="onOAuthCancelled()"
    />
  }
```

The `oauth.service.ts` no longer calls `window.open()` itself — it emits the event to the thread, which shows the panel. The panel's button click triggers the popup.

## Acceptance Criteria

- [ ] OAuth flow works end-to-end without requiring popup permissions in Chrome (popup opened from user click)
- [ ] Inline panel is displayed in the conversation input area when OAuth is required, consistent with the `choice-select` UX pattern
- [ ] Clear cancel option available in the panel
- [ ] No change to the callback flow (`OAuthCallbackComponent` / `postMessage` mechanism)
- [ ] Existing Basecamp + Google Calendar integrations tested and working

## Related

- #271 — original OAuth2 implementation issue
- #444 — generic HTTP integration with OAuth2 support (will benefit from this fix)

cetautomatix-coday:
## Problem

Chrome blocks `window.open()` calls triggered programmatically (from an SSE event, not a direct user gesture). The previous implementation in `oauth.service.ts` called `window.open()` directly upon receiving an `OAuthRequestEvent`, causing the popup to be silently blocked.

## Solution

Add an inline OAuth authorization panel in the conversation input area, following the existing `choice-select` pattern.

When an `OAuthRequestEvent` is received:
1. The service stores the pending request and exposes it via `pendingRequest$` (BehaviorSubject)
2. An inline panel appears in the input area showing the integration name and an "Authorize" button
3. The user clicks the button → `window.open()` is called **from a direct user gesture** → no browser blocking
4. The existing popup + `postMessage` flow continues unchanged

## Changes

- **`oauth.service.ts`** — no longer calls `window.open()` on SSE event; exposes `pendingRequest$` observable; adds `openPopup()` and `cancelRequest()` public methods
- **`oauth-request-panel` component** (new) — compact panel with integration name, "Authorize" and "Cancel" buttons
- **`thread.component`** — integrates the panel in the input area alongside `choice-select`

## Acceptance criteria

- [x] OAuth flow works without requiring popup permissions in Chrome
- [x] Inline panel displayed in the conversation input area, consistent with `choice-select` UX
- [x] Cancel option available
- [x] No change to `OAuthCallbackComponent` / `postMessage` mechanism

Closes #561

### DIFF
--- a/apps/client/src/app/components/thread/thread.component.ts
+++ b/apps/client/src/app/components/thread/thread.component.ts
@@ -1,9 +1,11 @@
 import {
   Component,
+  EventEmitter,
   Input,
   OnInit,
   OnDestroy,
   OnChanges,
+  Output,
   SimpleChanges,
   AfterViewChecked,
   ElementRef,
@@ -78,6 +80,10 @@ export class ThreadComponent implements OnInit, OnDestroy, OnChanges, AfterViewC
   // Required inputs - thread identification
   @Input({ required: true }) projectName!: string
   @Input({ required: true }) threadId!: string
+
+  /** Forwarded from ChatTextareaComponent — used to close sidenav on mobile. */
+  @Output() chatFocused = new EventEmitter<void>()
+  @Output() drawerStateChange = new EventEmitter<boolean>()
 
   private readonly destroy$ = new Subject<void>()
   private readonly connectionDestroy$ = new Subject<void>()
@@ -363,16 +369,23 @@ export class ThreadComponent implements OnInit, OnDestroy, OnChanges, AfterViewC
   }
 
   // Drawer toggle/close methods
+  private emitDrawerState(): void {
+    this.drawerStateChange.emit(this.isDrawerOpen)
+  }
+
   toggleFileDrawer(): void {
     this.drawerMode = this.drawerMode === 'files' ? 'none' : 'files'
+    this.emitDrawerState()
   }
 
   toggleSharePanel(): void {
     this.drawerMode = this.drawerMode === 'share' ? 'none' : 'share'
+    this.emitDrawerState()
   }
 
   closeDrawer(): void {
     this.drawerMode = 'none'
+    this.emitDrawerState()
   }
 
   onUserAdded(userId: string): void {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,32 @@
+## 0.103.0 (2026-03-24)
+
+### 🚀 Features
+
+- #636 support Cloudflare Access JWT (CF_Authorization) for user identity in --auth mode ([#638](https://github.com/whoz-oss/coday/pull/638), [#636](https://github.com/whoz-oss/coday/issues/636))
+
+### ❤️ Thank You
+
+- cetautomatix-coday
+- m1
+
+## 0.102.1 (2026-03-24)
+
+### 🚀 Features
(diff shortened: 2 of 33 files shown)


apps/client/src/app/components/thread/thread.component.ts
```diff
@@ -124,6 +127,7 @@ export class ThreadComponent implements OnInit, OnDestroy, OnChanges, AfterViewC
 
   // Modern Angular dependency injection
   private readonly codayService = inject(CodayService)
+  readonly oauthService = inject(OAuthService)
```
cetautomatix-coday: **Service exposed as non-private for template access**

`oauthService` is `readonly` but not `private`, making it part of the component's public API solely to allow `oauthService.pendingRequest$ | async` in the template. The architecture convention is to encapsulate service references and expose only what the template needs as a component-level property:

```typescript
private readonly oauthService = inject(OAuthService)
readonly pendingOAuthRequest = toSignal(this.oauthService.pendingRequest$, { initialValue: null })
```

Then in the template: `@if (pendingOAuthRequest())` — no service reference leaking out.

PR #727, opened 2026-04-03
fix: slack config not persisted to coday.yaml on startup or re-setup

vincent-palita-whoz:


cetautomatix-coday:
## Review

Single file changed: `apps/desktop-twin/src/main.ts`. Two distinct changes: a reworked `applySlackConfigToVault` function and a new call to it at every startup inside `initialize`. The bug fix rationale is sound — Slack config saved after vault creation was never applied to `coday.yaml` because the original substitution only happened during vault initialization. The startup re-application pattern mirrors the existing Google Calendar pattern.

Several issues to address before merge.

---

### `applySlackConfigToVault` — fallback logic creates data corruption risk

```typescript
// Fallback: full replace if the curl context scope found nothing
// (handles edge cases where the template format differs)
content = content.split(candidate).join(config.webhookUrl)
changed = true
log('INFO', `Replaced webhook URL (was "${candidate}") in coday.yaml (full replace fallback)`)
break
```

The fallback at the end of the `replaceInContext` block runs unconditionally when `replaceInContext` returns the unchanged string — but the condition is `updated !== content`, so the fallback only fires when the context-scoped replace produced no change. The problem: the fallback does a **global** `split/join` on a value that could appear legitimately elsewhere in the YAML (e.g. in a comment, a description, or another tool's command). If the webhook URL happens to appear in an unrelated line, it will be replaced there too. The comment says "handles edge cases where the template format differs" but a global replace is not a safe fallback — it trades one bug (missed replacement) for a potential silent data corruption.

Additionally, this fallback is unreachable in practice when `valueInFile` is null (no curl command found): in that case `candidates` contains only `previousConfig.webhookUrl` and/or `SLACK_NOTIFY_WEBHOOK_URL`, and `replaceInContext` will correctly find and replace the placeholder on the curl line if it exists. The fallback therefore only activates for a hypothetical template format change, and the global replace is the wrong remedy.

---

### `replaceInContext` — wrong separator for multi-occurrence lines

```typescript
return line.split(oldValue).join(newValue)
```

`String.prototype.split(str).join(str)` is a manual `replaceAll`. This is fine for correctness, but if `oldValue` appears more than once on the same line (e.g. both in a comment and in the URL), all occurrences are replaced. For the webhook URL this is unlikely but worth noting — the existing `replaceAll` calls elsewhere in the file have the same behavior and it was apparently intentional, so this is a minor consistency note rather than a blocker.

---

### `initialize` — startup re-application passes empty `previousConfig`

```typescript
const slackConfig = getStoredSlackConfig()
if (slackConfig.enabled && (slackConfig.webhookUrl || slackConfig.userId)) {
  applySlackConfigToVault(slackConfig, { enabled: false, webhookUrl: '', userId: '' })
}
```

`previousConfig` is passed as empty. The function's candidate list is `[valueInFile, previousConfig.webhookUrl, 'SLACK_NOTIFY_WEBHOOK_URL'].filter(Boolean)`. With an empty `previousConfig`, only `valueInFile` (discovered from the file) and the placeholder string remain as candidates. This means:

- **First startup after vault creation**: `valueInFile` will be the already-substituted URL (from `initializeVault`), which equals `config.webhookUrl`, so `candidate !== config.webhookUrl` is false and no replacement happens. Correct behavior.
- **First startup after vault existed before Slack was configured**: `valueInFile` will be the placeholder string `SLACK_NOTIFY_WEBHOOK_URL`, which will be found and replaced. Correct.
- **After a URL change in Preferences**: the old URL is `valueInFile`, `candidate !== config.webhookUrl` is true, and it gets replaced. Correct.

The empty `previousConfig` is therefore safe for the startup case, but the JSDoc comment says "previousConfig is used as a hint for the old value" — there is a discrepancy between the documented contract and the actual behavior. The function now discovers the current value from the file itself, making `previousConfig` a secondary fallback. The comment should be updated to reflect the actual resolution order, or the parameter should be removed from the startup call signature to make intent explicit.

---

### Semantic issue — `slackConfig.enabled` gate in `initialize` is too restrictive

```typescript
if (slackConfig.enabled && (slackConfig.webhookUrl || slackConfig.userId)) {
  applySlackConfigToVault(...)
}
```

If a user had Slack enabled, configured the URL, then disabled Slack (without clearing the URL), the re-application on startup is skipped. The placeholder would remain in `coday.yaml` until Slack is re-enabled and the app restarts. Whether this is intentional is unclear from the code — the preferences `saveSlackConfig` handler does not gate on `enabled` before calling `applySlackConfigToVault`. Inconsistency: the preferences save applies regardless of `enabled`, but the startup re-application requires `enabled: true`.

PR #738, opened 2026-04-08
fix: #WZ-31310 and #WZ-31311 user fixes in logs and local mode

cetautomatix-coday:
## Problem

Case audit data exposed the user's email address in two places:

1. **`CaseController.addMessage`** — `Actor` was built with `displayName = user.email`, so the email ended up stored in every `MessageEvent` and visible via the audit API.
2. **`CaseRuntime.addUserMessage`** — the log line used `actor.displayName`, producing entries like `actor: marcel.chombier@whoz.com`.

## Fix

**`CaseController`** — `displayName` is now composed from `firstname + lastname` when available, falling back to `user.metadata.id.toString()` (internal UUID). No PII in this field.

**`CaseRuntime`** — log now uses `actor.id` (internal UUID) instead of `actor.displayName`.

## Scope

`AgentSimple`/`AgentAdvanced` also reference `actor.displayName` but only for agent actors (not users), so there is no PII concern there.

Note: `AgentServiceImpl.buildInstructions` intentionally injects `user.email` into the LLM system prompt so the agent knows who it is talking to — this is a separate product-level decision and is out of scope for this fix.

## Testing

All existing tests pass (`nx test agentos-service`).


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -56,13 +56,38 @@ class UserController(
...
+    override fun update(
+        id: UUID,
+        resource: UserResource,
+    ): UserResource {
+        val existing =
+            userService.findById(id)
+                ?: throw org.springframework.web.server.ResponseStatusException(
+                    org.springframework.http.HttpStatus.NOT_FOUND,
+                    "Entity not found: $id",
+                )
```
thomasmartin-whoz: tu peux utiliser ResourceNotFoundException qui est une exception que j'ai ajoutée au projet.
De manière générale je préfère des exceptions "fonctionnelles" plutôt que de définir le status code à chaque fois

thomasmartin-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
vincent-audibert-whoz: changé aussi sur les autres controllers, #743 pour les autres status.

thomasmartin-whoz (approved):

PR #744, opened 2026-04-10
feat: issue global mission control

benjamin-valdes-whoz:
### Backend — In-memory Registry

ThreadCodayManager: pendingInvites = new Set() with setPendingInvite(), clearPendingInvite(), hasPendingInvite() — synchronous source of truth, zero race condition

broadcastEvent(): sets the registry when a real InviteEvent/ChoiceEvent is emitted, with an isReplaying flag to prevent spurious re-SETs during SSE replay

message.routes.ts: clears the registry + persists the AnswerEvent in the thread (fixes the bug where the invite was considered unanswered on page reload)

replayThreadHistory(): consults the registry before replaying the invite

GET /api/threads (global): enriched with pendingInvite (matching the per-project endpoint)

interactor.ts: clears lastInviteEvent after receiving the answer

### Frontend — Status derivation

MissionStatusService: deriveStatus() now relies solely on thread.pendingInvite from the backend, no longer on the global hasInvite state from CodayService (which was stale after navigation)

GlobalMissionService: removal of the liveStatuses cache that was preserving waiting-you indefinitely; deriveStatus() now uses thread.pendingInvite

GlobalMissionControlComponent: refreshes on ThreadUpdateEvent received via project SSE, to retrieve the up-to-date status after a CLEAR


apps/server/src/lib/project.routes.ts
```diff
@@ -23,8 +27,19 @@ import { getParamAsString } from './route-helpers'
...
+ * @param projectEventManager
+ * @param threadService
+ * @param getUsernameFn
  */
-export function registerProjectRoutes(app: express.Application, projectService: ProjectService): void {
+export function registerProjectRoutes(
+  app: express.Application,
+  projectService: ProjectService,
+  projectEventManager?: ProjectEventManager,
+  threadService?: import('@coday/service').ThreadService,
```
vincent-audibert-whoz: Weird syntax ?

vincent-audibert-whoz (dismissed):
You did some bug fixes on the way I think 👏 

Hard to review, LGTM and tested ok locally.

apps/client/src/app/components/confirm-dialog/confirm-dialog.component.html
```diff
@@ -0,0 +1,19 @@
+<h2 mat-dialog-title>
+  <mat-icon [class.destructive]="data.isDestructive !== false">warning</mat-icon>
```
leopsola: !!data.isDestructive non?

benjamin-valdes-whoz: bien vu

apps/server/src/lib/project.routes.ts
benjamin-valdes-whoz: effectivement

vincent-audibert-whoz (approved):

PR #745, opened 2026-04-10
feat: wz-28819: Graph relationships added for Case, CaseEvent, Namespace and Integration

thomasmartin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/neo4j/CaseEventNode.kt
```diff
@@ -26,7 +36,10 @@ sealed class CaseEventNode(
     open val modified: Instant = Instant.now(),
     open val modifiedBy: String? = null,
     open val removed: Boolean? = null,
-)
+) {
+    @Relationship(type = "BELONGS_TO", direction = OUTGOING)
+    var case: CaseNode? = null
```
vincent-couturier-whoz, line 41: question: it seems like the same information as the caseId

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/neo4j/CaseEventNodeNeo4jRepository.kt
```diff
@@ -8,13 +8,34 @@ import org.springframework.data.neo4j.repository.query.Query
...
+     *
+     * Returns `e, r, c` so SDN maps the [CaseEventNode.case] @Relationship field.
      */
     @Query(
-        "MATCH (e:CaseEvent) " +
-            "WHERE e.caseId = \$caseId AND (e.removed IS NULL OR e.removed = false) " +
-            "RETURN e ORDER BY e.timestamp ASC, e.id ASC",
+        $$"""MATCH (e:CaseEvent)
+            WHERE e.caseId = $caseId AND (e.removed IS NULL OR e.removed = false)
+            OPTIONAL MATCH (e)-[r:BELONGS_TO]->(c:Case)
```
vincent-couturier-whoz, line 18: suggestion: optional match is strange here perhaps we should avoid relationship if we have a caseId property. We should have either caseId or relationship with case and a condition on the case but not both

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/neo4j/CaseEventNode.kt
vincent-couturier-whoz, line 31: thought: I realize that caseId is perhaps better than having case property with relationship managed by spring data. Otherwise we will get caseNode retrieved for all caseEvent node retrieved in the database. I don't know how we could use a relationship and have at the same time no direct instance of caseNode in a caseEventNode....

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/neo4j/Neo4jCaseRepository.kt
```diff
@@ -21,6 +21,7 @@ open class Neo4jCaseRepository(
     override fun save(entity: Case): Case =
         caseNodeNeo4jRepository
             .save(CaseNode.fromDomain(entity))
+            .also { caseNodeNeo4jRepository.linkCaseToNamespace(it.id, entity.namespaceId.toString()) }
```
vincent-couturier-whoz, line 24: question: why we have to manage the link explicitly ?

vincent-couturier-whoz:
I will do some tests to understand the memory and processing impact. Not totally sure of the complexity and issues that can be brought by explicitly defining the relationship in the object model

vincent-couturier-whoz (approved):

vincent-audibert-whoz (approved):

PR #746, opened 2026-04-10
feat: #31209 namespace-level LLM configuration (LlmConfig + LlmModelConfig entities)

cetautomatix-coday:
## Summary

Implements namespace-level LLM configuration as described in [WZ-31209](https://whoz.atlassian.net/browse/WZ-31209).

Introduces two new entities — `AiProvider` and `AiModel` — scoped to a namespace, with full CRUD REST endpoints, Neo4j persistence, and agent-side resolution logic. Includes the corresponding frontend pages/forms and a regenerated OpenAPI spec + TypeScript client.

---

### Backend (`agentos-service`)

- **`AiProvider`** entity: holds provider credentials (API key masked on read), base URL, and namespace scoping. CRUD at `/api/ai-providers`.
- **`AiModel`** entity: holds model-specific parameters (temperature, max tokens, priority, aliases). Scoped to a namespace via `namespaceId`/`userId`. CRUD at `/api/ai-models`, with a `by-namespaceId` list endpoint.
- **`AiModelService.findModelConfig`**: resolves the best model config for an agent by alias (first matching `"default"` alias, then the requested name) then falls back to `apiName`, breaking ties by priority. This replaces the previous scattered `findDefaultModelConfig` logic.
- **`AgentServiceImpl`**: emits resolution logs so operator can trace which LLM config was selected.
- **Neo4j persistence** for both entities.
- **OpenAPI spec** regenerated.

### Frontend (`agentos-api-client` + Angular app)

- `AiProvider` and `AiModel` list pages and create/edit forms.
- `priority` field displayed in `AiModel` item (replacing temp/maxTokens).
- Frontend aligned with regenerated API client model.
- Namespace item kebab-menu consolidated (`ds-kebab-menu`).

---

## Key design decisions

- **`AiModel` is a first-class entity** (not embedded in `AiProvider`) so it can be queried, paginated, and managed independently.
- **Alias-based resolution** (`findModel`) allows agents to reference a logical name (e.g. `"default"`, `"fast"`) rather than a provider-specific model name, decoupling agent definitions from provider details.
- **Priority breaks ties within same-name matches only** — it does not override an exact alias match.
- **API key is masked** on read (last 4 chars visible) to avoid leaking credentials through the REST API.

---

## Testing

- Existing agent instantiation flow unaffected (non-regression).
- Namespace data isolation preserved.
- Manual verification of CRUD endpoints and model resolution logic via logs.


cetautomatix-coday:
## Code Review — feat: #31209 namespace-level LLM configuration

The overall structure is solid and follows established patterns. No security blockers. There are several correctness and quality issues to address before merge.

---

### 1. Security — `isMasked` is too permissive

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmConfig/ApiKeyMasking.kt`

```kotlin
fun isMasked(value: String?): Boolean = value != null && MASK in value
```

The sentinel `"****"` is a substring check. Any real API key that happens to contain four consecutive asterisks (e.g. a password-style key) would be incorrectly treated as masked, causing the controller to silently preserve the *old* key even when the caller intended to update it. The check should verify the full masked format (prefix + `****` + suffix, or the exact `"****"` constant) rather than a plain `contains`. This is a correctness issue with a security implication: a legitimate key update would be silently dropped.

---

### 2. Logic — `resolveAgentName` in `AgentServiceImpl` duplicates `findModelConfig`

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt`

```kotlin
override fun resolveAgentName(namePart: String, namespaceId: UUID): String? {
    val candidates = llmModelConfigService.findByNamespaceId(namespaceId)
    val match =
        candidates.filter { it.alias.equals(namePart, ignoreCase = true) }.maxByOrNull { it.priority }
            ?: candidates.filter { it.apiName.equals(namePart, ignoreCase = true) }.maxByOrNull { it.priority }
    return match?.let { it.alias ?: it.apiName }
}
```

This is an in-lined copy of `LlmModelConfigServiceImpl.findModelConfig`, fetching all namespace candidates twice (once here, once inside `resolveModelPair` when called from `findAgentByName`). It should delegate to `llmModelConfigService.findModelConfig(namespaceId, namePart)` to keep the resolution logic in one place. As written, a future change to the resolution algorithm in the service would not be reflected here.

Similarly, `findDefaultModelConfig` is a one-liner wrapper around `llmModelConfigService.findModelConfig(namespaceId)` — that's fine, but `getDefaultAgent` and `getDefaultAgentName` both call it and then immediately call `llmConfigService.getById`. Consider unifying them.

---

### 3. Logic — `LlmModelConfigServiceImpl.create` uniqueness checks miss the update path

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmModelConfig/LlmModelConfigServiceImpl.kt`

The uniqueness constraints on `(llmConfigId, apiName)` and `(llmConfigId, alias)` are enforced only in `create`. The `update` path goes directly to `repository.save(entity)` with no validation. A caller can therefore use `PUT /api/llm-model-configs/{id}` to assign an `apiName` or `alias` that is already taken by a sibling, breaking the invariant silently. The same applies to `LlmConfigServiceImpl.update` for the `(namespaceId, userId, name)` uniqueness constraint.

---

### 4. Logic — `LlmModelConfigServiceImpl.findByLlmConfigAndAlias` / `findByLlmConfigAndApiName` — N+1 on every uniqueness check

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmModelConfig/LlmModelConfigServiceImpl.kt`

```kotlin
override fun findByLlmConfigAndApiName(llmConfigId: UUID, apiName: String): LlmModelConfig? =
    repository.findByParent(llmConfigId).firstOrNull { it.apiName == apiName }

override fun findByLlmConfigAndAlias(llmConfigId: UUID, alias: String): LlmModelConfig? =
    repository.findByParent(llmConfigId).firstOrNull { it.alias == alias }
```

Both methods load **all** model configs for a provider just to filter in memory. In the Neo4j implementation this is a full scan of the `LlmModelConfig` nodes for a given `llmConfigId`. For the uniqueness checks in `create` this means two full scans per creation request. The Neo4j repository should expose dedicated `findByLlmConfigIdAndApiName` / `findByLlmConfigIdAndAlias` Cypher queries to avoid this. This is a correctness concern for production scale, not just a performance note.

---

### 5. Logic — `LlmConfigServiceImpl.findByNamespaceAndUserAndName` — unnecessary double scan

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmConfig/LlmConfigServiceImpl.kt`

```kotlin
val candidates = when {
    namespaceId != null -> repository.findByNamespaceId(namespaceId)
    userId != null -> repository.findByUserId(userId)
    else -> emptyList()
}
return candidates.firstOrNull { it.namespaceId == namespaceId && it.userId == userId && it.name == name }
```

When `namespaceId != null`, this loads *all* configs for that namespace (potentially from multiple users) and then re-filters by `userId` and `name` in memory. For the conflict check in `create` this is acceptable, but the logic is fragile: if a namespace has many configs scoped to different users, the in-memory filter is the only guard against returning a wrong result. This should be a single indexed query at the repository level.

---

### 6. Neo4j — `Neo4jLlmConfigRepository.deleteByParent` only deletes by namespace, not by userId

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/neo4j/Neo4jLlmConfigRepository.kt`

```kotlin
override fun deleteByParent(parentId: UUID): Int {
    val active = neo4jRepository.findActiveByNamespaceId(parentId.toString())
    ...
}
```

`deleteByParent` is documented as "by convention delegates to findByNamespaceId", but this means calling `deleteByParent` with a `userId` would silently delete nothing (or worse, delete configs from the wrong namespace if the UUID happens to collide). The asymmetry between `findByParent → findByNamespaceId` and the existence of a separate `findByUserId` path is a latent bug: any code that calls `deleteByParent` expecting user-scoped deletion would silently fail. At minimum this should throw `UnsupportedOperationException` for the ambiguous case, or the interface should be split.

---

### 7. Kotlin style — `when` not used for two-branch conditions

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmConfig/LlmConfigController.kt`

```kotlin
val resolvedApiKey = when {
    isMasked(resource.apiKey) -> existing.apiKey
    else -> resource.apiKey
}
```

This specific instance is fine (it uses `when`). However, in `AgentServiceImpl.buildInstructions`:

```kotlin
val base = if (baseInstructions.isNullOrBlank()) namespaceBlock else "$baseInstructions\n$namespaceBlock"
return if (userBlock != null) "$base\n$userBlock" else base
```

Two chained `if/else` expressions should be `when` per the code style guide (`agentos/docs/code-style.md`: "Prefer `when` over chained `if`/`else if`"). The second early-return-style `return if (...)` also violates "avoid early returns in the middle of a method".

---

### 8. Angular — `NamespaceLlmConfigsComponent` and `LlmModelConfigFormComponent` call API services directly

**Files:**
- `libs/agentos-ui/src/lib/components/namespace-llm-configs/namespace-llm-configs.component.ts`
- `libs/agentos-ui/src/lib/components/namespace-llm-models/` (by analogy)
- `libs/agentos-ui/src/lib/components/llm-config-form/llm-config-form.component.ts`
- `libs/agentos-ui/src/lib/components/llm-model-config-form/llm-model-config-form.component.ts`

All four components inject and call `LlmConfigControllerService` / `LlmModelConfigControllerService` directly. Per `apps/client/docs/angular-guidelines.md` and `apps/client/docs/architecture.md`, components must never call API services directly — all HTTP calls must go through a state service. This pattern is consistently enforced for the Coday frontend and is explicitly required for the AgentOS frontend as well ("Components inject only state services. Never API services directly."). The existing `IntegrationConfig` pages follow this rule; these new pages do not.

---

### 9. Angular — `NamespaceLlmConfigsComponent` missing `ChangeDetectionStrategy.OnPush`

**File:** `libs/agentos-ui/src/lib/components/namespace-llm-configs/namespace-llm-configs.component.ts`

The component is presentational (no mutable state, no form) and subscribes to an Observable via `AsyncPipe`. It should declare `changeDetection: ChangeDetectionStrategy.OnPush` per the guidelines. The sibling `LlmConfigItemComponent` and `LlmModelConfigItemComponent` correctly use `OnPush`.

---

### 10. Angular — `LlmModelConfigFormComponent` uses `listByParentLlmConfig` as a `toSignal` source without error handling

**File:** `libs/agentos-ui/src/lib/components/llm-model-config-form/llm-model-config-form.component.ts`

```typescript
protected readonly providers = toSignal(this.llmConfigController.listByParentLlmConfig(this.namespaceId), {
  initialValue: [] as LlmConfig[],
})
```

`toSignal` called with an Observable that can error will propagate the error as an unhandled exception into the signal's context. There is no `catchError` here. If the providers endpoint fails (network error, 404, etc.), the form will break silently. This should use `catchError(() => of([]))` in the pipe.

---

### 11. Minor — `toDomain` uses a nil-UUID placeholder that could leak

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmModelConfig/LlmModelConfigController.kt`

```kotlin
namespaceId = resource.namespaceId ?: UUID.fromString("00000000-0000-0000-0000-000000000000"),
```

The comment explains that `LlmModelConfigServiceImpl.create()` overwrites this before persisting. However, if `update()` is called on an existing entity that somehow has a null `namespaceId` in the resource (e.g. a client that omits the field), the nil-UUID placeholder would be persisted, breaking namespace-scoped queries. The placeholder pattern is fragile; a better approach is to make `namespaceId` non-nullable in `LlmModelConfig` (it is always resolved from the parent at creation time) or to throw explicitly if it is null at the controller boundary for updates.

cetautomatix-coday:
Addressed the backend review findings in commit c4b52ebe:

**#2 — Uniqueness enforced on `update`**
Both `LlmConfigServiceImpl.update` and `LlmModelConfigServiceImpl.update` now run the same conflict checks as `create`, skipping the entity being updated (self-conflict guard via `it.id != entity.id`). Tests added for: no self-conflict on same name, conflict on rename to an existing sibling, same for alias.

**#3 — `resolveAgentName` deduplicated**
The copy-pasted resolution logic is gone. `resolveAgentName` now delegates to `llmModelConfigService.findModelConfig(namespaceId, namePart)` — a one-liner. Tests updated to stub `findModelConfig` directly.

**#5 — Uniqueness checks pushed to the repository layer**
`LlmModelConfigRepository` now declares `findByLlmConfigAndApiName` and `findByLlmConfigAndAlias`. `InMemoryLlmModelConfigRepository` implements them with an in-memory filter on the already-scoped `findByParent` result. `LlmModelConfigNodeNeo4jRepository` has two new `@Query`-annotated methods with targeted Cypher (`WHERE … AND m.apiName = $apiName LIMIT 1` / `… m.alias = $alias LIMIT 1`) — no full namespace scan. `Neo4jLlmModelConfigRepository` delegates to those. `LlmModelConfigServiceImpl.findByLlmConfigAnd*` now calls the repository methods directly.

**#9 — `buildInstructions` uses `when`**
The two-branch `if/else` chain is replaced with a `when` expression covering all four combinations of `baseInstructions` / `userBlock` nullability.

Frontend items (#6, #7, #8) are tracked separately with Frontay.

cetautomatix-coday:
## Follow-up Review — current HEAD (0c0de32)

Most issues from the first review are resolved. The items below are what remains open, plus two new findings introduced by the frontend corrections.

---

### Still open from previous review

#### 1. `isMasked` is still a plain substring check — not fixed

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmConfig/ApiKeyMasking.kt`

```kotlin
fun isMasked(value: String?): Boolean = value != null && MASK in value
```

This is byte-for-byte identical to the first review. The fix was not applied. A real API key that contains `****` (e.g. a generated password-style key) will be treated as masked and its update silently dropped. The check should verify the full masked format produced by `maskApiKey` — e.g. match against the exact patterns `"****"`, `"??****??"`, `"????****????"` — rather than a substring search.

---

#### 2. `resolveModelPair` still duplicates `findModelConfig` logic — not fixed

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt`

```kotlin
private fun resolveModelPair(name: String, namespaceId: UUID): Pair<LlmModelConfig, LlmConfig> {
    val candidates = llmModelConfigService.findByNamespaceId(namespaceId)
    val aliasMatch = candidates.filter { it.alias.equals(name, ignoreCase = true) }.maxByOrNull { it.priority }
    val modelConfig = aliasMatch
        ?: candidates.filter { it.apiName.equals(name, ignoreCase = true) }.maxByOrNull { it.priority }
        ?: throw IllegalArgumentException(...)
    ...
}
```

`resolveAgentName` now correctly delegates to `llmModelConfigService.findModelConfig(...)`, but `resolveModelPair` — which is called by the hot path `findAgentByName` — still fetches all namespace candidates and re-implements the alias-then-apiName priority logic in-line. This duplicates `LlmModelConfigServiceImpl.findModelConfig` exactly. The method should call `llmModelConfigService.findModelConfig(namespaceId, name) ?: throw IllegalArgumentException(...)` and then load the provider config, eliminating the duplication and ensuring both code paths stay in sync.

---

#### 3. `LlmModelConfigController.toDomain` nil-UUID placeholder still present — not fixed

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmModelConfig/LlmModelConfigController.kt`

```kotlin
namespaceId = resource.namespaceId ?: UUID.fromString("00000000-0000-0000-0000-000000000000"),
```

The concern from the first review stands: `update` calls `toDomain` on the incoming resource, which may have `namespaceId = null` if a client omits it (e.g. a generated client that doesn't send read-only fields). The nil-UUID would then be persisted, breaking all namespace-scoped queries for that record. In `LlmModelConfigServiceImpl.update` the `namespaceId` is not re-resolved from the parent `LlmConfig` the way `create` does, so there is no server-side correction on the update path. Either (a) `update` should re-resolve `namespaceId`/`userId` from the existing persisted entity before calling `repository.save`, or (b) the placeholder should be replaced by an explicit `requireNotNull` that fails fast with a clear 400.

---

#### 4. `findByNamespaceAndUserAndName` double-scan still present — not fixed

**File:** `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/llmConfig/LlmConfigServiceImpl.kt`

```kotlin
val candidates = when {
    namespaceId != null -> repository.findByNamespaceId(namespaceId)
    userId != null -> repository.findByUserId(userId)
    else -> emptyList()
}
return candidates.firstOrNull { it.namespaceId == namespaceId && it.userId == userId && it.name == name }
```

The first call loads all configs for the namespace (across all users), then filters in memory. The same observation as the first review: this is acceptable for now but is the correct place for a dedicated repository query. Flagging as a carry-over since it was not addressed.

---

### New findings introduced by the frontend corrections

#### 5. Angular — form components still call API services directly

**Files:**
- `libs/agentos-ui/src/lib/components/llm-config-form/llm-config-form.component.ts`
- `libs/agentos-ui/src/lib/components/llm-model-config-form/llm-model-config-form.component.ts`
- `libs/agentos-ui/src/lib/components/namespace-llm-configs/namespace-llm-configs.component.ts`
- `libs/agentos-ui/src/lib/components/namespace-llm-models/namespace-llm-models.component.ts`

All four components still inject and call `LlmConfigControllerService` / `LlmModelConfigControllerService` directly. The architectural violation from the first review was reported as fixed, but the current HEAD is identical to what was reviewed the first time. For example, `LlmConfigFormComponent` still has:

```typescript
private readonly llmConfigController = inject(LlmConfigControllerService)
...
this.llmConfigController.getByIdLlmConfig(id).pipe(...).subscribe(...)
this.llmConfigController.updateLlmConfig(...).pipe(...).subscribe(...)
this.llmConfigController.createLlmConfig(payload).pipe(...).subscribe(...)
```

The `angular-guidelines.md` and `architecture.md` are unambiguous: components must inject state services only; HTTP calls must be mediated by a state service (`*-state.service.ts`). The existing `IntegrationConfig` pages follow this rule. The new LLM pages do not. No `LlmConfigStateService` or `LlmModelConfigStateService` exists in `libs/agentos-ui/src/lib/services/` — the directory still contains only `user-state.service.ts`.

---

#### 6. Angular — `@for` track uses a method call instead of an expression

**File:** `libs/agentos-ui/src/lib/components/llm-model-config-form/llm-model-config-form.component.html`

```html
@for (provider of providers(); track trackByProvider($index, provider)) {
```

The `track` expression in Angular's `@for` block must be a **property access expression** on the loop variable, not a method call. `track trackByProvider($index, provider)` is a function invocation — Angular evaluates it on every change detection cycle, defeating the purpose of `track` and potentially causing unnecessary re-renders. Per the guidelines the correct form is `track provider.id` directly.

---

#### 7. Angular — `NamespaceLlmConfigsComponent` and `NamespaceLlmModelsComponent` subscribe in constructor with side effects

**Files:**
- `libs/agentos-ui/src/lib/components/namespace-llm-configs/namespace-llm-configs.component.ts`
- `libs/agentos-ui/src/lib/components/namespace-llm-models/namespace-llm-models.component.ts`

Both components build a side-effectful `configsById` / `modelsById` map by subscribing inside the constructor:

```typescript
constructor() {
  this.configs$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((configs) => {
    this.configsById = new Map(configs.map((c) => [c.id ?? '', c]))
  })
}
```

This is a mutable non-signal field updated via a subscription, used to resolve item-template events (`resolveConfig(id)`). With `ChangeDetectionStrategy.OnPush` the map update does not trigger change detection, so `resolveConfig` may return stale data on the first render if the `configs$` emission races with the template rendering. The map should be a `signal<Map<string, LlmConfig>>` updated via `toSignal` or an explicit `effect`, so that the template reads from a reactive source.

cetautomatix-coday:
Round 2 backend findings addressed in commit 2b052567:

**#2 — `resolveModelPair` deduplicated**
`resolveModelPair` (the hot path behind `findAgentByName`) now delegates entirely to `llmModelConfigService.findModelConfig(namespaceId, name)`. The in-line alias→apiName→priority logic is gone. All `AgentServiceImplSpec` tests updated to stub `findModelConfig` instead of `findByNamespaceId`.

**#3 — Server-owned fields preserved on update**
`LlmModelConfigController` now overrides `update` to fetch the persisted entity first, then merges only client-mutable fields (`apiName`, `alias`, `priority`, `temperature`, `maxTokens`) onto it via `toDomainForUpdate`. `namespaceId`, `userId`, and `llmConfigId` always come from the persisted record — the client cannot overwrite them. `toDomain` (used by `create`) now sets `namespaceId = null` explicitly since the service always resolves it from the parent `LlmConfig` anyway. Test added to `LlmModelConfigControllerSpec` verifying that a client sending `namespaceId = null` still results in the correct server-side value being persisted.

Frontend items (#5, #6, #7) remain with Frontay.

agentos/aimodel/gpt-5-mini.yaml
```diff
@@ -1,9 +1,5 @@
-name: gpt-5-mini
-description: OpenAI GPT-5-mini general assistant
-modelName: gpt-5-mini
+alias: gpt-5-mini
+apiName: gpt-5-mini
```
vincent-couturier-whoz, line 2: suggestion: description could be useful for administrative purpose

agentos/aimodel/claude-haiku.yaml
```diff
@@ -1,8 +1,5 @@
-name: claude-haiku
-description: Anthropic Claude Haiku for quick chat and tool use
-modelName: claude-haiku-4-5
+alias: claude-haiku
+apiName: claude-haiku-4-5
```
vincent-couturier-whoz, line 2: suggestion: apiModelName could be more meaningful

agentos/aimodel/gpt-5-mini.yaml
vincent-audibert-whoz: ok

agentos/aimodel/claude-haiku.yaml
vincent-audibert-whoz: renamed

agentos/agentos-plugins-filesystem/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/aiprovider/AiProviderYamlModel.kt
```diff
@@ -2,11 +2,8 @@ package io.whozoss.agentos.plugin.filesystem.aiprovider
...
-    val apiType: String,
-    val baseUrl: String,
     val description: String? = null,
-    val defaultApiKey: String? = null,
-    val baseModel: String? = null,
-    val temperature: Double?,
-    val maxTokens: Int? = null,
+    val apiType: String,
+    val baseUrl: String? = null,
+    val apiKey: String? = null,
```
vincent-couturier-whoz, line 8: question : why did temperature an maxTokens disappear ? it would be interesting to have the ability to defined "default" value and perhaps we could keep them and add a default prefix to precise the intention

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/aiProvider/AiModel.kt
```diff
@@ -1,29 +1,47 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
 data class AiModel(
     override val metadata: EntityMetadata = EntityMetadata(),
-    val name: String,
-    val description: String,
-    val modelName: String,
-    val providerName: String,
+    val aiProviderId: UUID,
+    val namespaceId: UUID? = null,
+    val userId: UUID? = null,
```
vincent-couturier-whoz, line 40: suggestion : we could name this field ownerId because it would precise the meaning of this field

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -46,58 +54,98 @@ class AgentServiceImpl(
...
+    override fun getDefaultAgentName(namespaceId: UUID): String? {
+        val modelConfig = findDefaultModelConfig(namespaceId) ?: return null
+        return modelConfig.alias ?: modelConfig.apiModelName
     }
 
-    override fun getDefaultAgentName(): String? = aiModelRegistry.getDefault()?.name
+    override fun resolveAgentName(
+        namePart: String,
+        namespaceId: UUID,
+    ): String? = aiModelService.findAiModel(namespaceId, namePart)?.let { it.alias ?: it.apiModelName }
```
vincent-couturier-whoz, line 78: issue : I don't understand the relationship between the name "resolveAgentName" and the fact that the function return an aiModel using the name of an agent. Especially since the function is in the AgentServiceImpl class. It's coherent with the change on the agent definition but for me we want to decouple the definition of the agent from the definition of a model so we should have an aliasName and so we should get the agent and then get the model related to the modelAlias defined on the agent

agentos/agentos-plugins-filesystem/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/aiprovider/FilesystemAIProviderProvider.kt
```diff
@@ -68,16 +70,22 @@ class FilesystemAIProviderProvider : AiProviderPlugin {
...
     }
 
-    companion object : KLogging()
+    companion object : KLogging() {
+        /**
+         * Derives a stable UUID from a provider name so that AiModel entries loaded
+         * by [FilesystemAiModelProvider] can reference the same provider across restarts
+         * without a persistent store.
+         */
+        fun deterministicId(providerName: String): UUID =
```
vincent-couturier-whoz, line 88: question : why we need this mecanism as the definition is stored on filesystem ? logically the id could be stored also in the file.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -40,13 +39,12 @@ import java.util.UUID
  */
 class AgentAdvanced(
     override val metadata: EntityMetadata = EntityMetadata(),
-    private val model: AiModel,
+    override val name: String,
```
vincent-couturier-whoz, line 42: question : it seems that the name of the agent will become the name of the model. I'm surprised because I was thinking we would define a modelAlias on the agent. Why this decision ?

agentos/agentos-plugins-filesystem/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/aiprovider/AiProviderYamlModel.kt
vincent-audibert-whoz: Temperature could be useful at provider level, especially given the OpenAI tendency to expect `1.0` for some models and nothing else otherwise 4XX...

But the maxTokens make no sense to me to define on the aiProviders, given every model has its own context window limit, that can vary wildly between models of a given provider.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/aiProvider/AiModel.kt
vincent-audibert-whoz: Makes sense, will have to update ~20 files for that 🥲

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
vincent-audibert-whoz: Agents aren’t really there, here we have hardwired code to bypass this concept that does not exist yet. This is the topic of WZ-31326 about bringing the agent entity and ability to re-use the agents defined for Coday.

agentos/agentos-plugins-filesystem/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/aiprovider/FilesystemAIProviderProvider.kt
vincent-audibert-whoz: True, we had a discussion about it with @thomasmartin-whoz: was made that way for the sake of simplicity at the moment.
We need to think about the useability though, as a human I can invent a string name right away, not really a uuid 😆 . This would require a automatic "define & add the ID if missing" on that entities.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
vincent-audibert-whoz: Hardwiring pending agent concept real arrival.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/aiProvider/AiModel.kt
vincent-audibert-whoz: In fact, I’ll leave it that way because we will probably blow up this property when attributing these configuration elements to the various levels (namespace, user(s), group(s)).

agentos/agentos-plugins-filesystem/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/aiprovider/FilesystemAIProviderProvider.kt
vincent-audibert-whoz: And this topic will be left to the Agent story to explore how to reconcile files with ids in them.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelNode.kt
```diff
@@ -0,0 +1,83 @@
...
+ * [namespaceId] and [userId] are denormalised from the parent [io.whozoss.agentos.aiProvider.AiProviderNode] at
+ * creation time so that namespace-scoped queries can be served with a single
+ * WHERE clause without graph traversal.
+ */
+@Node("AiModel")
+data class AiModelNode(
+    @Id
+    val id: String,
+    val aiProviderId: String,
+    val namespaceId: String? = null,
```
vincent-couturier-whoz, line 26: suggestion: we should add the link with namespace but we could do that in another PR

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun findActiveByAiProviderIdAndApiName(
+        aiProviderId: String,
+        apiName: String,
+    ): AiModelNode?
+
+    /**
+     * Find the first non-removed model config under [aiProviderId] whose alias matches exactly.
+     */
+    @Query(
+        "MATCH (m:AiModel) " +
```
vincent-couturier-whoz, line 49: suggestion: use `$$""""   """"` to be able to copy past easily the request to neo4j if you have to test it

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/InMemoryAiModelRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import io.whozoss.agentos.entity.InMemoryEntityRepository
+import io.whozoss.agentos.sdk.aiProvider.AiModel
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
+import org.springframework.stereotype.Repository
+import java.util.UUID
+
+@Repository
+@ConditionalOnExpression(
+    "'\${agentos.persistence.mode:in-memory}' != 'neo4j' " +
```
vincent-couturier-whoz, line 11: suggestion: `$$""" """`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/Neo4JAiModelRepository.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.transaction.annotation.Transactional
+import java.util.UUID
+
+/**
+ * Neo4j-backed implementation of [AiModelRepository].
+ *
+ * Parent type is [java.util.UUID] representing the aiProviderId.
+ */
+open class Neo4JAiModelRepository(
```
vincent-couturier-whoz, line 14: question : why open is added here ? if it's spring constraint we could add the spring open plugin to avoid the add in the class and bad surprise of unwanted inheritence

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderServiceImpl.kt
```diff
@@ -0,0 +1,76 @@
...
+ * masked [AiProvider.apiKey] values before calling this method.
+ */
+@Service
+class AiProviderServiceImpl(
+    private val repository: AiProviderRepository,
+) : AiProviderService {
+    override fun create(entity: AiProvider): AiProvider {
+        // Domain invariant — also checked in the entity init block, but we surface a
+        // proper HTTP 400 here rather than letting an IllegalArgumentException bubble.
+        if (entity.namespaceId == null && entity.userId == null) {
```
vincent-couturier-whoz, line 24: suggestion: we should use require or check to do this test

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/Neo4JAiProviderRepository.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.util.UUID
+
+/**
+ * Neo4j-backed implementation of [AiProviderRepository].
+ *
+ * [findByParent] delegates to [findByNamespaceId] by convention — namespace is the
+ * primary scope for this ticket. [findByUserId] will be the primary path once
+ * user-scoped configs are introduced (WZ-31210).
+ */
+open class Neo4JAiProviderRepository(
```
vincent-couturier-whoz, line 16: suggestion: Neo4jAiProviderRepository instead of Neo4JAiProviderRepository

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
```diff
@@ -46,7 +52,13 @@ import org.springframework.data.neo4j.repository.config.EnableNeo4jRepositories
     "'\${agentos.persistence.mode:in-memory}' == 'neo4j' " +
         "or '\${agentos.persistence.mode:in-memory}' == 'embedded-neo4j'",
 )
-@EnableNeo4jRepositories(basePackages = ["io.whozoss.agentos.persistence.neo4j"])
+@EnableNeo4jRepositories(
+    basePackages = [
+        "io.whozoss.agentos.persistence.neo4j",
```
vincent-couturier-whoz, line 57: suggestion: it's seems not coherent to have part of the repository in io.whozoss.agentos.persistence.neo4j and another one in io.whozoss.agentos.aiModel and io.whozoss.agentos.aiProvider. We should have all neo4j repository in persistence.neo4j or split all in its dedicated package. I also prefer the a vertical split but it's not our initial choice so we should move the repositories to io.whozoss.agentos.persistence.neo4j

agentos/aimodel/claude-haiku.yaml
vincent-couturier-whoz: I don't see the difference... it's still apiName here

vincent-couturier-whoz (approved):

PR #753, opened 2026-04-15
fix: #752 use chooseOption for tool confirmation prompt instead of promptText

### LINKED ISSUE #752: fix: tool confirmation prompt should use ChoiceEvent instead of InviteEvent in web UI
## Problem

When a tool programmatically requires user confirmation (e.g. dangerous git commands like `push`, `reset`, `clean` — see `libs/integrations/git/src/lib/git.ts`), the backend currently uses `interactor.promptText()` which emits an **`InviteEvent`**.

In the web frontend, this renders as a plain textarea prompt with an **inverted and confusing UX**:
- Empty input → **accept** (execute the command)
- Typed text → **reject** (text is used as the rejection reason)

This pattern is error-prone and non-obvious, especially in a chat interface where the user has no visual cue that this is a critical decision point.

## Expected Behaviour

Tool confirmation prompts should use **`interactor.chooseOption()`** (which emits a `ChoiceEvent`) instead of `promptText()`.

The frontend already handles `ChoiceEvent` with an explicit choice UI. The proposed UX:

1. Present **two clear options**: e.g. `"OK"` and `"Cancel"`
2. Also allow **free-text input** for the user to provide a rejection reason
3. Any response that is not the `"OK"` option is treated as a **refusal** (with the typed text used as the reason if provided)

This maps cleanly onto the existing `ChoiceEvent` infrastructure and avoids introducing new components.

## Current Implementation

**Backend** (`libs/function/src/lib/run-bash.ts`):
```typescript
if (requireConfirmation) {
  const rejectReason = await interactor.promptText(
    `\nPlease type the reason to reject this command (nothing = validate)`
  )
  if (rejectReason) {
    return `Command execution was cancelled by user with following reason: ${rejectReason}`
  }
}
```

**Event flow:**
- `promptText()` → emits `InviteEvent` → frontend renders as plain textarea placeholder
- `chooseOption()` → emits `ChoiceEvent` → frontend renders explicit option buttons ✅

## Proposed Change

Replace `interactor.promptText(...)` in the confirmation path with `interactor.chooseOption(...)`, passing:
- A descriptive message showing what action is about to be executed
- Two options: `"OK"` (confirm) and `"Cancel"` (reject)
- Treat any response other than `"OK"` as a rejection; use the response text as the reason if it's not one of the predefined options

**Backend change (sketch):**
```typescript
if (requireConfirmation) {
  const answer = await interactor.chooseOption(
    `Confirm execution of: \`${command}\``,
    ['OK', 'Cancel']
  )
  if (answer !== 'OK') {
    return `Command execution was cancelled by user${answer && answer !== 'Cancel' ? ` with reason: ${answer}` : ''}`
  }
}
```

## Scope

- `libs/function/src/lib/run-bash.ts` — replace `promptText` with `chooseOption` in the confirmation path
- `libs/integration/src/lib/project-scripts-tools.ts` — same if it uses the same pattern
- Frontend: verify `ChoiceEvent` rendering supports free-text fallback alongside predefined options; add if missing

## Acceptance Criteria

- [ ] Dangerous tool operations (git push, reset, etc.) display an explicit **OK / Cancel** choice in the web UI
- [ ] User can also type a custom rejection reason instead of selecting "Cancel"
- [ ] Any non-"OK" response cancels the operation; typed text is forwarded as the reason
- [ ] Terminal interface remains unaffected (or degrades gracefully)
- [ ] No new event types or components introduced — reuses existing `ChoiceEvent` infrastructure

cetautomatix-coday:
## Summary

Fixes #752

Replaces `interactor.promptText()` with `interactor.chooseOption()` in the confirmation path of `run-bash.ts`.

## Changes

- `libs/function/src/lib/run-bash.ts` — confirmation now emits a `ChoiceEvent` with `OK` / `Cancel` options and `allowFreeText: true`, so the user can also type a custom rejection reason.

## Behaviour

- **OK** → command executes
- **Cancel** → command cancelled (no reason)
- **Custom text** → command cancelled with the typed reason forwarded
- Terminal interface unaffected (graceful degradation via existing `chooseOption` implementation)
- No new event types or components introduced — reuses existing `ChoiceEvent` infrastructure

### DIFF
--- a/libs/function/src/lib/run-bash.ts
+++ b/libs/function/src/lib/run-bash.ts
@@ -46,11 +46,15 @@ export const runBash = async ({
 
     // If confirmation is required, ask for it
     if (requireConfirmation) {
-      const rejectReason = await interactor.promptText(
-        `\nPlease type the reason to reject this command (nothing = validate)`
+      const answer = await interactor.chooseOption(
+        ['OK', 'Cancel'],
+        `Confirm execution of: \`${command}\``,
+        undefined,
+        true
       )
-      if (rejectReason) {
-        return `Command execution was cancelled by user with following reason: ${rejectReason}`
+      if (answer !== 'OK') {
+        const reason = answer && answer !== 'Cancel' ? ` with reason: ${answer}` : ''
+        return `Command execution was cancelled by user${reason}`
       }
     }
 


cetautomatix-coday:
## Review — fix: #752 use chooseOption for tool confirmation prompt

The core change is correct and well-executed: replacing `promptText` with `chooseOption` is the right fix, the `Interactor.chooseOption` signature is called with matching argument order (`options`, `question`, `invite`, `allowFreeText`), and the safe-default-to-cancel logic is sound. One functional bug and one incomplete scope item need addressing before merge.

---

### `libs/function/src/lib/run-bash.ts` — argument order mismatch with `allowFreeText`

```typescript
const answer = await interactor.chooseOption(
  ['OK', 'Cancel'],
  `Confirm execution of: \`${command}\``,
  undefined,       // ← third param: optionalQuestion
  true             // ← fourth param: allowFreeText
)
```

The `Interactor.chooseOption` signature is:
```typescript
async chooseOption(options: string[], question: string, invite?: string, allowFreeText?: boolean): Promise<string>
```

The call is correct — `undefined` is passed explicitly for `invite` and `true` for `allowFreeText`. However, with `allowFreeText: true`, the user can type anything as a free-text response. The cancellation branch then treats any non-`'OK'` string that is also not `'Cancel'` as a rejection reason and surfaces it:

```typescript
const reason = answer && answer !== 'Cancel' ? ` with reason: ${answer}` : ''
```

This means if a user accidentally types `'ok'` (wrong case), `'yes'`, or any whitespace-only string, the command is **silently cancelled** without a clear indication that the case mismatch was the problem. The free-text path conflates "typed reason" with "mistyped confirmation". Since the UI presents explicit `OK`/`Cancel` buttons, consider whether `allowFreeText: true` is actually needed here — if it is, the cancellation message should make the case-sensitivity requirement explicit to the user.

---

### Incomplete scope: `project-scripts-tools.ts` still uses the old pattern indirectly

Issue #752 explicitly listed `libs/integration/src/lib/project-scripts-tools.ts` in scope. That file calls `runBash` with `requireConfirmation: entry[1].requireConfirmation`, so it **does** benefit from this fix automatically — no direct change needed there. However, the issue also noted verifying that the `ChoiceEvent` rendering supports free-text fallback alongside predefined options. There is no evidence in this PR that this was verified. If the frontend `ChoiceEvent` component does not render the free-text input when `allowFreeText: true` is set, the custom rejection reason feature is silently broken. This should be confirmed before merge.

---

### `lastInviteEvent` replay not updated for `ChoiceEvent`

In `Interactor`, `replayLastInvite()` only stores and replays `InviteEvent` instances (via `lastInviteEvent`). Now that dangerous confirmations use `chooseOption` (which emits `ChoiceEvent`), a client reconnecting mid-confirmation will not have the choice prompt replayed. This is a pre-existing architectural gap but this PR widens its impact to a security-relevant interaction (command confirmation). It should be documented as a known limitation or tracked as a follow-up issue.

vincent-audibert-whoz (approved):
Tested locally.

PR #754, opened 2026-04-15
feat: #31416 remove TextChunkEvents from saved case events

cetautomatix-coday:
## Summary

`TextChunkEvent`s are streaming-only fragments emitted during live LLM generation for progressive display on the client. They are superseded by the final `MessageEvent` and have no replay value — persisting them bloats the event store unnecessarily.

## Change

In `CaseServiceImpl.storeEvent()`, `TextChunkEvent`s are now returned as-is without being written to the repository. They still flow through the runtime's in-memory event list and SSE flow unchanged, so the client experience is unaffected.

## Test

Added a test in `CaseServiceImplSpec` that verifies: an agent emitting `TextChunkEvent`s followed by `AgentFinishedEvent` leaves no `TextChunkEvent` in the persisted event store, while all orchestration events are still saved correctly.

Closes #WZ-31416

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/caseEvent/CaseEvent.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/caseEvent/CaseEvent.kt
@@ -81,6 +81,23 @@ sealed interface CaseEvent : Entity {
 }
 
 /**
+ * Marker interface for [CaseEvent] subtypes that must NOT be persisted.
+ *
+ * Transient events are emitted on the SSE flow for real-time display but carry
+ * no replay value — they are superseded by the durable events that follow them
+ * (e.g. [TextChunkEvent]s are superseded by the final [MessageEvent]).
+ *
+ * Rules for transient events:
+ * - They reach the SSE flow unconditionally.
+ * - They are never written to the event store (neither in-memory nor Neo4j).
+ * - They are never pushed into the runtime's in-memory event list.
+ *
+ * Plugin authors may mark their own event subtypes as transient by implementing
+ * this interface alongside [CaseEvent].
+ */
+interface TransientCaseEvent
+
+/**
  * Emitted when the case status changes.
  */
 data class CaseStatusEvent(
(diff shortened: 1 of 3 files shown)


cetautomatix-coday:
## Review — `feat: #31416 remove TextChunkEvents from saved case events`

The implementation is correct and the logic is sound: `TextChunkEvent`s are streaming-only, the early return in `storeEvent` is the right place to intercept them, and the event still flows through the runtime's in-memory list and SSE channel untouched. The test validates the right thing. Two issues to address before merging.

---

### 1. `TextChunkEvent` is added to the in-memory event list unnecessarily — `CaseServiceImpl.kt`

```kotlin
// CaseServiceImpl.runAgent — collect block
val saved = storeEvent(event)          // ← for TextChunkEvent, returns the *unsaved* event as-is
if (event.caseId == caseId) {
    runtime.pushEvents(listOf(saved))  // ← TextChunkEvent is pushed into the runtime's InMemoryCaseEventList
}
runtime.emitEvent(saved)              // ← TextChunkEvent is also emitted on SSE — correct
```

`storeEvent` short-circuits for `TextChunkEvent` and returns it unchanged, but the caller still calls `runtime.pushEvents(listOf(saved))`. This means every `TextChunkEvent` is inserted into `InMemoryCaseEventList`, which:

- bloats the list that `processNextStep` scans on every iteration (proportional to the number of chunks per response);
- is inconsistent with the stated intent: "they are superseded by the final `MessageEvent` and have no replay value".

The same guard that blocks persistence should block in-memory accumulation. The SSE emission path (`runtime.emitEvent`) is the only path that should remain unconditional.

---

### 2. Test does not assert that `TextChunkEvent`s are emitted on the SSE flow — `CaseServiceImplSpec.kt`

The new test `"TextChunkEvents emitted by the agent are not persisted in the event store"` only asserts the negative (chunks absent from the event store). It does not verify the positive requirement: that `TextChunkEvent`s *do* reach the SSE flow for progressive display. Without this assertion, a future regression that silently drops chunks before emission would not be caught by this test.

The existing `"handleStatusChange emits RUNNING then IDLE CaseStatusEvents on the runtime Flow"` test shows the pattern: subscribe to `runtime.events`, collect, assert. The same approach should be applied here to confirm that both `TextChunkEvent`s (chunk `"Hello"` and `" world"`) appear in the collected SSE events.

---

### Code style note — `CaseServiceImpl.kt`

Per `agentos/docs/code-style.md`: *"Prefer `when` over chained `if`/`else if`"*. The new `storeEvent` uses a bare `if` for a two-branch dispatch. `when (event)` with a `is TextChunkEvent` branch and an `else` branch is the expected form here, and also makes it straightforward to add future exclusions without introducing a chain.

vincent-couturier-whoz (approved):

PR #757, opened 2026-04-16
feat: #WZ-31326 add optional configPath to Namespace entity

cetautomatix-coday:
## WZ-31326 — Align agent structure between Coday & AgentOS

Adds an optional `configPath` property to the `Namespace` entity — a filesystem path pointing to a directory containing base configuration for the namespace (agent definitions, tool configs, etc.).

Analogous to `projectPath` in Coday: it tells the filesystem plugin where to resolve namespace-scoped resources. When absent, no filesystem-based configuration is loaded.

### Changes

**Backend (agentos-service)**
- `Namespace` domain model — new `configPath: String?` field
- `NamespaceResource` DTO — new `configPath: String?` field with `@Schema` description
- `NamespaceController` — `toResource()` / `toDomain()` updated
- `NamespaceNode` (Neo4j) — property added, round-trip in `toDomain()` / `fromDomain()`
- Tests — `configPath` round-trip added to `NamespacePersistenceLifecycleSpec` and `AbstractNamespacePersistenceSpec`

**Frontend / API client** — updated on the Angular side (included in this branch).

### Notes
- The seeded `default_namespace` intentionally has no `configPath` (generic namespace, no project directory)
- Neo4j: no migration needed — new nullable property, backward-compatible

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
@@ -2,6 +2,7 @@ package io.whozoss.agentos.namespace
 
 import io.swagger.v3.oas.annotations.media.Schema
 import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
 import java.util.UUID
 
 /**
@@ -17,5 +18,11 @@ data class NamespaceResource(
     val name: String,
     val description: String? = null,
     @Schema(description = "Optional filesystem path to a directory containing base configuration for this namespace (agents, tools, etc.)")
+    // Reject path-traversal sequences (".." segments) to prevent a malicious or mistaken value
+    // from escaping the intended directory when the filesystem plugin resolves resources.
+    @field:Pattern(
+        regexp = """^(?!.*\.\.).*""",
+        message = "configPath must not contain path-traversal sequences (..)"
+    )
     val configPath: String? = null,
 )
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
@@ -2,7 +2,6 @@ package io.whozoss.agentos.namespace
 
 import io.swagger.v3.oas.annotations.media.Schema
 import jakarta.validation.constraints.NotBlank
-import jakarta.validation.constraints.Pattern
 import java.util.UUID
 
 /**
@@ -18,11 +17,5 @@ data class NamespaceResource(
     val name: String,
     val description: String? = null,
     @Schema(description = "Optional filesystem path to a directory containing base configuration for this namespace (agents, tools, etc.)")
-    // Reject path-traversal sequences (".." segments) to prevent a malicious or mistaken value
-    // from escaping the intended directory when the filesystem plugin resolves resources.
-    @field:Pattern(
-        regexp = """^(?!.*\.\.).*""",
-        message = "configPath must not contain path-traversal sequences (..)"
-    )
     val configPath: String? = null,
 )
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
@@ -16,4 +16,6 @@ data class NamespaceResource(
     @field:NotBlank(message = "name must not be blank")
     val name: String,
     val description: String? = null,
+    @Schema(description = "Optional filesystem path to a directory containing base configuration for this namespace (agents, tools, etc.)")
+    val configPath: String? = null,
 )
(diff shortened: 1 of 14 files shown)


cetautomatix-coday:
## Review

The change is well-scoped and clean. The full round-trip (domain model → Neo4j node → DTO → Angular form) is consistent, and the test coverage follows the established contract-spec pattern (in-memory lifecycle + abstract Neo4j spec inherited by both embedded and Testcontainers runners). No security or structural blockers.

Two issues to address before merge:

**1. `configPath` is not validated server-side** — `NamespaceResource.configPath` accepts any arbitrary string and is stored as-is. Since this path will be used by the filesystem plugin to resolve namespace-scoped resources (agent definitions, tool configs, etc.), an unvalidated path is a path-traversal risk: a malicious or mistaken value like `../../etc` would be handed directly to the filesystem plugin at runtime. At minimum, basic sanitization or a format constraint should be applied here before the field is consumed downstream.

**2. Angular form: `configPath` control is `nonNullable: true` with default `''`** — this means an empty form submission sends `configPath: ''` (empty string) to the `submit()` handler, which then trims it and converts to `undefined` only for the create path. For the update path the trimmed empty string is passed as `configPath: ''` inside `...this.existingNamespace!` spread because the explicit `configPath:` key overrides the spread only when the trimmed value is truthy. Verify that `'' || undefined` is correctly evaluated and that the update path reliably clears `configPath` to `null`/`undefined` rather than storing an empty string in the backend.

---

**Minor observation** — `AbstractNamespacePersistenceSpec` does not cover the `configPath` null→value update direction (only value→null is tested). This is not a blocker but leaves a gap in the Neo4j persistence contract.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
```diff
@@ -16,4 +17,12 @@ data class NamespaceResource(
     @field:NotBlank(message = "name must not be blank")
     val name: String,
     val description: String? = null,
+    @Schema(description = "Optional filesystem path to a directory containing base configuration for this namespace (agents, tools, etc.)")
+    // Reject path-traversal sequences (".." segments) to prevent a malicious or mistaken value
+    // from escaping the intended directory when the filesystem plugin resolves resources.
+    @field:Pattern(
```
vincent-couturier-whoz: Thought: Perhaps we should have the same control as the one in file but I'm not sure of the root of the config.
issue: event if you cannot use ".." you can use "/" so you can go the root and then in the path you want

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/namespace/NamespaceControllerMvcIntegrationSpec.kt
vincent-couturier-whoz, line 1: praise: ❤️ to have sliced or integration test on namespace endpoints

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceResource.kt
vincent-audibert-whoz: Not sure it is relevant here to block as much the configPath potential values: it will be set by namespace administrators, could be anywhere, on windows systems, remote disks, etc... 🤷 .

Also, this folder is will not be reachable by this way to agents, if agents would access this folder, it would be through the File tools that have their own guards.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/namespace/NamespaceControllerMvcIntegrationSpec.kt
vincent-audibert-whoz: Dev agent initiative 🤖 ❤️

cetautomatix-coday:
## Review

This is a clean, well-scoped PR. The `configPath` field is correctly threaded through all layers — domain model, DTO, Neo4j node, controller mapping, and frontend form — with consistent null/blank normalization and solid test coverage for the round-trip and edge cases. No security issues, no dependency changes, no unrelated files.

Two points to address:

**1. Security concern — `configPath` accepts arbitrary filesystem paths with no validation (`NamespaceResource.kt`)**

```kotlin
@Schema(description = "Optional filesystem path to a directory...")
val configPath: String? = null,
```

The field accepts any string and is stored as-is. There is no validation preventing path traversal sequences (`../..`), absolute paths pointing to sensitive system directories (`/etc`, `/root`), or paths outside an expected base directory. Since this value is described as "tells the filesystem plugin where to resolve namespace-scoped resources," it will eventually be used to load files — making it a potential path traversal vector. At minimum, a `@Pattern` constraint should reject traversal sequences; ideally the accepted format should be documented and enforced (e.g. absolute path allowed or not, max length). This applies to both `NamespaceResource` (input validation) and the controller's `toDomain()` (sanitization).

**2. Branch naming does not follow the convention (`feat/vincent.audibert/wz-31326-namespace-optional-base-config-path`)**

The branch uses `feat/` while the convention requires `feature/username/issue-XXXX-description`. This is a minor process issue but worth flagging for consistency.

cetautomatix-coday:
The security concern on `configPath` is acknowledged and lifted: namespace editing is restricted to administrators, making the access control layer the appropriate enforcement boundary. A pattern-level check would be both insufficient and misleading. No remaining blockers — this PR is approved. ✅

vincent-couturier-whoz (approved):

PR #763, opened 2026-04-16
fix: #655 scheduler threads crash on execution - addAnswerEvent on undefined context

### LINKED ISSUE #655: fix: scheduler threads crash on execution - addAnswerEvent on undefined context
### Symptoms
- Scheduled threads fail to execute and are inaccessible in the UI
- Warning: `"No thread ID in context, MCP tools will not be available"`
- Error: `TypeError: Cannot read properties of undefined (reading 'addAnswerEvent')`
- Regression reported ~2 weeks ago (previously threads would eventually become available)

### Root Cause

**Race condition / timing issue during scheduled execution initialization.**

The execution flow in `prompt-execution.service.ts` creates a thread and passes its ID via `CodayOptions`, but `context.aiThread` is only populated asynchronously after `aiThreadService.select()` completes inside `Coday.run() → initThread()`.

If MCP tools or agents are invoked before `context.aiThread` is set:
1. `toolbox.ts` sees `context.aiThread?.id === undefined` → MCP tools unavailable (warning)
2. `agent.ts` calls `thread.addAnswerEvent()` on `undefined` → TypeError crash

### Key Files
- `libs/service/src/lib/prompt-execution.service.ts` — scheduled execution entry point
- `libs/core/src/lib/core.ts` — `initThread()` sets `context.aiThread` asynchronously
- `libs/model/src/lib/agent.ts` — calls `thread.addAnswerEvent()` (crashes if thread undefined)
- `libs/agent/src/lib/toolbox.ts` — checks `context.aiThread?.id` for MCP tools

### Expected Behavior
Scheduled threads execute successfully and are visible/accessible in the thread list UI.

### Actual Behavior
Scheduled threads crash on startup, leaving empty/broken thread entries. The scheduler feature is effectively non-functional.

### Fix Direction
Ensure `context.aiThread` is fully initialized before any agent or tool execution begins in the scheduled (oneshot) execution path. Consider:
- Awaiting `initThread()` completion before proceeding with agent execution
- Adding a defensive guard in `agent.ts` with a clear error if thread context is missing

cetautomatix-coday:
## Problem

Scheduled threads crash on startup with:
- Warning: `"No thread ID in context, MCP tools will not be available"`
- Error: `TypeError: Cannot read properties of undefined (reading 'addAnswerEvent')`

This was a **race condition** in `core.ts`. `initThread()` called `await this.aiThreadService.select(this.options.thread)` and returned immediately, but `context.aiThread` was only populated via an RxJS subscription declared in the constructor:

```typescript
this.aiThreadService.activeThread.subscribe((aiThread) => {
  if (!this.context || !aiThread) return
  this.context.aiThread = aiThread  // ← async, may not fire before run() continues
  this.replayThread(aiThread)
})
```

The `run()` loop continued before the subscription fired, leaving `context.aiThread` as `undefined` when `AiHandler.runAgent()` tried to call `agent.run(cmd, context.aiThread!, ...)`.

## Fix

**`libs/core/src/lib/core.ts`** — `initThread()` now directly assigns the return value of `select()` to `context.aiThread`, guaranteeing it is set synchronously before any agent or tool execution begins. `select()` already returns the resolved `AiThread`, so no API changes were needed.

**`libs/model/src/lib/agent.ts`** — Added a defensive guard at the top of `Agent.run()` that throws a clear, actionable error if `thread` is undefined, instead of the cryptic `TypeError: Cannot read properties of undefined (reading 'addAnswerEvent')`.

## Testing

- `@coday/core` and `@coday/model`: lint ✅ build ✅

Closes #655

### DIFF
--- a/libs/core/src/lib/core.ts
+++ b/libs/core/src/lib/core.ts
@@ -366,7 +366,14 @@ export class Coday {
     if (!this.options.thread) {
       throw Error('No thread given, cannot start Coday instance')
     }
-    await this.aiThreadService.select(this.options.thread)
+    // Directly assign the resolved thread to context so it is guaranteed to be set
+    // before any agent or tool execution begins. Relying solely on the RxJS subscription
+    // in the constructor creates a race condition: select() may return before the
+    // subscription fires, leaving context.aiThread undefined when run() continues.
+    const thread = await this.aiThreadService.select(this.options.thread)
+    if (this.context) {
+      this.context.aiThread = thread
+    }
   }
 
   private async initCommand(): Promise<string | undefined> {
--- a/libs/model/src/lib/agent.ts
+++ b/libs/model/src/lib/agent.ts
@@ -53,6 +53,11 @@ export class Agent {
    * @param thread
    */
   async run(command: string, thread: AiThread, username?: string): Promise<Observable<CodayEvent>> {
+    if (!thread) {
+      throw new Error(
+        `[Agent:${this.name}] Cannot run: thread context is undefined. Ensure initThread() completed before agent execution.`
+      )
+    }
     const trimmedCommand = command.trim()
 
     // Add AnswerEvent to thread with the original command (including @agentName if present)
--- a/libs/service/src/lib/prompt-execution.service.ts
+++ b/libs/service/src/lib/prompt-execution.service.ts
@@ -278,20 +278,20 @@ export class PromptExecutionService {
       }
     } else {
       // Asynchronous mode: return immediately with thread ID
-      // Start Coday run in background
-      instance.coday!.run().catch((error: unknown) => {
-        console.error('[PROMPT_EXEC] Error during prompt Coday run:', error)
-      })
-
-      // Schedule cleanup after a reasonable timeout (e.g., 5 minutes)
-      setTimeout(
-        () => {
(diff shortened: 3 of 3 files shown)


cetautomatix-coday:
## Second fix added

A second bug was identified during testing: the thread was created and appeared empty, but only ran when the user opened it manually.

**Root cause:** After the scheduled `run()` completed, the `ThreadCodayInstance` was kept alive in the manager (30-min oneshot timeout). When the user opened the thread via SSE, `getOrCreate()` found the existing instance, but `replayThreadHistory()` failed silently because `coday.context` had been nulled out by `Coday.cleanup()`. Then `startCoday()` triggered `run()` again on the exhausted Coday instance — which re-ran the original prompts from scratch instead of loading history.

**Fix in `libs/service/src/lib/prompt-execution.service.ts`:** Call `threadCodayManager.cleanup()` immediately in the `.finally()` of the background run, rather than after a 5-minute timeout. This removes the instance from the map as soon as execution finishes. When the user then opens the thread, `getOrCreate()` creates a **fresh** instance with standard (non-oneshot, no-prompts) options — it loads the persisted thread history from disk and waits for user input normally.

vincent-audibert-whoz (approved):

PR #770, opened 2026-04-17
feat: #WZ-31326 use AgentConfig entities in AgentService resolution

cetautomatix-coday:
## What

Wire the existing `AgentConfig` entities into `AgentServiceImpl` so that agents are resolved from persistent configuration rather than directly from `AiModel` aliases.

## Why

WZ-31326 — `AgentConfig` entities exist in the database but were not yet used anywhere. The service was resolving agents purely from `AiModel`/`AiProvider` pairs, ignoring agent-level instructions and names stored in `AgentConfig`.

### `AgentConfigService` — two new methods
- `findByName(namespaceId, name)` — case-insensitive name lookup within a namespace
- `findDefault(namespaceId)` — returns the oldest `AgentConfig` in the namespace (first created), or null

Both are implemented in `AgentConfigServiceImpl` on top of the existing `findByParent` call — no new repository queries needed.

### `AgentServiceImpl` — AgentConfig-first resolution

All four public methods now follow the same pattern: **AgentConfig first, AiModel fallback**.

| Method | With AgentConfig | Without AgentConfig (fallback) |
|---|---|---|
| `findAgentByName` | Uses `AgentConfig.name` + `instructions`, resolves model via `AgentConfig.modelName` | Legacy: resolves directly against `AiModel` alias/apiName |
| `getDefaultAgent` | Returns agent built from the oldest `AgentConfig` in the namespace | Legacy: first `AiModel` with alias \"default\" |
| `getDefaultAgentName` | Returns `AgentConfig.name` | Legacy: `AiModel` alias or apiName |
| `resolveAgentName` | Returns `AgentConfig.name` | Legacy: `AiModel` alias or apiName |

When an `AgentConfig` has no `modelName`, the namespace default `AiModel` is used as fallback.

### Tests

`AgentServiceImplUnitSpec` fully updated: existing tests retain the legacy fallback coverage, new tests cover the AgentConfig-first path (name resolution, instruction injection, model lookup via `modelName`, null-`modelName` fallback).


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -49,38 +48,66 @@ class AgentServiceImpl(
...
+        context: AgentExecutionContext,
+    ): Agent {
+        val modelLookupName = config.modelName
+        val (modelConfig, providerConfig) =
+            when {
+                modelLookupName != null -> resolveModelPair(modelLookupName, context.namespaceId)
+                else -> {
+                    val defaultModel =
+                        findDefaultModelConfig(context.namespaceId)
+                            ?: throw IllegalArgumentException(
```
vincent-couturier-whoz, line 104: suggestion: perhaps we should have a status on the agent config validation

cetautomatix-coday:
## Review — feat: #WZ-31326 use AgentConfig entities in AgentService resolution

The PR is well-scoped and architecturally coherent: `AgentConfigService.findByName/findDefault` are clean, the `FilesystemAgentConfigRepository` decorator pattern is solid, and test coverage is thorough. A few issues need to be addressed before merge.

---

### `AgentConfigServiceImpl.findDefault` — wrong return type contracts

`findDefault` is declared on the interface as returning `AgentConfig` (non-null), but the implementation can silently return `DEFAULT_AGENT_CONFIG` whose `namespaceId` is the all-zeros UUID (`00000000-0000-0000-0000-000000000000`). In `AgentServiceImpl.getDefaultAgent`, this flows into `createAgentFromConfig` which then calls `resolveModelPair` or `findDefaultModelConfig` scoped to that bogus namespace UUID, silently failing with no model found. The fallback returns `null` via `runCatching.getOrNull()`, which is swallowed — but the root cause (wrong namespace on the fallback entity) is invisible to the caller. The `DEFAULT_AGENT_CONFIG` constant should either carry no `namespaceId` at all (if the type allows it) or `getDefaultAgent` must substitute the real `context.namespaceId` before delegating to `createAgentFromConfig`.

---

### `buildIntegrationsPayload` — `null as any` type cast

In `agent-config-form.component.ts`:

```typescript
result[name] = null as any
```

The comment acknowledges the mismatch with the generated TS type (`Array<string>`). This is a type-safety violation: the generated client declares `[key: string]: Array<string>`, but `null` is intentionally pushed through. The correct fix is either to align the OpenAPI spec so the generated type accepts `null` values (i.e. `Array<string> | null`), or to use a local intermediate type for the payload that correctly models the nullable list. A `null as any` cast hides a contract mismatch between the frontend and the API client and will silently break if the generated client adds runtime validation.

---

### `AgentConfigServiceImpl.findDefault` — O(n) full scan on every call

`findByParent` fetches all configs for the namespace, then `minByOrNull` scans the entire list to find the oldest. For `getDefaultAgentName` (called e.g. on every case load), this means a full namespace scan with no short-circuit. This is acceptable now given the small data volumes, but there is no dedicated repository method for "find oldest by namespace". At minimum, a `// TODO: add findOldestByParent to avoid full scan` comment would make the intent explicit and prevent this pattern from being copied to higher-cardinality entities.

---

### `FilesystemAgentConfigRepository` — `PLACEHOLDER_NAMESPACE_ID` leaks on partial failures

In `parseYamlFile`, the returned `AgentConfig` carries `PLACEHOLDER_NAMESPACE_ID` as `namespaceId`. The overwrite in `findByParent` is correct for the happy path, but if a future caller invokes `parseYamlFile` directly (e.g. for validation), the placeholder leaks. The method is `private` today, so this is low risk, but the companion object comment "Always overwritten in `findByParent` before the config is returned" documents a fragile invariant that should ideally be enforced structurally (e.g. by making `namespaceId` a parameter of `parseYamlFile`).

---

### `Neo4jAgentConfigRepository.deleteByParent` — escaped string interpolation bug

```kotlin
logger.debug { "[Neo4jAgentConfigRepository] Soft-deleted \${active.size} agent configs under namespace $parentId" }
```

The `\${active.size}` is escaped with a backslash — this will print the literal string `${active.size}` rather than the count. The `$parentId` on the same line works correctly because it is not escaped. This is a cosmetic bug in logging but should be fixed for debuggability.

---

### `AgentConfigServiceImplUnitSpec` — `findDefault` ordering test relies on `Instant.now()` for non-"oldest" entries

In the `findDefault returns the oldest config when multiple exist` test, `older` is created with `Instant.ofEpochSecond(1000)` and `Beta` with `Instant.ofEpochSecond(2000)`, which is correct and deterministic. However, the `findDefault returns the single config when only one exists` test uses `config(...)` without specifying `createdAt`, which defaults to `Instant.now()`. This is fine in isolation but creates a temporal coupling if the test infrastructure ever freezes time. Minor, but worth making explicit with a fixed timestamp.

---

### `AgentConfigServiceImplUnitSpec` — uses `nsId = UUID.fromString("00000000-0000-0000-0000-000000000001")` as default

The `config()` helper hardcodes a fixed namespace UUID as default (`nsId`), which differs from the `namespaceId` variable used in the test assertions. Tests that pass `nsId = namespaceId` explicitly are correct, but the asymmetry is a subtle trap: a new test that forgets to pass `nsId = namespaceId` will silently scope the config to the wrong namespace. Consider making `nsId` default to `namespaceId` in the helper, or removing the default entirely to force explicit scoping.

cetautomatix-coday:
## Updated review — 2026-04-28

I've re-read all modified files at the current HEAD (`c15fe1ef`). Here is a precise status of every item from the original review.

---

### Fix assessment: 5/5 confirmed adequate

**1. Interpolation bug in `Neo4jAgentConfigRepository.deleteByParent`** ✅  
`${active.size}` is now correct Kotlin string interpolation. The log line in `deleteByParent` reads cleanly.

**2. Bogus namespace UUID in `getDefaultAgent`** ✅  
The fix is correct and minimal. The sentinel is compared by value, not by identity, which is right for UUID. The `BOGUS_NAMESPACE_ID` constant is private to `AgentServiceImpl.companion` and clearly documented. One minor note: `getDefaultAgentName` also calls `agentConfigService.findDefault` but does **not** apply the same substitution — this is intentional and correct because `getDefaultAgentName` only returns the name string, not an object that will be passed to `createAgentFromConfig`, so the bogus namespace never propagates to a downstream call.

**3. O(n) full-scan TODO** ✅  
The comment is in place and clearly describes the required fix (`findOldestByParent` repository method). Acceptable for now.

**4. Test helper `config()` default `nsId`** ✅  
`nsId: UUID = namespaceId` correctly references the test-level `val namespaceId = UUID.randomUUID()`. The scoping trap is gone.

**5. Test `findDefault` single config determinism** ✅  
`createdAt: Instant = Instant.ofEpochSecond(1_000_000)` is a fixed epoch value. The test `"findDefault returns the single config when only one exists"` is now deterministic.

---

### Remaining open item: `null as any` in `agent-config-form.component.ts`

The cast is still present (line ~183):
```typescript
result[name] = null as any
```

The stated reason for not fixing it — "generated TS type declares `Array<string>`" — is accurate: the generated client types for `AgentConfig.integrations` are `{ [key: string]: Array<string> }`, which does not include `null` as a value type, even though the backend contract accepts it. The cast is therefore the pragmatic workaround for a generated-type gap.

**This is not a blocker for this PR**, but it should not be left as technical debt without a tracking path. The correct fix is either:
- Annotate the OpenAPI spec so the generator emits `Array<string> | null` for the integration map values, or
- Add a comment next to the cast explaining the contract mismatch and referencing the spec fix.

Currently there is no comment explaining why `null as any` is used here, which means the next developer to touch this code will not understand whether it is intentional or a bug. **Frontay should add a one-line comment before this is merged.**

---

### One new concern spotted during re-read

**`FilesystemYamlCache.reload` uses `Files.walk` without closing the stream.**

```kotlin
val items =
    Files
        .walk(directory)
        .filter { Files.isRegularFile(it) }
        .filter { it.toString().endsWith(".yaml") || it.toString().endsWith(".yml") }
        .toList()   // ← stream is consumed but never explicitly closed
        .mapNotNull { … }
```

`Files.walk` returns a `Stream<Path>` that holds an open directory handle. Calling `.toList()` consumes the stream but does not close it. On most JVMs the stream will be closed by GC, but the JDK documentation explicitly states that the stream **must** be closed to release the associated file-system resources. Under load (many namespaces, frequent TTL refreshes) this can exhaust file descriptors. The stream should be wrapped in a `use {}` block or collected inside a `Files.walk(directory).use { … }` call. This is a resource-leak bug — not a blocker today given the 5-minute TTL, but it is a correctness issue in production-grade code (`agentos-service`).

---

### Summary

5 of the 7 original issues are correctly resolved. The `null as any` cast can be unblocked with a one-line explanatory comment assigned to Frontay. The resource leak in `FilesystemYamlCache.reload` is a new finding that must be fixed before merge. No other concerns.

cetautomatix-coday:
The `null as any` cast has been addressed in commit `7825ad1b`.

The cast is kept intentionally — the Kotlin backend declares `Map<String, List<String>?>?` where a null list value means "all tools from this integration are allowed", which is the correct contract. The generated TS type omits `| null` because springdoc-openapi does not emit `nullable: true` on `additionalProperties` for Kotlin nullable generics — this is a known springdoc limitation.

The fix adds:
- A detailed explanatory comment next to the cast (backend contract, springdoc limitation, intentional nature of the cast) + `eslint-disable-next-line` scoped to that line
- A "Rules" section in `agentos/docs/openapi-and-client.md` documenting the OpenAPI workflow rules (never edit the spec by hand, use springdoc annotations) and the known nullable Map limitation as a reference for future occurrences

cetautomatix-coday:
## Review — follow-up on `null as any` cast (commit `7825ad1b`)

The last open item is resolved. Here is what was verified:

**`agent-config-form.component.ts` — `buildIntegrationsPayload()`**

The `null as any` cast now has a three-part inline comment that covers exactly what is needed: the backend Kotlin type (`Map<String, List<String>?>?`), the specific springdoc limitation (does not emit `nullable: true` on Map `additionalProperties` for nullable generics), and an explicit statement that the cast is intentional. The `eslint-disable-next-line` directive is correctly scoped to that single line — it neither suppresses anything broader nor leaves the line uncommented. The cast itself is contained within the local `result` map assignment and does not propagate to the method's return type, so the `any` stays local.

**`agentos/docs/openapi-and-client.md`**

The new document covers the full OpenAPI/client generation workflow (spec generation, client generation, the CI check, the `src/custom/` escape hatch, and the four hard rules). The known springdoc limitation is documented in a dedicated section at the end, which is exactly the right place for future developers and agents to find it when they encounter the same constraint.

No remaining issues. **This PR is ready to merge (squash).**

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -49,38 +48,69 @@ class AgentServiceImpl(
...
+                )
+        return createAgentFromConfig(agentConfig, context)
     }
 
     override fun getDefaultAgent(context: AgentExecutionContext): Agent? {
-        val modelConfig = findDefaultModelConfig(context.namespaceId) ?: return null
-        val providerConfig = aiProviderService.getById(modelConfig.aiProviderId)
-        return createAgentInstance(modelConfig.alias ?: modelConfig.apiModelName, modelConfig, providerConfig, context)
+        val config = agentConfigService.findDefault(context.namespaceId)
+            .let { if (it.namespaceId == BOGUS_NAMESPACE_ID) it.copy(namespaceId = context.namespaceId) else it }
```
vincent-couturier-whoz, line 67: question: why this "BOGUS_NAMESPACE_ID" ? In which case we should have it ? suggestion: Perhaps we could have a function with a more meaningfull name like it.hasNoRealNamespace()

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigServiceImpl.kt
```diff
@@ -21,4 +22,37 @@ class AgentConfigServiceImpl(
...
+        /**
+         * Built-in fallback agent returned when a namespace has no persisted [AgentConfig].
+         *
+         * Uses a stable UUID derived from the name so the identity is consistent across
+         * restarts. [modelName] is null: the namespace's default [AiModel] will be used.
+         */
+        val DEFAULT_AGENT_CONFIG =
+            AgentConfig(
+                metadata = EntityMetadata(id = UUID.nameUUIDFromBytes("default-agent".toByteArray())),
+                namespaceId = UUID.fromString("00000000-0000-0000-0000-000000000000"),
```
vincent-couturier-whoz, line 51: suggestion : Perhaps we could/should reuse this "DEFAULT_NAMESPACEID"

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -209,5 +253,8 @@ class AgentServiceImpl(
             .joinToString("\n")
     }
 
-    companion object : KLogging()
+    companion object : KLogging() {
+        /** Sentinel namespace UUID emitted by [AgentConfigServiceImpl.DEFAULT_AGENT_CONFIG]. */
+        private val BOGUS_NAMESPACE_ID: UUID = UUID.fromString("00000000-0000-0000-0000-000000000000")
```
vincent-couturier-whoz, line 258: suggestion: Perhaps we could have a more specific / representative name like NO_NAMESPACE_ID

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNode.kt
```diff
@@ -27,6 +32,7 @@ data class AgentConfigNode(
     val description: String? = null,
     val instructions: String? = null,
     val modelName: String? = null,
+    val integrationsJson: String? = null,
```
vincent-couturier-whoz, line 35: thought: it's strange because I think we can store those information as neo4j objects (nodes + relationships) but I think spring data is not well designed for that. It's not necessary for us today but I think it could be easier to request the database to see for example where an integration is used. We can store the data this way at the beginning because it's more straight forward and perhaps more efficient in terms of request but in near future we should switch to a more flexible and readable model. It would be interesting to test what's the result to use directly a map of list of string to see how spring data manage this information. I think it won't work and we will need to create custom query to consolidate custom node and relationship to store the map as relationship to "integration" node.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/FilesystemYamlCache.kt
```diff
@@ -0,0 +1,107 @@
...
+ * [parser] receives each regular `.yaml` / `.yml` file and returns the parsed
+ * entity, or null if the file should be skipped (parse error, invalid content).
+ * Errors inside [parser] are caught and logged — a bad file never prevents the
+ * rest of the directory from loading.
+ *
+ * Thread-safe: concurrent calls to [getAll] during a reload are serialised via
+ * the instance lock. The lock is only held during the filesystem scan, not during
+ * the return of the cached list.
+ */
+class FilesystemYamlCache<T>(
```
vincent-couturier-whoz, line 25: suggestion: We could use cacheable annotation to do it... but it's clearly not a priority but something we can do in another PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
```diff
@@ -77,9 +79,13 @@ class Neo4jPersistenceConfiguration {
     fun neo4jAgentConfigRepository(
         agentConfigNodeNeo4jRepository: AgentConfigNodeNeo4jRepository,
         childLinkService: Neo4jChildLinkService,
+        namespaceService: NamespaceService,
     ): AgentConfigRepository {
-        logger.info { "[Persistence] Neo4jAgentConfigRepository active" }
-        return Neo4jAgentConfigRepository(agentConfigNodeNeo4jRepository, childLinkService)
+        logger.info { "[Persistence] Neo4jAgentConfigRepository active (filesystem augmentation enabled)" }
+        return FilesystemAgentConfigRepository(
```
vincent-couturier-whoz, line 85: suggestion : perhaps we should condition the filesystem agent config repository to a specific config properties because I'm not it's always useful.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
```diff
@@ -497,6 +497,94 @@ class CaseServiceImplSpec :
...
+            // First message: explicit @mention
+            service.addMessage(
+                caseId = case.id,
+                actor = userActor,
+                content = listOf(MessageContent.Text("@$selectedAgentName hello")),
+            )
+            val deadline1 = System.currentTimeMillis() + 5_000
+            while (System.currentTimeMillis() < deadline1) {
+                if (service.getById(case.id).status == CaseStatus.IDLE) break
+                Thread.sleep(50)
```
vincent-couturier-whoz, line 562: question : why those thread sleeps if we mock agentService ? it seems weird in a unit test. Perhaps we can use tick function to simulate time progression to improve the test

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/tool/ToolRegistryServiceSpec.kt
```diff
@@ -58,9 +58,8 @@ class ToolRegistryServiceSpec : StringSpec({
         plugins: List<ToolPlugin> = emptyList(),
         configs: List<IntegrationConfig> = emptyList(),
     ): ToolRegistryService {
-        val pluginManager = mockk<PluginManager>(relaxed = true)
+        val pluginManager = mockk<org.pf4j.PluginManager>(relaxed = true)
```
vincent-couturier-whoz, line 61: suggestion : use an import

agents/PM.yaml
```diff
@@ -4,43 +4,35 @@ aiProvider: anthropic
 modelName: BIG
 integrations:
   GITHUB:
-#    - add_comment_to_pending_review
     - add_issue_comment
-#    - add_reply_to_pull_request_comment
-#    - assign_copilot_to_issue
 #    - create_branch
+    - create_issue
```
vincent-couturier-whoz, line 9: question: what's the relationship with this PR ?

vincent-couturier-whoz (approved):

vincent-audibert-whoz:
@vincent-couturier-whoz comments grouped in github issues #824, #825, #826, #827 and #828, merging right away

agents/PM.yaml
vincent-audibert-whoz: opportunistic PM fix.

PR #783, opened 2026-04-21
feat: #781 OAuth discovery support for HTTP integrations (RFC 8414 / OIDC)

### LINKED ISSUE #781: feat: OAuth discovery support for HTTP integrations (RFC 8414 / OIDC)
Context
Currently, Coday's OAuth system (Authorization Code + PKCE via oauth4webapi) requires all endpoints to be configured manually. There is no automatic discovery — neither RFC 8414 (OAuth Server Metadata) nor OpenID Connect Discovery (.well-known/openid-configuration).

To configure an HTTP integration with OAuth today, users must explicitly provide:
```
oauth2:
  client_id: "..."
  authorization_endpoint: "https://provider.com/oauth/authorize"  # required
  token_endpoint: "https://provider.com/oauth/token"              # required
  redirect_uri: "http://localhost:3000/oauth/callback"
  scope: ["openid", "email"]
```
This is friction for well-known providers (Google, Microsoft, Okta, etc.) that all expose standard metadata endpoints.

Note on MCP OAuth
mcp-oauth-provider.ts already handles RFC 9728 discovery, but this is delegated to the MCP SDK — it does not help generic HTTP integrations.

Proposed solution
In generic-oauth.ts: if authorization_endpoint and token_endpoint are absent but an issuer or discovery_url is provided, automatically fetch .well-known/openid-configuration or .well-known/oauth-authorization-server to resolve them.

oauth4webapi already supports this natively via its discoveryRequest() function, so the implementation should be straightforward.

# Simplified config after this feature:
```
oauth2:
  client_id: "..."
  issuer: "https://accounts.google.com"   # discovery resolves the rest
  redirect_uri: "http://localhost:3000/oauth/callback"
  scope: ["openid", "email"]
```

Acceptance criteria
 If authorization_endpoint + token_endpoint are present → current behaviour unchanged (no discovery attempted)
 If issuer or discovery_url is provided instead → endpoints are resolved automatically via discovery
 Supports both RFC 8414 (.well-known/oauth-authorization-server) and OIDC (.well-known/openid-configuration)
 Discovery errors surface a clear, actionable message to the user
 Non-breaking: existing configurations continue to work as-is
 Tested against at least one real provider (Google, Microsoft, or Okta)

Labels
enhancement, oauth, integrations

cetautomatix-coday:
## Summary

Implements automatic OAuth endpoint discovery for HTTP integrations (closes #781).

Previously, users had to configure `authorization_endpoint` and `token_endpoint` explicitly. With this change, simply providing an `issuer` (or `discovery_url`) is enough — Coday resolves the endpoints automatically.

### `GenericOAuthConfig` (generic-oauth.ts)
- `authorizationEndpoint` and `tokenEndpoint` are now **optional**
- New optional fields: `issuer` (RFC 8414 / OIDC issuer URL) and `discoveryUrl` (explicit discovery document URL)

### `GenericOAuth` class
- Authorization server metadata is now resolved **lazily** in `startAuthFlow()` via the new `resolveAuthorizationServer()` method
- Discovery strategy:
  1. If `authorizationEndpoint` + `tokenEndpoint` are set → used directly (existing behaviour, no breaking change)
  2. If `issuer` is set → `oauth4webapi.discoveryRequest()` with OIDC first, RFC 8414 fallback
  3. If `discoveryUrl` is set → fetch the document directly
  4. None configured → clear error message
- Discovered metadata is cached after first resolution — no repeated network calls
- `refreshToken()` also uses `resolveAuthorizationServer()` for consistency

### `OAuth2Config` model (integration-config.ts)
- Added `issuer?: string` and `discovery_url?: string` fields with JSDoc

### `HttpTools` (http.tools.ts)
- Validation updated: accepts configs with `issuer` or `discovery_url` as an alternative to direct endpoints
- Passes the new fields through to `GenericOAuth`

## Example — simplified config after this change

```yaml
integration:
  MY_GOOGLE_CAL:
    type: HTTP
    http:
      baseUrl: "https://www.googleapis.com/calendar/v3"
    oauth2:
      client_id: "..."
      client_secret: "..."
      redirect_uri: "http://localhost:3000/oauth/callback"
      issuer: "https://accounts.google.com"   # discovery resolves the rest
      scope:
        - "https://www.googleapis.com/auth/calendar.readonly"
```

## Tests
- All 42 existing tests pass unchanged
- 9 new tests added covering: issuer-based discovery, RFC 8414 fallback, cache behaviour, full flow via discovery, explicit `discoveryUrl`, HTTP error handling, missing fields error, and configuration error

### DIFF
--- a/libs/integrations/http/src/lib/generic-oauth.test.ts
+++ b/libs/integrations/http/src/lib/generic-oauth.test.ts
@@ -653,6 +653,126 @@ describe('GenericOAuth', () => {
     })
   })
 
+  // ── Security: URL validation ────────────────────────────────────────────────
+
+  describe('security: validateDiscoveryUrl', () => {
+    let originalFetch: typeof global.fetch
+    const mockFetch = jest.fn()
+
+    beforeEach(() => {
+      originalFetch = global.fetch
+      global.fetch = mockFetch
+    })
+
+    afterEach(() => {
+      global.fetch = originalFetch
+    })
+
+    const makeInstanceWithIssuer = (issuer: string) =>
+      new GenericOAuth(
+        { clientId: 'id', clientSecret: 'secret', redirectUri: 'https://app.example.com/callback', issuer },
+        makeInteractor(),
+        makeUserService(),
+        'testProject',
+        'testIntegration'
+      )
+
+    const makeInstanceWithDiscoveryUrl = (discoveryUrl: string) =>
+      new GenericOAuth(
+        { clientId: 'id', clientSecret: 'secret', redirectUri: 'https://app.example.com/callback', discoveryUrl },
+        makeInteractor(),
+        makeUserService(),
+        'testProject',
+        'testIntegration'
+      )
+
+    it('should reject issuer with http:// scheme (SSRF prevention)', async () => {
+      const instance = makeInstanceWithIssuer('http://accounts.example.com')
+      await expect(instance.authenticate()).rejects.toThrow('must use HTTPS')
+    })
+
+    it('should reject discoveryUrl with http:// scheme (SSRF prevention)', async () => {
+      const instance = makeInstanceWithDiscoveryUrl('http://accounts.example.com/.well-known/openid-configuration')
+      await expect(instance.authenticate()).rejects.toThrow('must use HTTPS')
+    })
+
+    it('should reject issuer pointing to localhost', async () => {
+      const instance = makeInstanceWithIssuer('https://localhost/auth')
+      await expect(instance.authenticate()).rejects.toThrow('private/loopback')
+    })
+
+    it('should reject discoveryUrl pointing to 127.0.0.1', async () => {
+      const instance = makeInstanceWithDiscoveryUrl('https://127.0.0.1/.well-known/openid-configuration')
+      await expect(instance.authenticate()).rejects.toThrow('private/loopback')
+    })
+
+    it('should reject discoveryUrl pointing to a private 10.x.x.x address', async () => {
+      const instance = makeInstanceWithDiscoveryUrl('https://10.0.0.1/.well-known/openid-configuration')
+      await expect(instance.authenticate()).rejects.toThrow('private/loopback')
+    })
+
+    it('should reject discoveryUrl pointing to 169.254.x.x (cloud metadata)', async () => {
+      const instance = makeInstanceWithDiscoveryUrl('https://169.254.169.254/latest/meta-data/')
+      await expect(instance.authenticate()).rejects.toThrow('private/loopback')
+    })
+
+    it('should reject when authorization_endpoint has a different origin than discoveryUrl', async () => {
+      const instance = makeInstanceWithDiscoveryUrl('https://legit.example.com/.well-known/openid-configuration')
+      mockFetch.mockResolvedValueOnce({
+        ok: true,
+        json: async () => ({
+          issuer: 'https://legit.example.com',
+          authorization_endpoint: 'https://attacker.example.com/oauth/authorize',
+          token_endpoint: 'https://legit.example.com/oauth/token',
+        }),
+      } as any)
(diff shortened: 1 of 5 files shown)


cetautomatix-coday:
## Review — feat: #781 OAuth discovery support

Two security issues to address before merge, both in `generic-oauth.ts`.

---

### 🔴 SSRF via unvalidated `discoveryUrl`

**File:** `libs/integrations/http/src/lib/generic-oauth.ts` — `resolveAuthorizationServer()`, `discoveryUrl` branch

```typescript
if (discoveryUrl) {
  const response = await fetch(discoveryUrl)
  // ...
  const metadata = (await response.json()) as oauth.AuthorizationServer
```

The `discoveryUrl` value comes directly from user-controlled YAML config and is passed to `fetch()` without any URL validation. This is a Server-Side Request Forgery risk: a user can point the discovery URL to internal infrastructure (`http://169.254.169.254/...`, `http://localhost:6379`, etc.), causing Coday's server process to make requests to internal services and potentially exfiltrate responses into the OAuth flow or error messages. The `issuer` field passed to `new URL(issuer)` and `oauth.discoveryRequest()` has the same exposure, but `oauth4webapi` presumably validates issuer URLs against its own allowlist — the raw `fetch(discoveryUrl)` path does not benefit from that protection.

Mitigation: validate that `discoveryUrl` and `issuer` use `https:` scheme and have a non-private hostname before making any network call.

---

### 🔴 Open redirect via unvalidated `authorization_endpoint` in auth URL construction

**File:** `libs/integrations/http/src/lib/generic-oauth.ts` — `startAuthFlow()`

```typescript
const authorizationUrl = new URL(as.authorization_endpoint!)
```

When using the `discoveryUrl` path, `as.authorization_endpoint` is taken verbatim from the fetched JSON document without further validation. If the discovery document is tampered with (e.g. via DNS rebinding, a compromised provider, or an MITM on a non-HTTPS discovery URL), the `authorization_endpoint` could point to an attacker-controlled URL. The `OAuthRequestEvent` then sends this URL to the frontend, which opens it as the OAuth provider — the user's browser is redirected to an attacker site that can harvest the authorization code or credentials. This is compounded by the SSRF issue above (attacker controls both the discovery document and the resulting auth URL). The `issuer`-based path uses `oauth4webapi.processDiscoveryResponse()` which validates issuer binding, but the `discoveryUrl` path skips this check entirely.

Mitigation: after fetching a discovery document via `discoveryUrl`, validate that `authorization_endpoint` and `token_endpoint` share the same origin as the discovery URL, or require `https:` scheme and apply the same issuer-binding check that `oauth4webapi` does for the `issuer` path.

cetautomatix-coday:
Thanks for the thorough review — both issues are addressed in the latest commit.

**SSRF via unvalidated `discoveryUrl` / `issuer`**

Added `validateDiscoveryUrl(rawUrl, fieldName)` called before any network request for both `issuer` and `discoveryUrl`. It enforces:
- `https:` scheme only (rejects `http:`, `file:`, etc.)
- Blocks private/loopback hostnames: `localhost`, `127.x`, `10.x`, `172.16–31.x`, `192.168.x`, `169.254.x`, and IPv6 equivalents

For `issuer`, `oauth4webapi.processDiscoveryResponse()` already validates issuer binding, but the upfront URL check adds a consistent first line of defence.

**Open redirect via unvalidated `authorization_endpoint`**

After fetching the discovery document via `discoveryUrl`, we now verify that both `authorization_endpoint` and `token_endpoint` share the same origin as the discovery URL. If either diverges, the flow is aborted with a clear error message before any URL is sent to the frontend.

The `issuer`-based path is unaffected — `oauth4webapi.processDiscoveryResponse()` already enforces issuer binding there.

**Tests**: 9 new tests added covering all the new security validations (SSRF scenarios + cross-origin endpoint detection).

vincent-audibert-whoz (approved):

PR #785, opened 2026-04-21
feat: #WZ-31170 port tmux tools as agentos-tmux-plugin

cetautomatix-coday:
## Summary

Ports `tmux.tools.ts` from Coday into a proper AgentOS plugin following the datetime plugin pattern.

## New module: `agentos-tmux-plugin`

**`TmuxTool`** — `StandardTool<Input>` with 6 actions: `list`, `status`, `start`, `logs`, `send`, `stop`.

Notable differences vs the TypeScript original:
- `ProcessBuilder` instead of `execFile` — no shell intermediary, no injection risk
- `window` parameter (optional) — targets a specific tmux window within a session via `session:window` syntax (index or name). Missing from the original, which always targeted the active window.
- `workingDirectory` configurable via `IntegrationConfig` (replaces the runtime-injected `projectRoot`)
- Error responses are structured JSON `{"success": false, "error": "..."}` consistent with other AgentOS tools

**`TmuxPlugin` + `TmuxToolProvider`** — PF4J plugin lifecycle + `@Extension` / `ToolPlugin` with a `configSchema` exposing `workingDirectory`.

**`TmuxToolUnitSpec`** — 17 unit tests covering metadata, session name validation, missing required parameters per action, window parameter behaviour, and graceful handling when tmux is not installed.

## Build wiring

- `agentos/settings.gradle.kts` — `includeBuild("agentos-tmux-plugin")` added
- `agentos/build.gradle.kts` — `agentos-tmux-plugin` added to `pluginBuilds` so `./gradlew deployPlugins` copies its JAR into `plugins/`


cetautomatix-coday:
## Review

The plugin structure is clean and follows the datetime plugin pattern correctly. Build wiring (`settings.gradle.kts`, `build.gradle.kts`, `project.json`) is consistent with existing plugins. However there are several issues to address.

---

### `TmuxTool.kt` — Security: no timeout on `ProcessBuilder`

`runTmux()` calls `process.waitFor()` with no timeout. A hung or misbehaving tmux command (e.g. `send-keys` to a session that blocks) will block the calling thread indefinitely. This is a reliability/availability risk in a production-grade plugin. `process.waitFor(timeout, TimeUnit)` should be used with a reasonable bound (e.g. 10–30 s), and the process should be `destroyForcibly()`-ed on timeout.

---

### `TmuxTool.kt` — Security: `window` parameter not validated

The `session` parameter is validated against `VALID_SESSION_NAME` before any tmux command is issued. The `window` parameter is accepted without any validation and is directly interpolated into the tmux target string via `buildTarget()` (e.g. `"$session:$window"`). While `ProcessBuilder` prevents shell injection, a malformed or adversarial `window` value (e.g. containing `:`, newlines, or control characters) could cause unexpected tmux target resolution. The same `VALID_SESSION_NAME` regex (or a dedicated one allowing `.` for window names) should be applied to `window` when it is non-null.

---

### `TmuxTool.kt` — Code style: early returns in `execute()`

Per `agentos/docs/code-style.md`, early returns in the middle of a method should be avoided. The `execute()` method has two early returns (`if (input == null)` and `if (sessionError != null)`) before the `when` block. This violates the stated code style convention for AgentOS.

---

### `TmuxTool.kt` — Code style: chained `if/else` instead of `when`

Per `agentos/docs/code-style.md`, `when` is preferred over chained `if/else if` even for simple two-branch conditions. `buildTarget()` uses an `if/else` expression — minor, but inconsistent with the stated style. Same for the `if (configName != null)` expression in the `name` property.

---

### `TmuxPlugin.kt` — Missing `jar` manifest attributes

The plugin JAR must declare `Plugin-Id`, `Plugin-Version`, `Plugin-Provider`, and `Plugin-Class` in its manifest for PF4J to load it (as documented in `agentos-sdk/README.md` and implemented in the examples). `build.gradle.kts` has no `tasks.jar { manifest { attributes(...) } }` block. Without this the plugin JAR will not be recognized by PF4J at runtime.

---

### `TmuxToolUnitSpec.kt` — Test relies on live tmux behaviour

The test `"execute should accept session names with hyphens and underscores"` calls `tool.execute(TmuxTool.Input(action = "status", session = "my-service_01"))` which actually invokes `ProcessBuilder("tmux", ...)`. On CI environments without tmux this may behave differently from the stated intent (the comment says "just check no validation error"). This makes the test environment-dependent. The test should be restructured to assert only the validation path without invoking the real process, or the tmux invocation should be abstracted/mockable.

Similarly, `"list should return a success response regardless of whether tmux is installed"` asserts `"success":true` — but if tmux is not installed and the `runTmux` call throws an `IOException` (command not found), `runCatching` will catch it and `executeList()` will return a success response (by design). This is fine for unit testing the fallback, but the test comment is misleading about what is actually being tested.

cetautomatix-coday:
## Follow-up Review — All Six Issues Addressed ✅

All six items from the previous review have been correctly resolved. One minor new issue is noted below.

---

### Previous issues — verification

| # | Issue | Status |
|---|---|---|
| 1 | No timeout on `ProcessBuilder` | ✅ Fixed — `process.waitFor(TMUX_TIMEOUT_SECONDS, TimeUnit.SECONDS)` with `destroyForcibly()` on expiry |
| 2 | `window` not validated | ✅ Fixed — `VALID_WINDOW_NAME` regex validated in the `when` guard at the top of `execute()` |
| 3 | Early returns in `execute()` | ✅ Fixed — single `return when(validationError) { … }` expression, no mid-method early returns |
| 4 | `if/else` instead of `when` | ✅ Fixed — `name`, `buildTarget()`, and all action dispatchers now use `when` |
| 5 | Missing JAR manifest attributes | ✅ Fixed — `Plugin-Id`, `Plugin-Version`, `Plugin-Provider`, `Plugin-Class` all present in `tasks.jar` |
| 6 | Real `ProcessBuilder` in tests / misleading comments | ✅ Fixed — validation-only tests never reach `ProcessBuilder`; the `list` test is explicitly documented as a graceful-fallback test; the misleading comment on the previous `list` test is gone |

---

### New issue

**`TmuxTool.kt` — stream read after `waitFor`, before `exitValue`**

In `runTmux`, after the process has been waited on, the output stream is read **before** checking the exit code:

```kotlin
val output = process.inputStream.bufferedReader().readText()
val exitCode = process.exitValue()
if (exitCode != 0) throw RuntimeException(output.trim()…)
```

Reading `inputStream` after `waitFor` is safe when `redirectErrorStream(true)` is set (which it is here), so this is not a correctness bug. However, when the exit code is non-zero the full stdout is read into memory only to be used as an error message, which is wasteful for commands that produce large output on failure. This is a minor style point, not a blocker.

---

### Overall assessment

The code is correct, well-structured, and consistent with AgentOS code-style rules. The test suite covers validation exhaustively without touching real processes. **Ready to merge.**

agentos/agentos-tmux-plugin/src/main/kotlin/io/whozoss/agentos/plugins/tmux/TmuxTool.kt
```diff
@@ -0,0 +1,260 @@
...
+
+    override val version: String = "1.0.0"
+
+    override val paramType: Class<Input> = Input::class.java
+
+    // language=JSON
+    override val inputSchema: String =
+        """
+        {
+            "${"\$"}schema": "https://json-schema.org/draft/2020-12/schema",
```
vincent-couturier-whoz: suggestion: use `$$"""` so you can user just `$schema`

agentos/agentos-tmux-plugin/src/main/kotlin/io/whozoss/agentos/plugins/tmux/TmuxTool.kt
```diff
@@ -0,0 +1,260 @@
...
+     * Uses [ProcessBuilder] to avoid shell injection.
+     * Enforces a [TMUX_TIMEOUT_SECONDS]-second timeout; kills the process forcibly on expiry.
+     */
+    private fun runTmux(vararg args: String): Result<String> =
+        runCatching {
+            val process =
+                ProcessBuilder("tmux", *args)
+                    .redirectErrorStream(true)
+                    .start()
+            val timedOut = !process.waitFor(TMUX_TIMEOUT_SECONDS, TimeUnit.SECONDS)
```
vincent-audibert-whoz, line 328: Not suspendable ?

agentos/agentos-tmux-plugin/src/main/kotlin/io/whozoss/agentos/plugins/tmux/TmuxTool.kt
```diff
@@ -0,0 +1,260 @@
...
+        val validationError = when {
+            input == null -> "Input is required"
+            input.session != null && !VALID_SESSION_NAME.matches(input.session) ->
+                "Invalid session name '${input.session}'. Only alphanumeric characters, hyphens and underscores are allowed."
+            input.window != null && !VALID_WINDOW_NAME.matches(input.window) ->
+                "Invalid window name '${input.window}'. Only alphanumeric characters, hyphens, underscores, dots and plus signs are allowed."
+            else -> null
+        }
+        return when (validationError) {
+            null -> when (input!!.action) {
```
vincent-couturier-whoz: suggestion: you could replace advantageously the when by an enum class and anonymous classes and a method execute which take the input as argument (see https://kotlinlang.org/docs/enum-classes.html#anonymous-classes)

vincent-couturier-whoz (dismissed):

cetautomatix-coday (commented):
Applied in 256a4e59. The `when` dispatch on the action string has been replaced by a private enum `Action` with anonymous classes, each overriding `execute(input: Input, tool: TmuxTool): String`. The `fromString` companion resolves the action by name (case-insensitive), keeping the unknown-action error path. Members accessed from the enum classes (`runTmux`, `buildTarget`, `createSuccessResponse`, `createErrorResponse`, `workingDirectory`) had to be widened from `private` to `internal` since Kotlin nested classes cannot access private members of their enclosing class.

vincent-audibert-whoz:
Agent review of the tool by a claudenthusiastic agent:

## 📋 Résumé des tests effectués 

### ✅ **Ce qui fonctionne très bien :**
1. **Gestion des sessions** 
- Création de session : ✓ Fonctionne parfaitement 
- Listage des sessions : ✓ Affiche toutes les sessions actives 
- Vérification du statut : ✓ Retourne correctement "running" 
- Arrêt/suppression de session : ✓ Tue la session et toutes ses fenêtres 2

2. **Gestion des fenêtres** 
- Création de nouvelles fenêtres : ✓ Marche bien (j'ai créé window_1 et window_2) 
- Identification par index (0, 1, 2) : ✓ Fonctionne 
- Identification par nom : ✓ Fonctionne (window_1, window_2) 
- Fermeture de fenêtres spécifiques : ✓ Fonctionne (j'ai fermé window_1) 

3. **Temporisation (Wait)** 
- Attentes de 1 à 5 secondes : ✓ Précises et fiables 
- Permet de synchroniser les opérations asynchrones : ✓ Excellent 

4. **Logs et suivi** 
- `logs` sans fenêtre spécifiée : ✓ Affiche les logs de la fenêtre active 
- `logs` avec fenêtre spécifiée : ✓ Affiche exactement les logs de la fenêtre 
- Les timestamps et progressions sont capturées correctement 

5. **Envoi de commandes** 
- `send` avec index de fenêtre : ✓ Fonctionne parfaitement 
- Les commandes s'exécutent et les résultats s'affichent 

### ⚠️ **Détails intéressants observés :** 
- **Timing prévisible** : Les commandes avec `sleep` s'exécutent exactement comme prévu. Les attentes entre les lectures de logs permettent de capturer les résultats progressifs. 
- **Exécution parallèle** : Les fenêtres fonctionnent vraiment indépendamment. J'ai pu voir les messages s'accumuler au fur et à mesure. 
- **Formatage des logs** : Les logs incluent les commandes exécutées, l'invite shell, et avec beaucoup de lignes blanches de padding (apparemment normal). 

### 🎯 **Cas d'usage validés :** 
- ✓ Démarrer un serveur/processus long dans une fenêtre et le monitorer 
- ✓ Créer plusieurs fenêtres pour des processus parallèles 
- ✓ Envoyer des commandes à une fenêtre spécifique 
- ✓ Attendre entre les opérations pour la synchronisation 
- ✓ Consulter les logs à tout moment 
- ✓ Gérer le cycle de vie complet (création → exécution → arrêt) 

**Conclusion** : L'outil Tmux est très robuste et fiable pour orchestrer des processus long-running !

agentos/agentos-tmux-plugin/src/main/kotlin/io/whozoss/agentos/plugins/tmux/WaitTool.kt
```diff
@@ -0,0 +1,91 @@
...
+    data class Input(
+        val seconds: Int,
+    )
+
+    override fun execute(input: Input?): String {
+        if (input == null) return createErrorResponse("Input is required")
+        if (input.seconds < 1 || input.seconds > MAX_WAIT_SECONDS) {
+            return createErrorResponse("seconds must be between 1 and $MAX_WAIT_SECONDS, got ${input.seconds}")
+        }
+        Thread.sleep(input.seconds * 1000L)
```
vincent-couturier-whoz, line 72: issue(non-blocking): it seems not compliant with our strategy to avoid blocking calls. If there's a way to transform this call in a coroutine delay it would be better. It's not blocking because essentially limited to tmux use so most of the time a local use but if it's a tool we want to deploy centrally we should consider this issue as a blocking one.

vincent-couturier-whoz (approved):

vincent-couturier-whoz (approved):

PR #786, opened 2026-04-21
feat: #31170 bash plugin - exposes configured bash commands as tools

cetautomatix-coday:
## What

New `agentos-bash-plugin` that exposes configured bash commands as individual AgentOS tools, following the same pattern as `agentos-datetime-plugin` and `agentos-file-plugin`.

## Design

Each integration config declares a `workingDirectory` and a list of tool definitions. For each tool, `BashToolProvider` creates a distinct `BashTool` registered in the AgentOS tool registry.

Two modes per tool:
- **Fixed command** — no `PARAMETERS` placeholder, the tool takes no input and runs the command as-is
- **Parameterised command** — command contains the literal token `PARAMETERS`, which is substituted at runtime by the value the LLM supplies. `parametersDescription` is required and becomes the field description in the JSON schema.

Raw bash mode (the entire command is `PARAMETERS`) is supported for maximum flexibility, with the expectation that access control is handled at the permissions layer.

## Config schema

```json
{
  "workingDirectory": "/absolute/path",
  "defaultTimeoutSeconds": 30,
  "tools": [
    { "name": "run_tests", "description": "Runs the test suite", "command": "./gradlew test" },
    { "name": "search", "description": "Search codebase", "command": "grep -r PARAMETERS src", "parametersDescription": "The pattern to search for" },
    { "name": "bash", "description": "Raw bash (use with care)", "command": "PARAMETERS", "parametersDescription": "The full bash command to execute" }
  ]
}
```

Per-tool optional fields: `path` (subdirectory relative to `workingDirectory`), `timeoutSeconds` (overrides the integration default).

## Implementation

- `BashToolConfig` / `BashIntegrationConfig` — config data classes + `PARAMETERS_PLACEHOLDER` constant
- `BashConfigParser` — validates the JSON config with descriptive error messages (duplicate names, missing `parametersDescription`, mismatched placeholder, non-positive timeouts…)
- `BashCommandExecutor` — runs `/bin/bash -c <command>` with async stdout/stderr drain (same anti-deadlock pattern as ripgrep in `SearchFilesTool`) and `destroyForcibly` in `finally`
- `BashTool` — `StandardTool` implementation; generates the correct `inputSchema` depending on mode
- `BashPlugin` / `BashToolProvider` — PF4J plugin entry point; invalid config logs an error and returns an empty list rather than crashing the service

## Tests (57 total, all passing)

- `BashConfigParserUnitSpec` — 27 cases covering all validation rules
- `BashToolUnitSpec` — 20 cases including real process execution (echo, pwd, sleep, true, stderr)
- `BashToolProviderUnitSpec` — 13 cases covering null/invalid/valid configs and schema wiring

## Gradle wiring

- `agentos/settings.gradle.kts` — `includeBuild("agentos-bash-plugin")`
- `agentos/build.gradle.kts` — `"agentos-bash-plugin"` added to `pluginBuilds` so `deployPlugins` copies the JAR into `plugins/`
- `agentos-bash-plugin/project.json` — Nx `build`, `test`, `clean` targets

Closes #31170


cetautomatix-coday:
## Code Review — `agentos-bash-plugin`

The overall structure is solid and consistent with the existing plugin pattern. Three issues need to be addressed before merge: one security concern, one correctness bug in `BashCommandExecutor`, and one Kotlin style violation from `code-style.md`. Several smaller observations follow.

---

### 🔴 Security — Unvalidated `workingDirectory` and `path` allow filesystem escape

**`BashToolConfig.kt` / `BashConfigParser.kt` / `BashTool.kt`**

`workingDirectory` is accepted as any string and `path` is appended to it with `File(base, toolConfig.path)` without any canonicalization or boundary check. Two concrete problems:

1. A `path` value of `../../etc` resolves to a directory outside the intended working tree. `BashConfigParser` validates that `path` is non-blank when present but does nothing else with it.
2. `workingDirectory` itself is never verified to exist or to be a directory at parse time. The executor will fail at `ProcessBuilder.start()` with an opaque `IOException` rather than a clear config error.

The file plugin addresses this class of problem with `BoundaryPathResolver` — the same defense is needed here. At minimum: (a) resolve `workingDirectory` to a canonical path at parse time and assert it is an existing directory; (b) resolve `File(base, path).canonicalPath` and assert it starts with `base.canonicalPath` before executing.

---

### 🔴 Correctness — `destroyForcibly` in `finally` kills every successfully-completed process

**`BashCommandExecutor.kt`**

```kotlin
} finally {
    process?.destroyForcibly()
}
```

`destroyForcibly` is called unconditionally in the `finally` block, even on the happy path where the process already exited. On Linux, `destroyForcibly` on a process that has already terminated is a no-op (the PID is gone), so in practice it does not corrupt results. However, it is logically wrong and fragile: on some JVM implementations `exitValue()` can throw `IllegalThreadStateException` if called after `destroyForcibly` races with the process exit. The correct pattern is to call `destroyForcibly` only in the timeout branch (which already does so) and in the `catch` block, not in `finally`. The `finally` block should only be reached from the `Completed` or `Error` paths and should not re-destroy an already-dead process.

---

### 🔴 Kotlin style — early return in `execute()`

**`BashTool.kt`** — `code-style.md` states: *"Avoid early returns in the middle of a method"*.

```kotlin
val params = input?.parameters?.takeIf { it.isNotBlank() }
    ?: return "Error: tool '${toolConfig.name}' requires a 'parameters' value but none was provided"
```

This is an early return in the middle of `execute()`. Restructure using `when` or an expression body so the single return is at the end of the function.

---

### 🟡 Correctness — `stdoutFuture`/`stderrFuture` not cancelled on timeout

**`BashCommandExecutor.kt`**

When `waitFor` returns `false` (timeout), `destroyForcibly` is called and `BashExecutionResult.Timeout` is returned — but `stdoutFuture` and `stderrFuture` are still running, draining the (now closed) streams in background threads from the common ForkJoinPool. The futures are never cancelled. For a single-command tool this leaks two threads until the streams drain. Add `stdoutFuture.cancel(true); stderrFuture.cancel(true)` before returning the `Timeout` result.

---

### 🟡 Correctness — `BashConfigException` is declared but never thrown

**`BashConfigParser.kt`**

```kotlin
class BashConfigException(message: String) : IllegalArgumentException(message)
```

All `require()` calls throw plain `IllegalArgumentException`, not `BashConfigException`. The class exists but is dead code. Either use it consistently (replace `require` with `if (!condition) throw BashConfigException(...)`) or remove it. The catch in `BashPlugin.kt` catches `IllegalArgumentException` which is the actual supertype, so the current code works, but the declared exception class is misleading.

---

### 🟡 Security — `PARAMETERS` substitution is a simple string replace, not a shell-safe quoting

**`BashTool.kt`**

```kotlin
toolConfig.command.replace(PARAMETERS_PLACEHOLDER, params)
```

This is documented in the PR description as intentional ("raw bash mode"), but the risk is not surfaced to the operator in the config schema or tool description. When `PARAMETERS` is the entire command, the LLM can execute arbitrary shell code. When it appears as a fragment (e.g. `grep -r PARAMETERS src`), the LLM can break out of the intended command with shell metacharacters (e.g. `.; rm -rf /`). The PR description says "access control is handled at the permissions layer", but that layer (PR #756) is not yet merged. Until it is, the documentation/schema should make the injection risk explicit — especially the raw bash mode — so operators can make an informed choice. A comment or `description` field in the config schema noting the shell injection risk would be sufficient.

---

### 🟢 Minor — `require` vs `BashConfigException` inconsistency in error message contract

**`BashConfigParser.kt`**

The `require()` calls throw `IllegalArgumentException` whose message is always non-null here, but callers rely on `e.message` being non-null (`logger.error { "... — ${e.message}" }`). This is safe in practice because `require` always populates the message, but it would be cleaner to use the named exception type so the contract is explicit.

---

### 🟢 Minor — `workingDirectory` not validated to be absolute

**`BashConfigParser.kt`**

The config doc and schema both say "Absolute path", but the parser only checks non-blank. A relative path silently resolves against the JVM working directory (the service's CWD), which is unpredictable in a containerised deployment. Add `require(File(workingDirectory).isAbsolute)` at parse time.

---

### Test coverage

The 57 tests cover the contract well. One gap: there is no test for the case where `workingDirectory` does not exist on disk — currently this surfaces as a generic `BashExecutionResult.Error` rather than a config-time error, which is another argument for the filesystem validation noted above.

cetautomatix-coday:
All review comments addressed in the follow-up commit:

**🔴 Security — path traversal via `path` and `workingDirectory`**
`BashConfigParser` now validates both fields:
- `workingDirectory` must be an absolute path (checked via `File.isAbsolute`)
- `path`, when set, must be relative and must not escape `workingDirectory` — validated by canonicalizing `File(workingDirectory, path)` and checking it starts with the canonical base. This covers `../../etc`-style traversal without requiring the directory to exist at parse time.

**🔴 Correctness — `destroyForcibly` on happy path**
`BashCommandExecutor` restructured with `when` on the `completed` boolean. `destroyForcibly` is now called only in the timeout branch and in the error catch block after `waitFor`. On the normal completion path the process has already exited cleanly, so it is left alone.

**🔴 Kotlin style — early return in `execute()`**
Extracted a `resolveCommand(input): String?` private method that returns `null` when parameters are missing. `execute()` now uses the result of that call with a single `?: return` at the top, keeping the rest of the method linear. The logic is cleaner and the early-return is confined to the guard at the entry point.

**Non-blocking — `stdoutFuture`/`stderrFuture` not cancelled on timeout**
Both futures are now cancelled before `destroyForcibly()` on the timeout path. The futures may still block briefly on `stream.read()` until the OS closes the pipe after the process is killed, but they are no longer leaked indefinitely.

**Non-blocking — `BashConfigException` dead code**
Removed. The class was never thrown; validation uses `require()` which throws `IllegalArgumentException` directly.

**Non-blocking — `workingDirectory` not validated as absolute**
Now validated in `BashConfigParser.parse()` with a clear error message. Test coverage added in `BashConfigParserUnitSpec` for relative path, path traversal, and absolute path in `path` field.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashCommandExecutor.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun execute(
+        command: String,
+        workingDirectory: File,
+        timeoutSeconds: Long,
+        maxOutputChars: Int = DEFAULT_MAX_OUTPUT_CHARS,
+    ): BashExecutionResult {
+        logger.debug { "Running bash command in ${workingDirectory.absolutePath}: $command" }
+
+        val process = try {
+            ProcessBuilder("/bin/bash", "-c", command)
```
vincent-couturier-whoz, line 32: suggestion: It seems we've got similar needs to execute external process in tmux plugins. Perhaps we could have a common wrapper to manage "cleanly" and "securely" the execution of external process

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
```diff
@@ -0,0 +1,131 @@
...
+    fun parse(config: JsonNode): BashIntegrationConfig {
+        val workingDirectory = config.get("workingDirectory")?.asText()?.trim()
+        require(!workingDirectory.isNullOrBlank()) {
+            "BASH integration config: 'workingDirectory' is required and must not be blank"
+        }
+        require(File(workingDirectory).isAbsolute) {
+            "BASH integration config: 'workingDirectory' must be an absolute path, got: $workingDirectory"
+        }
+
+        val defaultTimeout = config.get("defaultTimeoutSeconds")?.asLong()?.also { timeout ->
```
vincent-couturier-whoz, line 35: suggestion: we should add a constant to avoid maintaining this ligne and le error message below (inside the require statement)

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
```diff
@@ -0,0 +1,131 @@
...
+            require(timeout > 0) {
+                "BASH integration config: 'defaultTimeoutSeconds' must be positive, got $timeout"
+            }
+        } ?: 30L
```
vincent-couturier-whoz: suggestion: transform this magical number into a constant

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
```diff
@@ -0,0 +1,131 @@
...
+
+        val toolsNode = config.get("tools")
```
vincent-couturier-whoz, line 41: suggestion: Perhaps this tools filtering should be externalized because it seems a common requirement in lots of integrations

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
vincent-couturier-whoz, line 1: thought: I realize that we should/could replace bash tooling by ssh tooling in order to be able to call process external to the agentos hosting and so avoid the risk of destabilize agentos

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashPlugin.kt
```diff
@@ -0,0 +1,137 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import io.whozoss.agentos.sdk.tool.StandardTool
+import io.whozoss.agentos.sdk.tool.ToolPlugin
+import mu.KLogging
+import org.pf4j.Extension
+import org.pf4j.Plugin
+
+class BashPlugin : Plugin() {
```
vincent-couturier-whoz, line 11: Perhaps we could have a `DefautPluginImplementation` with a name as parameter to avoid boilerplate like the start/stop method

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashPlugin.kt
```diff
@@ -0,0 +1,137 @@
...
+            logger.error { "BASH integration '$configName': invalid config — ${e.message}" }
+            return emptyList()
+        }
+
+        if (integrationConfig.tools.isEmpty()) {
+            logger.warn { "BASH integration '$configName': no tools defined in config" }
+            return emptyList()
+        }
+
+        return integrationConfig.tools.map { toolConfig ->
```
vincent-couturier-whoz, line 58: suggestion: perhaps we could have a enableAllTools to be able to avoid listing all tools.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
```diff
@@ -0,0 +1,131 @@
...
+        val tools = when {
+            toolsNode == null || toolsNode.isNull || !toolsNode.isArray -> emptyList()
+            else -> toolsNode.mapIndexed { index, toolNode ->
+                parseTool(toolNode, index, workingDirectory)
+            }
+        }
+
+        val duplicates = tools.map { it.name }.groupBy { it }.filter { it.value.size > 1 }.keys
+        require(duplicates.isEmpty()) {
```
vincent-couturier-whoz, line 50: question : what's the effect of duplicate tools ? I don't know what the result will be in case of wrong configuration

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashTool.kt
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * The `inputSchema` is generated accordingly so the LLM sees the right contract.
+ */
+class BashTool(
+    private val toolConfig: BashToolConfig,
+    private val integrationConfig: BashIntegrationConfig,
+    configName: String? = null,
+) : StandardTool<BashTool.Input> {
+
+    companion object : KLogging() {
```
vincent-couturier-whoz: suggestion : put the companion object at the end of the class

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashTool.kt
```diff
@@ -0,0 +1,128 @@
...
+        private val objectMapper = jacksonObjectMapper()
+    }
+
+    private val isParameterised: Boolean = toolConfig.command.contains(PARAMETERS_PLACEHOLDER)
+
+    override val name: String = if (configName != null) "${configName}__${toolConfig.name}" else toolConfig.name
```
vincent-couturier-whoz, line 29: suggestion: add a common functions in agentOS SDK in order to have coherent naming everywhere.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashTool.kt
```diff
@@ -0,0 +1,128 @@
...
+                    "parameters": {
+                        "type": "string",
+                        "description": ${objectMapper.writeValueAsString(toolConfig.parametersDescription)}
+                    }
+                },
+                "required": ["parameters"],
+                "additionalProperties": false
+            }
+        """.trimIndent()
+        else -> """
```
vincent-couturier-whoz: `$$"""` in order to be able to use directly `"$schema"`

agentos/agentos-bash-plugin/src/test/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParserUnitSpec.kt
```diff
@@ -0,0 +1,359 @@
...
+                "tools": [{
+                    "name": "t",
+                    "description": "desc",
+                    "command": "ls",
+                    "path": "../../etc"
+                }]
+            }
+        """)
+
+        val ex = shouldThrow<IllegalArgumentException> { BashConfigParser.parse(config) }
```
vincent-couturier-whoz, line 306: suggestion: perhaps we should validate config at the moment integration config is saved or retrieved from databas and so the plugin to  could be indicated as "wrong" configuration

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelNodeNeo4jRepository.kt
```diff
@@ -49,8 +49,8 @@ interface AiModelNodeNeo4jRepository : Neo4jRepository<AiModelNode, String> {
      * Find the first non-removed model config under [aiProviderId] whose alias matches exactly.
      */
     @Query(
-        $$"""MATCH (m:AiModel)
-            WHERE m.aiProviderId = $aiProviderId AND m.alias = ${GqlParams.StringParam.alias}
+        $"""MATCH (m:AiModel)
+            WHERE m.aiProviderId = $aiProviderId AND m.alias = $alias
```
vincent-couturier-whoz, line 53: praise: ❤️

vincent-couturier-whoz (dismissed):

cetautomatix-coday:
Retours de Vincent Couturier adressés dans le commit `bb2549ad` :

**Fait :**
- `DEFAULT_TIMEOUT_SECONDS = 30L` extrait en constante dans `BashToolConfig.kt`, utilisé dans le default de `BashIntegrationConfig` et dans le fallback de `BashConfigParser` — le magic number `30L` n'apparaît plus qu'à un seul endroit.
- `companion object` déplacé à la fin de `BashTool` (convention Kotlin).

**Non fait — suggestion `$$"""` (multi-dollar interpolation) :**
La syntaxe `$$"""..."""` requiert le flag de compilateur `-Xmulti-dollar-interpolation`, non activé dans le build actuel. L'activation mériterait une décision consciente (à faire dans une issue séparée si souhaité). L'échappement `${'$'}` existant est conservé en attendant.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashCommandExecutor.kt
vincent-audibert-whoz: Whole epic by itself, we already have a ticket on that, no ? Could not find it in Jira, as usual...

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
vincent-audibert-whoz: Integrations expose a tool set or list, that will have to be filtered depending on agent integration configuration, doing that could be placed inside the plugin but it then becomes the plugin responsability to handle the filtering. In the end, I would rather have the tool filtering outside of the plugin

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
vincent-audibert-whoz: Same epic as the bash topic => for later.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashPlugin.kt
vincent-audibert-whoz: WZ-31549

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashPlugin.kt
vincent-audibert-whoz: I don’t get your point: integrationConfig is coming from the outside of the plugin and tells what config of tools are to be converted into standardTools ?

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParser.kt
vincent-audibert-whoz: 400 on AI provider side. Should be standard plugin implementation, added to WZ-31549.

agentos/agentos-bash-plugin/src/main/kotlin/io/whozoss/agentos/plugins/bash/BashTool.kt
vincent-audibert-whoz: added to WZ-31549.

agentos/agentos-bash-plugin/src/test/kotlin/io/whozoss/agentos/plugins/bash/BashConfigParserUnitSpec.kt
vincent-audibert-whoz: added to WZ-31549

vincent-couturier-whoz (approved):

Issue #788, opened 2026-04-22
fix: case chat view does not scroll vertically


cetautomatix-coday:
### Description

The case chat view does not scroll vertically, making it impossible to read older messages when the conversation grows beyond the visible viewport.

### Steps to Reproduce

1. Open a case with enough messages to overflow the chat view
2. Observe that the view does not scroll — messages above the fold are inaccessible

### Expected Behavior

The chat view scrolls vertically, allowing the user to navigate the full message history.

### Actual Behavior

The chat view is locked / overflow is hidden, messages outside the viewport are unreachable.

### Labels
`bug`, `frontend`

PR #791, opened 2026-04-22
fix: #788 case-chat vertical scroll with magnetic auto-scroll and floating button

cetautomatix-coday:
## Problem

The `CaseChatComponent` did not scroll vertically as the conversation grew. The composer (input + actions) was not guaranteed to stay visible either.

## Root cause

The `router-outlet` inserts the component's host element into the DOM without inheriting flex constraints from the parent. Without `:host` styles, `.case-chat` grew unboundedly instead of being constrained by the viewport chain.

### Phase 1 — Scroll layout fix

- **`:host`** gets `display: flex; flex-direction: column; flex: 1; min-height: 0; overflow: hidden` — plugs the component into the `case-shell__content` flex chain
- **`.case-chat`** gains `flex: 1; min-height: 0`
- **`.case-chat__messages`** gains `min-height: 0` (critical: without it, flex prevents the element from shrinking below its content height)
- **`.case-chat__composer`** gets `flex-shrink: 0` — always fully visible, never compressed

### Phase 2 — Magnetic auto-scroll + floating button

- **`isAtBottom` signal** (default `true`): updated by a passive `scroll` listener on the messages container. Threshold: 64px from the bottom.
- **Auto-scroll effect**: reacts to `timeline()` and `streamingText()` changes. Scrolls to bottom via `queueMicrotask` (after DOM update) only when `isAtBottom()` is true.
- **Magnet break**: scrolling up past the threshold sets `isAtBottom` to `false` — auto-scroll stops.
- **Floating button**: `ds-icon-button` with `keyboard_arrow_down`, absolutely positioned above the composer, visible only when `!isAtBottom()`. Clicking it re-enables the magnet and scrolls to bottom. Appears with a subtle fade + slide animation.

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
+++ b/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
@@ -1,5 +1,5 @@
 <div class="case-chat">
-  <div class="case-chat__messages">
+  <div #messagesContainer class="case-chat__messages">
     @if (timeline().length === 0) {
       <div class="case-chat__empty">
         <p>No messages yet.</p>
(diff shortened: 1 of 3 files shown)


vincent-audibert-whoz (approved):
tested ok for front rendering

PR #796, opened 2026-04-22
feat: unified npm & JAR release pipeline via Nx

mathieu-dearmey:
## What

Extends the Nx release pipeline to publish both npm packages and Gradle JARs in a single coordinated step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,7 +110,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.RELEASE_PAT }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
         run: |
-          pnpm nx release --yes
+          npx tsx scripts/release.ts
           # Capture the new version for desktop build from client package.json
           NEW_VERSION=$(node -p "require('./apps/client/package.json').version")
           echo "new_version=$NEW_VERSION" >> $GITHUB_OUTPUT
@@ -243,6 +243,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: master
 
       - name: Set up JDK 25
         uses: actions/setup-java@v4
@@ -272,4 +274,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_ACTOR: ${{ github.actor }}
+          SDK_VERSION: ${{ needs.release.outputs.new_version }}
         run: pnpm nx publish agentos-sdk
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -274,5 +274,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_ACTOR: ${{ github.actor }}
-          SDK_VERSION: ${{ needs.release.outputs.new_version }}
         run: pnpm nx publish agentos-sdk
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,6 +71,7 @@ jobs:
       id-token: write
     outputs:
       new_version: ${{ steps.release.outputs.new_version }}
+      release_sha: ${{ steps.release_sha.outputs.sha }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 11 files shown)


cetautomatix-coday:
## Code Review

The PR correctly implements its stated goal: the npm release script (`scripts/release.ts`) now also updates the Gradle version catalog (`agentosSdk` in `libs.versions.toml`) before the changelog commit, and a new `publish-sdk` job in the CI workflow calls `pnpm nx publish agentos-sdk` after the npm release. The `publish` target was already present on `master` in `agentos-sdk/project.json`, so the Nx wiring is sound. The `updateTomlVersion` utility is pure, well-tested, and the regex is correct.

There are two issues to address.

**1. `publish-sdk` job checks out `master` instead of the release commit**

In the current `publish-sdk` job:
```yaml
- name: Checkout repository
  uses: actions/checkout@v4
  with:
    ref: master
```
The `release` job commits, tags, and pushes the bumped `libs.versions.toml` to `master`. Because `publish-sdk` runs in parallel with `desktop-release` (both depend only on `release`), checking out `master` creates a race condition: if the release commit has not yet propagated to the remote `master` ref when `publish-sdk` starts, the job will build and publish the SDK at the **old version**. The `desktop-release` job avoids this race correctly with `ref: master` too, but it is also susceptible — however for the SDK the consequence is a published artifact with a stale version, which is a hard-to-debug artifact integrity issue. The job should pin to the exact commit SHA that was produced by the `release` job (e.g. via a `release_sha` output), or at minimum add an explicit `git pull` after checkout to guarantee the latest state.

**2. `scripts/release.ts` uses `__dirname` which is not available in ESM context**

```typescript
const tomlPath = join(__dirname, '..', tomlRelativePath)
```

`scripts/tsconfig.json` targets `CommonJS`, so `__dirname` works when the script is compiled with `tsc`. However, `scripts/release.ts` is invoked via `npx tsx scripts/release.ts` (from the workflow), and `tsx` by default executes TypeScript as ESM. In ESM, `__dirname` is undefined at runtime, causing the `join(__dirname, ...)` call to silently produce a wrong path (joining `undefined` with `..`) or throw. The `agentos/agentos-sdk/package.json` added in this PR declares `"type": "module"`, which is unrelated to the scripts directory, but the invocation method matters. This should be verified and, if ESM execution is confirmed, replaced with `fileURLToPath(new URL('..', import.meta.url))`.

**Minor: `agentos/agentos-sdk/package.json` is a phantom npm package**

```json
{
  "name": "@agentos/agentos-sdk",
  "main": "./src/index.js",
  "types": "./src/index.d.ts"
}
```

This file references `./src/index.js` and `./src/index.d.ts`, which do not exist (the SDK is a pure Kotlin/Gradle project with no TypeScript source). The `private: true` field prevents accidental npm publishing, but the file still misleads Nx's project graph and any tooling that inspects `package.json` entry points. If this file is only needed to register the project in the Nx workspace, the `main`/`types`/`dependencies`/`devDependencies` fields should be removed to avoid confusion.

agentos/gradle/libs.versions.toml
```diff
@@ -32,7 +32,7 @@ netty = "4.2.9.Final"
 testcontainers = "1.20.4"
 
 # AgentOS modules versions
-agentosSdk = "1.0.0"
+agentosSdk = "0.115.2"
```
vincent-audibert-whoz: We are currently at 0.125.3 for coday, should match the coday version ?

vincent-audibert-whoz (dismissed):

scripts/tsconfig.spec.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "module": "CommonJS",
```
vincent-couturier-whoz, line 3: question: why we use CommonJS here and not ESNext ?

vincent-couturier-whoz (dismissed):

scripts/tsconfig.spec.json
mathieu-dearmey: Jest doesn't handle ESNext. This is why we need a separate tsconfig file specifically for tests (in the jest.config we point to it in the "transform")

.github/workflows/release.yml
```diff
@@ -125,7 +130,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: master
+          ref: ${{ needs.release.outputs.release_sha }}
```
vincent-couturier-whoz, line 133: praise: well done. I hope there's no other places where we have same issue

vincent-couturier-whoz (approved):

vincent-audibert-whoz (approved):

.github/workflows/release.yml
mathieu-dearmey: yes, this was a good catch by Vincent A.'s reviewer 😉

PR #797, opened 2026-04-23
fix: #WZ-31209 maxTokens for OpenAI/Gemini and relax AiModel uniqueness to alias-only

vincent-audibert-whoz:
## Context

WZ-31209 — two bugs identified during local review of the namespace LLM config feature.

### `maxTokens` was silently ignored for OpenAI and Gemini

`ChatModelFactory` received `maxTokens` but never passed it to the OpenAI or Gemini options builders. For Gemini the builder method is `maxOutputTokens()` (not `maxTokens()`), which also caused a compilation error. Anthropic was already correct.

### AiModel uniqueness constraint relaxed to alias-only

The `(aiProviderId, apiName)` uniqueness constraint on `AiModelServiceImpl` has been removed. Only `(aiProviderId, alias)` is now enforced. This allows two `AiModel` entries with the same `apiModelName` under the same provider but with different inference parameters (e.g. different temperature or maxTokens) as long as their aliases differ — which is the intended use case.

Tests updated accordingly.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelServiceImpl.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelServiceImpl.kt
@@ -24,12 +24,9 @@ class AiModelServiceImpl(
     private val aiProviderService: AiProviderService,
 ) : AiModelService {
     override fun create(entity: AiModel): AiModel {
-        findByAiProviderAndApiName(entity.aiProviderId, entity.apiModelName)?.let {
-            throw ResponseStatusException(
-                HttpStatus.CONFLICT,
-                "A model config for apiName '${entity.apiModelName}' already exists in AiProvider ${entity.aiProviderId}",
-            )
-        }
+        // Uniqueness is enforced on alias only — two configs may share the same
+        // apiModelName under the same provider (e.g. same model with different
+        // temperature or maxTokens), as long as their aliases are distinct.
         entity.alias?.let { alias ->
             findByAiProviderAndAlias(entity.aiProviderId, alias)?.let {
                 throw ResponseStatusException(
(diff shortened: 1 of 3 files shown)


vincent-audibert-whoz (commented):
## Review — fix: #WZ-31209 maxTokens for OpenAI/Gemini and relax AiModel uniqueness to alias-only

Both fixes are correct in intent and the test suite is well-structured. Four issues to address before merge.

---

### `ChatModelFactory.kt` — dead default parameters on private functions

`createOpenAiModel` and `createGeminiModel` now declare `maxTokens: Int? = null` as a default on the private function signature, but `createChatModel` always passes the value explicitly (including `null`). The defaults are never exercised, they create a false impression that the parameter is independently optional at the private-function level. Remove the `= null` defaults from both private functions to make the signatures honest.

---

### `AiModelServiceImpl.kt` — stale KDoc contradicts the implementation

The class-level KDoc still lists:
```
* - (aiProviderId, apiName): a model can only be registered once per provider config
```
That constraint was removed by this PR. The comment now actively lies about the contract. Update the KDoc to describe only the remaining `(aiProviderId, alias)` constraint.

---

### `AiModelServiceImpl.kt` — null-alias bypass is implicit and undocumented

When `entity.alias` is `null`, the `entity.alias?.let { … }` block is skipped entirely, so any number of records with `null` alias under the same provider can be created without any uniqueness check. This may be intentional — null-alias models cannot be looked up by alias anyway — but the intent is invisible from the code. Either add a comment explaining the deliberate choice, or decide whether a `(aiProviderId, apiModelName)` guard should still apply when alias is null to prevent silent accumulation of phantom entries.

---

### `AiModelServiceImplSpec.kt` — misleading test name

The test `"update does not conflict with itself on apiName"` describes a self-conflict guard on `apiName` that no longer exists. The body only exercises a plain temperature update and says nothing about `apiName` uniqueness — it is a leftover name from the old constraint. Rename it (e.g. `"update persists field changes without spurious conflict"`) or merge with the adjacent generic update test.

vincent-couturier-whoz (approved):

PR #804, opened 2026-04-24
fix: skip publish jobs when no version bump was produced and include agentos projects in release

mathieu-dearmey:
Two fixes to the release pipeline:

**Skip downstream jobs when no release happened**
When a commit only touches `agentos/` files, Nx detects no version bump is needed
and exits early — but `publish-sdk` and `desktop-release` were still running,
causing a 409 conflict when trying to republish an already existing SDK version.
The release job now only outputs `new_version` when a release tag was actually
created (git tag check against HEAD), so downstream jobs skip automatically.

**Include JVM projects in release versioning**
Added `tag:platform:jvm` to a dedicated `jvm` release group in `nx.json` so that
changes to any JVM project (`agentos-sdk`, `agentos-service`, plugins, etc.)
correctly trigger a version bump alongside the npm release. New JVM projects are
automatically included as long as they carry the `platform:jvm` tag.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -71,7 +71,6 @@
     }
   ],
   "release": {
-    "projects": ["server", "client", "web", "desktop", "desktop-twin"],
     "releaseTagPattern": "release/{version}",
     "changelog": {
       "workspaceChangelog": {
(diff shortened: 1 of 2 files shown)


nx.json
```diff
@@ -85,14 +84,27 @@
...
       }
+    },
+    "groups": {
+      "npm": {
+        "projects": ["server", "client", "web", "desktop", "desktop-twin"],
+        "version": {
+          "manifestRootsToUpdate": ["apps/{projectName}"]
+        }
+      },
+      "jvm": {
```
vincent-couturier-whoz, line 102: remark: I discover we have defined tags on the projects.... thanks to coday 😄

vincent-couturier-whoz (approved):

PR #805, opened 2026-04-24
feat(agentos): #WZ-29564 namespace management (Epic 2)

selimbensenouciep-prog:
## Summary

Implements Epic 2 of the AgentOS permissions system: full namespace management on top of the foundation laid by Epic 1. **Stacked on top of #803 (Epic 1 PR)** — merge Epic 1 first, then this retargets to master.

## Stories covered

- **2.1 Super-admin namespace CRUD** — `NamespaceController` migrates to `SecuredEntityController`. CREATE/DELETE are restricted to super-admin callers. The creator automatically receives an ADMIN relation on the new namespace. DELETE cascade-revokes all namespace permission relations (ADMIN + MEMBER) before the soft delete; a failure to enumerate affected users aborts the whole operation so no orphan relations remain.
- **2.2 Admin management** — `PUT/DELETE /api/namespaces/{id}/admins/{userId}` in a dedicated `NamespacePermissionEndpoints` controller. Caller must hold ADMIN on the namespace (or be super-admin). Idempotent via Neo4j `MERGE` / `DELETE` on absent pattern.
- **2.3 Member management** — `PUT/DELETE /api/namespaces/{id}/members/{userId}` with the same authorization model. Revoking MEMBER leaves any concurrent ADMIN relation intact — verified end-to-end by a dedicated Neo4j integration test that also proves the MEMBER relation is actually removed (not a no-op).
- **2.4 Permission-filtered namespace listing** — `GET /api/namespaces` now returns only the namespaces the caller has at least READ on, each carrying the caller's effective `role` (`SUPER-ADMIN` / `ADMIN` / `MEMBER`). Implementation uses `listEntitiesForUser(READ)` + `listEntitiesForUser(WRITE)` + `findByIds` → constant number of permission-service calls regardless of result size (no N+1 on `hasPermission`).
- **2.5 Namespace user listing** — `GET /api/namespaces/{id}/users` lists users with a direct `ADMIN` / `MEMBER` relation on the namespace. Users holding both relations appear once with `role = \"ADMIN\"`. Orphan relations (pointing to deleted users) are filtered silently with a WARN log; malformed UUIDs from Neo4j are logged and skipped defensively.

## Security model

- All namespace-scoped endpoints follow the architectural rule *\"always 404 if no READ\"*: a caller with no READ on a namespace sees 404 regardless of whether the namespace exists, hiding its existence from non-members.
[...]

PR #803, opened 2026-04-24
feat(agentos): #WZ-29564 implement permissions system (Epic 1)

selimbensenouciep-prog:
## Context

WZ-29564 — AgentOS needs a permissions system to control access to entities (namespaces, cases, agent configs, integration configs, AI providers, AI models). This is Epic 1 of the permissions v2 roadmap, covering the core infrastructure: permission service, bootstrap, Neo4j relations, transitive evaluation, identity resolution, and controller integration.

### Permission Service (Story 1.1)
- `PermissionService` interface with fail-closed default behavior: denies access when no explicit permission exists
- Super-admin bypass: users with `isAdmin=true` skip all permission checks
- Caffeine cache layer with 5-minute TTL and configurable max size (10,000 entries)
- Cache key format: `perm:${userId}:${entityType}:${entityId}:${action}`

### Bootstrap Service (Story 1.2)
- `BootstrapServiceImpl` as `ApplicationRunner`: idempotent startup bootstrap
- First user auto-promoted to super-admin via `resolveOrCreateByExternalId()`
- Configurable via `agentos.bootstrap.enabled` property

### Neo4j Permission Relations (Story 1.3)
- `PermissionNodeNeo4jRepository` with `@Query` annotations following Spring Data Neo4j pattern
- `Neo4jPermissionRepository` bridge implementing domain `PermissionRepository`
- Separate type-safe methods for ADMIN and MEMBER relations (no parameterized Cypher relationship types)
- Proper label matching with `IN labels()` for multi-label node compatibility

### Transitive Permission Evaluation (Story 1.4)
- ADMIN on namespace → full access (READ/WRITE/DELETE) to all child entities
- MEMBER on namespace → READ-only access to child entities
- Child entity types: Case, AgentConfig, IntegrationConfig, AiProvider, AiModel
- Direct permissions take precedence over transitive permissions

### User Identity Resolution (Story 1.5)
- Resolve user identity from local OS username (local mode) or Cloudflare JWT email (auth mode)
- Auto-create user on first access with `resolveOrCreateByExternalId()`
- First user gets `isAdmin=true`, subsequent users get `isAdmin=false`
- Email extraction from externalId for display purposes

### Controller Integration (Story 1.6)
- `SecuredEntityController` abstract class with permission-filtered CRUD operations
- READ denied → 404 (hides entity existence), WRITE/DELETE denied → 403
- `checkCreatePermission()` defaults to deny (fail-closed)
[...]

PR #807, opened 2026-04-24
feat(agentos): #WZ-29564 case lifecycle and permissions (Epic 3)

selimbensenouciep-prog:
## Summary

Implements Epic 3 of the AgentOS permissions system: case creation, per-case access control, and owner-private visibility rules. **Stacked on top of #805 (Epic 2 PR)** — merge Epic 1 then Epic 2 first, then this retargets to master.

## Stories covered

- **3.1 Case creation with auto-admin** — `CaseController.checkCreatePermission` is relaxed from `Action.WRITE` to `Action.READ` so namespace MEMBERs can create cases (FR11). `create()` auto-grants `[:ADMIN]` on the new case to the creator (best-effort, non-transactional — same pattern as Story 2.1).
- **3.2 Namespace ADMIN access to cases** — `listByParent` short-circuits when the caller holds ADMIN on the parent namespace: one query instead of N+1 `hasPermission` calls per case. A new Neo4j integration test exercises the full transitive-admin path end-to-end.
- **3.3 Member limited to own cases (FR15)** — Case is declared owner-private in `Neo4jPermissionRepository.OWNER_PRIVATE_ENTITY_TYPES`. Case+MEMBER is dispatched to `hasAdminAccessViaNamespace` so a namespace MEMBER no longer gains transitive READ on cases they did not create. A dedicated Cypher `CaseNodeNeo4jRepository.findAccessibleByUserInNamespace` returns the MEMBER-visible case set in a single query (no N+1). Other child entity types (AgentConfig, IntegrationConfig, AiProvider, AiModel) keep the permissive MEMBER → transitive READ path required by FR21/FR27/FR32/FR35 — verified by an anti-regression test.
- **3.4 CRUD operations on cases** — fix `CaseServiceImpl.create` which silently dropped `title` and `status`; align `CaseController.toDomain` to use the `Case` data-class default `\"Case <id>\"` when the client omits a title. A new Neo4j integration test proves that soft-deleting a case leaves its `[:ADMIN]` relation intact for audit.

## Security model (unchanged)

- Super-admin bypass via `isAdmin`, handled centrally in `PermissionService.hasPermission`.
- 404 (not 403) to hide resource existence from callers with no READ.
- 403 is returned only once READ is established (explicit denial).

## Intentionally out of scope

- **PUT/PATCH semantics & merge-on-update** — the framework `SecuredEntityController.update` uses `toDomain` for both CREATE and UPDATE paths. Any field the client omits is silently reset to the DTO default (title becomes `\"Case <id>\"`, description becomes `null`, etc.). This pre-existing framework behaviour surfaces in Epic 3 only because we added content assertions on updated entities. Fixing it requires a cross-cutting decision (PUT = replace / PUT = merge + explicit clear / PATCH endpoint / JsonNullable) that should apply to every controller and will land in a dedicated story.
- **Entity metadata auto-population** — `createdBy` / `created` / `modifiedBy` / `modified` on `EntityMetadata` are not populated by this work. Covered by Epic 7 Story 7.2 (FR50/FR51).
- **Soft-delete + ID collision MERGE** — an edge case where creating a case with the id of a soft-deleted case would resurrect it. Marginal probability (UUID collision), pre-existing, to be addressed in a later cleanup story.

## Tests

- **Unit** (Kotest + MockK): new `CaseControllerSpec` covering mapping, create (MEMBER allowed, 403 on no-access, best-effort grant failure), listByParent (short-circuit and filtered paths, no-N+1 assertion), update/delete 403 for non-ADMIN (Story 3.4 AC3).
- **MVC** (`@SpringBootTest` + `@AutoConfigureMockMvc`, `test` profile): new `CaseControllerMvcIntegrationSpec` covers Bean Validation, POST 201 with title preserved, GET list (super-admin path), PUT update (success, 404, 400), DELETE 204 + invisible-in-listing + 404 on second attempt.
- **Neo4j integration** (`Neo4jTransitivePermissionsSpec`): two new tests — (1) namespace ADMIN has transitive ADMIN on cases created by other users (Story 3.2 AC1/AC2), (2) MEMBER on namespace does NOT grant READ on child Case (Story 3.3 FR15 — inverts the pre-existing test that documented the bug), plus MEMBER-with-direct-ADMIN keeps access, AC6 anti-regression on AgentConfig, and soft-delete preserves `[:ADMIN]` (Story 3.4 AC2).

## Test plan

- [ ] \`cd agentos && ./gradlew :agentos-service:test\` passes (full regression)
- [ ] Manual smoke:
  - [ ] Super-admin creates a namespace, creates a case → receives direct ADMIN; updates + deletes OK
  - [ ] Namespace ADMIN sees every case in the namespace (transitive); can update/delete any of them
  - [ ] Namespace MEMBER creates their own case → direct ADMIN; cannot see or update another member's case (404)
  - [ ] Deleting a case removes it from listings; the `[:ADMIN]` relation persists in Neo4j for audit
- [ ] Verify the PR auto-retargets to \`master\` after Epic 2 (#805) is merged (or retarget manually)

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
@@ -2,12 +2,14 @@ package io.whozoss.agentos.caseFlow
 
 import io.whozoss.agentos.entity.SecuredEntityController
 import io.whozoss.agentos.permissions.Action
+import io.whozoss.agentos.permissions.PermissionRelation
 import io.whozoss.agentos.permissions.PermissionService
 import io.whozoss.agentos.sdk.actor.Actor
 import io.whozoss.agentos.sdk.actor.ActorRole
 import io.whozoss.agentos.sdk.caseEvent.MessageContent
 import io.whozoss.agentos.sdk.entity.EntityMetadata
 import io.whozoss.agentos.user.UserService
+import jakarta.validation.Valid
 import mu.KLogging
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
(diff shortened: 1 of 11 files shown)


selimbensenouciep-prog:
Consolidated into #811 to reduce the review stack from 5 PRs to 2.

The full diff (Epic 2 → Epic 5) is now reviewable as a single PR against #803's branch. Per-epic commits are preserved — `git log` on the consolidated branch still shows the original epic boundaries for bisect / archaeology.

No code change, no force-push: this branch's commits are unchanged, only the PR pointer moved.

PR #810, opened 2026-04-25
WZ-29564 permissions v2 — Epic 4: configuration entities (namespace-bound CRUD)

selimbensenouciep-prog:
## Summary

Epic 4 of the WZ-29564 permissions-v2 initiative migrates the four configuration-entity controllers — AgentConfig, IntegrationConfig, AiProvider, AiModel — to namespace-bound transitive permissions.

Stacked on top of `permissions-v2-epic3` (#807).

### Story 4.1 — AgentConfig

- **FR17** — ADMIN of a namespace can create AgentConfig entities within that namespace
- **FR18** — ADMIN of a namespace can update AgentConfig entities within that namespace
- **FR19** — ADMIN of a namespace can delete AgentConfig entities within that namespace
- **FR21** — MEMBER of a namespace can view and use AgentConfig entities but cannot modify them
- **FR22** — User can list all AgentConfig entities they have access to in a namespace

### Story 4.2 — IntegrationConfig

- **FR23** — ADMIN of a namespace can create IntegrationConfig entities within that namespace
- **FR24** — ADMIN of a namespace can update IntegrationConfig entities within that namespace
- **FR25** — ADMIN of a namespace can delete IntegrationConfig entities within that namespace
- **FR27** — MEMBER of a namespace can view and use IntegrationConfig entities but cannot modify them

### Story 4.3 — AiProvider

- **FR28** — ADMIN of a namespace can create AiProvider entities within that namespace
- **FR29** — ADMIN of a namespace can update AiProvider entities including API keys
- **FR30** — ADMIN of a namespace can delete AiProvider entities
- **FR32** — MEMBER of a namespace can view and use AiProvider entities (relaxed: API keys are no longer masked — project decision, see story 4.3)

### Story 4.4 — AiModel

- **FR33** — ADMIN of a namespace can create AiModel entities attached to AiProviders
- **FR35** — MEMBER of a namespace can use AiModel entities without modifying them

### Dropped or replaced

- **FR20**, **FR26** — dedicated entity-level ADMINs (AgentConfig, IntegrationConfig): abandoned in favour of namespace-level ADMIN
- **FR31** — backend masking of API keys: dropped (project decision; confidentiality is handled by namespace READ access control)
- **FR34** — AiModel transitivity via AiProvider: replaced by direct transitivity via Namespace (`AiModel.namespaceId` denormalised from parent at creation time)

## Implementation notes

- All four controllers extend `SecuredEntityController` (introduced in Epic 1)
[...]

Issue #809, opened 2026-04-24
Drop user-scoped AiProvider leftover (no active consumer after Epic 4.3 migration)


selimbensenouciep-prog:
## Context

Epic 4 of the permissions v2 initiative (#WZ-29564) introduced namespace-bound security for configuration entities. Story 4.3 migrated \`AiProviderController\` to \`SecuredEntityController\`, gating CREATE/UPDATE/DELETE on namespace ADMIN.

\`AiProvider\` was originally designed to support **two scopes**:
- \`namespaceId != null\` — shared across a namespace
- \`userId != null\` — private to a single user

Story 4.3 only secures the namespace-scoped path. The user-scoped path is kept code-wise but:
- \`POST /api/ai-providers\` with \`userId\` set → 403 (no namespace-ADMIN context)
- \`GET /api/ai-providers/by-userId/{userId}\` → still reachable, but nothing currently creates new user-scoped providers

## Evidence that user-scoped is dead code

- **No UI consumer**: grep of \`libs/agentos-ui/\` and \`apps/client/\` finds zero call to \`listByUserIdAiProvider\` or equivalent
- **Auto-generated OpenAPI client only**: \`libs/agentos-api-client/.../ai-provider-controller.service.ts\` has the method, but it's never imported by a real component
- **Tests only**: \`AiProviderServiceImplSpec\` / \`AiProviderControllerSpec\` cover the code path but no integration consumer invokes it

## Proposed cleanup

Drop the user-scoped leftover entirely:

1. Remove \`userId\` field from \`AiProvider\` (SDK), \`AiProviderResource\`, \`AiProviderNode\`
2. Make \`namespaceId\` non-nullable (currently nullable because \`userId\` was the alternative)
3. Remove \`findByUserId\` from service, repo, Neo4j repo
4. Remove \`findActiveByUserId\` Cypher
5. Remove \`GET /api/ai-providers/by-userId/{userId}\` endpoint
6. Remove \`AiProviderServiceImpl\` validation that "one of namespaceId/userId must be non-null" — namespaceId becomes required
7. Clean \`AiProviderServiceImplSpec\` / \`AiProviderControllerSpec\` tests that cover user-scoped cases
8. Update frontend to regenerate the OpenAPI client

## Alternative: revive user-scoped with proper security

If a real use case emerges for user-owned providers, redesign with explicit direct-ADMIN auto-grant (same pattern as Case auto-grant in Story 3.1) rather than implicit "userId match means owner".

[...]

Issue #15, opened 2025-03-19
Add web-search capability


vincent-audibert-whoz:
Let an agent go on the web:
- search through an engine
- read web pages and get synthetic/cleaned content of the page

Options at the moment:
- OpenAI web search tool
- code tools for search through google search API or equivalent, then a parsing tool with playwright + cheerio to render js dom and clean it.

PR #811, opened 2026-04-25
WZ-29564 permissions v2

selimbensenouciep-prog:
## Summary

WZ-29564 permissions-v2 — full delivery in a single PR. Replaces the
absent permission model of `agentos-service` with a transitive Cypher
permission graph backed by Spring Security `@PreAuthorize` annotations
on every secured controller endpoint.

## Why

Before this work, `agentos-service` had no domain-level permission model:
identities were resolved (Cloudflare JWT or OS username), but every
authenticated caller could do everything. As the platform expanded to
namespaces, cases, and configuration entities, this became a blocker for
multi-tenant deployments.

We deliver the model end-to-end in one go: foundation, application
across all entity families, and the migration to declarative annotations
that make the security contract visible at the method signature.

### Permission model

- Two relations on the Neo4j graph: `[:ADMIN]` and `[:MEMBER]`
- Transitive evaluation via dedicated Cypher queries (a namespace ADMIN
  is transitively ADMIN of every entity that `[:BELONGS_TO]` that
  namespace)
- Super-admin bypass via `User.isAdmin` flag, mapped to a
  `ROLE_SUPER_ADMIN` Spring Security authority
- Fail-closed default: any error in the evaluation path yields denied

### Entity-level CRUD with permission checks

- **Namespace** — SUPER_ADMIN can create/delete; ADMIN can grant/revoke
  ADMIN+MEMBER; users see the namespaces they have access to with their
  effective role
- **Case** — namespace MEMBER can create their own cases (FR15
  OWNER_PRIVATE); namespace ADMIN sees every case (transitive); MEMBER
  without a direct grant cannot see cases owned by others
- **AgentConfig / IntegrationConfig / AiProvider / AiModel** —
  namespace-bound via the `[:BELONGS_TO]` edge wired in each
  repository's `save` path; namespace ADMIN can CRUD, MEMBER reads
- All PUT handlers preserve server-owned fields (`namespaceId`,
  `userId` on AiProvider, `aiProviderId` on AiModel,
  `externalId`+`isAdmin` on User) — mass-assignment guard against
  cross-namespace privilege escalation

### Declarative authorization at the controller layer

All 12 controllers in `agentos-service` use `@PreAuthorize` on every
`@*Mapping`. The check is at the method signature, not buried in the
body. A reviewer scanning a controller can verify the security contract
at a glance.

Three pre-existing security gaps surfaced during the migration audit
and closed:
- `CaseEventRestController` + `CaseEventSseController` had no auth at
  all — any authenticated user could read events of any case
- `PluginController.{upload, start, stop, reload, unload}` had no auth
  — a remote-code-execution path (uploading an arbitrary JAR plugin)
- `IntegrationTypeController` had no auth — fail-closed posture
  (`isAuthenticated()`); relaxable to `permitAll()` if the signup flow
  needs anonymous access

`PluginController.uploadPlugin` also rejects path-traversal attempts in
filenames (`../etc/foo.jar` → 400).

## Technical ROI

- **Defense by construction**: a missing `@PreAuthorize` is visible to
  anyone scanning `@*Mapping`. The bug class that motivated the
  declarative migration (a forgotten `update` override allowing
  cross-namespace mass-assignment, caught in mid-flight code review)
  cannot recur.
- **Transitive Cypher with one source of truth**: `PermissionService`
  centralizes the model. Future entity types only need a
  `[:BELONGS_TO]` edge to gain transitivity; no controller-side logic
  to duplicate.
- **Native `@PostFilter` for batch reads**: per-entity filtering is a
  one-line annotation.
- **Test ergonomics**: `MethodSecurityIntegrationSpec` covers the auth
  contract end-to-end once; per-controller specs focus on mapping and
  business logic.
- **`Authentication.name = User UUID`** lets SpEL self-or-admin
  expressions like
  `hasRole('SUPER_ADMIN') or #id.toString() == authentication.name`
  work without a guard bean.

## Reviewer guide

The PR is large by design (a complete permission model is hard to
deliver incrementally without intermediate states that are themselves
reviewable surfaces). To make it digest, a few orientations:

- **Read deeply**: `PermissionService` + Cypher (`Neo4jPermissionRepository`),
  `AgentOsPermissionEvaluator`, `AgentOsAuthenticationFilter`,
  `MethodSecurityConfig`. These define the model.
- **Read for the pattern, then skim**: the 12 controllers all use the
  same `@PreAuthorize` pattern. Once you understand it on one
  controller (e.g. `AgentConfigController`), the others scan in
  seconds.
- **Skip these files entirely**: `SecuredEntityController.kt` and its
  two specs are introduced and deleted within this PR — a transit
  pattern that lived across early epics and was replaced by the
  declarative migration. Reviewing them is wasted effort.

The 12-commit history is preserved for bisect and per-epic
archaeology — `git log master..HEAD` shows the original epic
boundaries.

## Test plan

- [ ] Unit specs green: 12 controller specs + 2 guard beans + 4
      security-infrastructure specs + permissions package + namespace
      permission endpoints (~250 tests passing locally)
- [ ] `MethodSecurityIntegrationSpec` exercises the full auth chain
      end-to-end: `@PreAuthorize` SpEL → `PermissionEvaluator` →
      `PermissionService`, plus 404/403 mapping and
      `hasRole('SUPER_ADMIN')` resolution
- [ ] `Neo4jTransitivePermissionsSpec` validates the Cypher transitive
      paths against a real Neo4j (testcontainers) for AgentConfig +
      Case + namespace edges
- [ ] Manual smoke: regular user cannot upload a plugin (403),
      super-admin can; cross-namespace PUT is blocked on the 5
      namespace-bound entity types; CaseEvent reads require READ on
      the parent case
- [ ] Confirm with the front-end whether `IntegrationTypeController`
      should be relaxed to `permitAll()` (signup page) or stay
      `isAuthenticated()`

## Closed sub-PRs (consolidated here)

- #803 — permissions foundation (Epic 1)
- #805 — namespace management (Epic 2)
- #807 — case lifecycle (Epic 3)
- #810 — configuration entities (Epic 4)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * Note: Neo4j does not support parameterized relationship types or label comparisons
+ * via parameters, so we use separate methods per relationship type and IN-based label checks.
+ */
+interface PermissionNodeNeo4jRepository : Neo4jRepository<UserNode, String> {
+
+    // Direct permission queries
+
+    @Query("""
+        MATCH (u:User {id: ${'$'}userId})-[r:ADMIN|MEMBER]->(e {id: ${'$'}entityId})
```
vincent-audibert-whoz, line 25: Heavy usage of same relations between different entities did seem to confuse neo4j graph queries and lead to unexpectedly high db hits => move to `[r:ADMIN_OF_{entityLabel}]` to disambiguate and reduce cardinality ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
```diff
@@ -54,5 +59,54 @@ class AgentConfigController(
...
+    ): AgentConfig =
+        existing.copy(
+            name = resource.name,
+            description = resource.description,
+            instructions = resource.instructions,
+            modelName = resource.modelName,
+        )
+
+    @GetMapping("/{id}")
+    @PreAuthorize("hasPermission(#id, 'AgentConfig', 'READ')")
```
vincent-audibert-whoz, line 85: Could it be made generic by use of the Entity class passed to the EntityController ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNodeNeo4jRepository.kt
```diff
@@ -18,4 +18,24 @@ interface AgentConfigNodeNeo4jRepository : Neo4jRepository<AgentConfigNode, Stri
...
+     * description) onto the existing Namespace node when the relationship is
+     * expressed via the @Relationship field.
+     */
+    @Query(
+        $$"""MATCH (a:AgentConfig {id: $agentConfigId})
+            MATCH (ns:Namespace {id: $namespaceId})
+            MERGE (a)-[:BELONGS_TO]->(ns)
+            """,
+    )
+    fun linkAgentConfigToNamespace(
```
vincent-audibert-whoz: Seems high candidate for generic behavior, extract and expose from the namespace package ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelGuard.kt
```diff
@@ -0,0 +1,64 @@
...
+     * - Provider user-scoped (legacy): same — returns true, body returns []
+     * - Provider namespace-scoped: requires namespace READ on the parent namespace
+     *
+     * The fall-through to true for missing/user-scoped providers is deliberate: returning false
+     * would surface 403 and disclose to a non-MEMBER caller that a namespace-scoped sibling
+     * provider exists for that id-space. Empty list is the safer default.
+     */
+    fun canListByProvider(providerId: UUID): Boolean {
+        val provider = aiProviderService.findById(providerId) ?: return true
+        val nsId = provider.namespaceId ?: return true
```
vincent-audibert-whoz: Should default to false when no `provider.namespaceId` ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelNode.kt
```diff
@@ -37,6 +43,8 @@ data class AiModelNode(
     val modified: Instant = Instant.now(),
     val modifiedBy: String? = null,
     val removed: Boolean? = null,
+    @Relationship(type = "BELONGS_TO", direction = OUTGOING)
+    val namespace: NamespaceNode? = null,
```
vincent-audibert-whoz, line 47: Reflexion: move to a domain model not referencing `namespaceId: String?` and `userId: String?` but rather a `belongsTo: { entityType: PARENT_ENTITY_TYPE_ENUM, entityId: String}` ?

What implications for the node model, could be an equivalent of `belongsTo: NamespaceNode | UserNode | null` in kotlin ?

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agentConfig/AgentConfigControllerUnitSpec.kt
vincent-audibert-whoz, line 1: Test seems simplified too much, see to re-add some cases on the AgentConfigControllerIntegrationSpec ?

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/bootstrap/BootstrapServiceImplSpec.kt
```diff
@@ -0,0 +1,91 @@
...
+        every { mockUserService.count() } returns 1L
+
+        // When - appelé plusieurs fois
+        bootstrapService.bootstrap()
+        bootstrapService.bootstrap()
+        bootstrapService.bootstrap()
+
+        // Then - seulement vérifié, pas de modifications
+        verify(exactly = 3) { mockUserService.count() }
+    }
```
vincent-audibert-whoz, line 137: Add a test to cover some existing non-root users : no super-admin added.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+    ): List<String> {
+        return try {
+            permissionRepository.listUsersWithPermission(entityType, entityId, relation)
+        } catch (e: Exception) {
+            logger.error(e) { "Failed to list users with permission on $entityType:$entityId, relation=$relation" }
+            emptyList() // Fail-closed: return empty list on error
+        }
+    }
+
+    override fun listEntitiesForUser(
```
vincent-audibert-whoz, line 150: Surprised to see it there:
- should it be fully generic
- if yes, should it be here ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+                logger.debug { "Invalid UUID format for userId: $userId" }
+                null
+            }
+            if (user?.isAdmin == true) {
+                logger.debug { "Super-admin listing all $entityType entities" }
+                // Super-admin bypasses per-entity checks in hasPermission(), so controllers
+                // (declarative @PreAuthorize / @PostFilter paths) never reach this code path
+                // for super-admins. Return empty list as a safe fallback — callers that need
+                // "all entities" should query the entity service directly.
+                return emptyList()
```
vincent-audibert-whoz: Weird ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceController.kt
```diff
@@ -53,23 +61,142 @@ class NamespaceController(
...
+        val currentUser = userService.getCurrentUser()
+
+        if (currentUser.isAdmin) {
+            return namespaceService.findAll()
+                .filter { it.metadata.removed != true }
+                .map { toListItem(it, SUPER_ADMIN) }
+        }
+
+        val userId = currentUser.id.toString()
+        val readableIds = permissionService.listEntitiesForUser(userId, ENTITY_TYPE, Action.READ)
```
vincent-audibert-whoz: move call to namespaceService

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventRestController.kt
```diff
@@ -25,22 +33,26 @@ import java.util.UUID
...
     fun getById(
         @PathVariable id: UUID,
     ): CaseEvent =
         caseEventService.findById(id)
             ?: throw ResourceNotFoundException("CaseEvent not found: $id")
 
-    /** POST /api/case-events/by-ids — get multiple events by IDs. */
+    /** POST /api/case-events/by-ids — get multiple events. Per-event filter on parent Case READ. */
     @PostMapping("/by-ids", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @PostFilter("hasPermission(filterObject.caseId, 'Case', 'READ')")
```
vincent-audibert-whoz: warning, potentially 1000s or more events in one go.

vincent-audibert-whoz: For that entity, we should either check the read on the parent Case, or go  with the `by-parentId/{caseId}`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
```diff
@@ -37,26 +58,123 @@ class CaseController(
...
-    // -------------------------------------------------------------------------
-    // Additional endpoints
-    // -------------------------------------------------------------------------
+    @GetMapping("/{id}")
+    @PreAuthorize("hasPermission(#id, 'Case', 'READ')")
+    @HideOnAccessDenied
+    override fun getById(@PathVariable id: UUID): CaseResource = super.getById(id)
+
+    @PostMapping("/by-ids", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @PostFilter("hasPermission(filterObject.id, 'Case', 'READ')")
```
vincent-audibert-whoz: Warning, can be expensive if many cases per user.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/bootstrap/BootstrapServiceImpl.kt
```diff
@@ -1,72 +1,93 @@
...
         logger.debug { "[Bootstrap] Current user count: $userCount" }
         return userCount > 0
     }
+
+    private fun createDefaultAdmin() {
+        logger.info { "[Bootstrap] No users found. Creating default admin (externalId='$defaultAdminExternalId')" }
+        userService.create(
+            User(
+                metadata = EntityMetadata(),
+                externalId = defaultAdminExternalId,
```
vincent-audibert-whoz: To ease the local usage, rather:
- have defaultAdminExternalId not defined by default in conf file
- take defaultAdminExternalId or else the local username from the os/jvm
Maybe use the SecurityService ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/Neo4jChildLinkService.kt
vincent-audibert-whoz, line 1: Great !
Is it also covered by neo4j tests in the concerned entities ?

vincent-audibert-whoz:
For the CI, the missing part is re-generating the openapi spec, then re-generating the agentos-api-client, if using Coday, Nxay knows how to do it.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
selimbensenouciep-prog: Not addressed in this PR — we refactored the application layer (typed `EntityType` enum, declarative `@PreAuthorize`) without touching the Cypher schema.

Your suggestion is valid on the graph perf side, but we're reluctant to ship it blindly: we don't have a measurement showing that the generic `:ADMIN` / `:MEMBER` edges are actually a bottleneck (no slow-query log, no abnormal hit count observed on staging).

If you already have internal numbers that motivate the observation, share them and we'll prioritize.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
selimbensenouciep-prog: Considered, but not folded into the base class. The 5 `toDomainForUpdate` implementations look similar at first glance, but the rules diverge per entity:

- `AgentConfig` / `IntegrationConfig` — preserve `namespaceId` (mass-assignment guard).
- `Case` — preserves `namespaceId` AND `status` (`status` is a lifecycle field driven by the runtime, never settable from a PUT).
- `AiProvider` — preserves `namespaceId` + `userId`, plus has a per-field fallback rule for the masked API key (incoming masked value → keep persisted secret).

A generic implementation would need either `@ServerManaged` field annotations + reflection, or an abstract `merge(resource, existing)` hook on the base. The first adds runtime metadata coupling on every Entity; the second is essentially what the per-controller private fun already is, with extra ceremony.

Net: the duplication is shallow (5 functions × ~6 lines), the rules are genuinely entity-specific, and keeping the merge logic next to the `update` endpoint helps readability when debugging a PUT. We can revisit if the pattern grows past 8-10 entities.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNodeNeo4jRepository.kt
selimbensenouciep-prog: Done — extracted as `Neo4jChildLinkService` in the `persistence` package. Single `link(childLabel, childId, parentLabel, parentId, relationship = "BELONGS_TO")` call, idempotent via `MERGE`.

Used by all 6 child-node repositories: `Neo4jAgentConfigRepository`, `Neo4jAiProviderRepository`, `Neo4jAiModelRepository`, `Neo4jIntegrationConfigRepository`, `Neo4jCaseRepository`, `Neo4jCaseEventRepository`. Removed ~6 duplicated `@Query` declarations.

Placed under `persistence/` rather than `namespace/` because the parent-child pattern isn't namespace-specific (`Case → CaseEvent` uses the same plumbing). Cf. your comment #15 on the resulting service file.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelGuard.kt
selimbensenouciep-prog: Done — both `canCreate` (line 39) and `canListByProvider` (line 58) fail-closed when `provider.namespaceId == null` (legacy user-scoped path). Same posture for missing provider and missing `aiProviderId` on create. KDoc updated to make the fail-closed cases explicit.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelNode.kt
selimbensenouciep-prog: This PR is scoped to authorization plumbing — the dual `namespaceId / userId` predates it.

Two separate questions in your suggestion:
- **Domain sealed class**: would shift the conditional from `if (nsId != null)` to `when (belongsTo)` without reducing it — net gain is marginal as long as both branches stay alive.
- **Node union**: SDN doesn't support union `@Relationship` targets natively (would require two `@Relationship` props or a custom dispatch).

Worth revisiting once #809 lands and user-scoped is removed — at that point `namespaceId` becomes non-null and the duality disappears, making the sealed class redundant rather than necessary.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agentConfig/AgentConfigControllerUnitSpec.kt
selimbensenouciep-prog: Done in commit 751faa75d.

The unit spec had been simplified from 14 to 4 tests during an earlier migration — restored back to 18, covering `toResource` / `toDomain` (incl. null optional fields), `getById` happy path + 404, `getByIds` (admin bypass + filter + empty short-circuit), `listByParent`, and the new mass-assignment guard cases.

The integration spec also gained `GET 200` and `DELETE 204` happy paths plus the `POST /by-ids` route — verbs that were only validation-tested before. Now 9 scenarios. Split principle: UnitSpec covers mock-reachable logic (mapping, delegates, mass-assignment); IntegrationSpec covers HTTP-only concerns (Bean Validation, `@PreAuthorize`, JSON serialization).

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/bootstrap/BootstrapServiceImplSpec.kt
selimbensenouciep-prog: Done in commit 653c4cccb. Added test `"is a no-op when 2 or more users exist"` (line 101-109) — mocks `userService.count() returns 2L`, verifies `create` / `update` / `findAll` are never called.

The spec now covers all four bootstrap paths: empty DB → create admin; single admin → no-op; single non-admin → legacy promote (explicit design choice for migration); 2+ users → full no-op.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
selimbensenouciep-prog: Cleaned up. The super-admin special-case at line 150 returned `emptyList()`, which contradicted `hasPermission`'s `return true` for the same user — both inconsistent and at the wrong layer.

Current state:
- `hasPermission` keeps the super-admin short-circuit → returns `true` (semantically correct: admin has all permissions).
- `listEntitiesForUser` / `filterVisibleIds` / `listUsersWithPermission` no longer have a super-admin branch. The service can't answer "all entities" without a circular dep on the entity service.
- The super-admin bypass for list-style queries is now at the boundary (`EntityController.getByIds`: `if (currentUser.isAdmin) ids.toSet() else permissionService.filterVisibleIds(...)`), where the caller has the entity service to return all ids directly.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
selimbensenouciep-prog: Same block as your comment #8 — `return emptyList()` for super-admin in `listEntitiesForUser`, contradicting `hasPermission`'s `return true`. Removed entirely. See the reply on #8 for the new layering (super-admin bypass moved to `EntityController.getByIds`).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceController.kt
selimbensenouciep-prog: Done. `permissionService.listEntitiesForUser(...)` was moved into `NamespaceService.findIdsVisibleTo(userId, action): List<UUID>` (declared in `NamespaceService.kt:30`, implemented in `NamespaceServiceImpl.kt:36`).

The controller now calls `namespaceService.findIdsVisibleTo(userId, Action.READ)` (line 167) and `findIdsVisibleTo(userId, Action.WRITE)` (line 171). The service handles the underlying permission lookup, UUID parsing, and malformed-id filtering with a WARN log — controller no longer reaches into the permission graph.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventRestController.kt
selimbensenouciep-prog: `getByIds` now caps the input at `EntityController.MAX_BATCH_SIZE = 1000` (line 82-87 → HTTP 400 above the cap). Same constant as the factored `EntityController.getByIds`.

Worth flagging that `listByCase` (`@GetMapping("/by-parentId/{caseId}")`, line 110-112) returns all events for a case without pagination — same scaling risk on the read side, not addressed here. Pagination (cursor- or offset-based) on that route is a separate task.

selimbensenouciep-prog: Went with option 1 (parent Case READ check) on all three routes:
- `getById` → `@PreAuthorize("@caseEventGuard.canRead(#id)")`. `CaseEventGuard.canRead(eventId)` resolves the event to its parent caseId, then delegates to the standard `permissionService.hasPermission(EntityType.CASE, ..., READ)`.
- `getByIds` → 2-step batch (load events → distinct caseIds → `permissionService.filterVisibleIds(EntityType.CASE, caseIds, READ)` → filter events). Single Cypher round-trip on the visibility resolution regardless of input size.
- `listByCase` → `@PreAuthorize("hasPermission(#caseId, 'Case', 'READ')")` directly on the path variable.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
selimbensenouciep-prog: Fixed — the `@PostFilter("hasPermission(filterObject.id, 'Case', 'READ')")` was the N+1 culprit (one permission check per case in the result list).

`getByIds` is now inherited from `EntityController.getByIds`, which resolves visibility in a single batch Cypher via `permissionService.filterVisibleIds(...)` — constant cost regardless of input size. Super-admin path is short-circuited at the controller boundary so it doesn't even hit the permission graph.

`listByParent` (same scaling concern on the by-parent route) also has two fast-paths: namespace-ADMIN → unfiltered single query; non-admin → dedicated `findAccessibleByUserInNamespace` Cypher applying FR15 in one round-trip.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/bootstrap/BootstrapServiceImpl.kt
selimbensenouciep-prog: Done in commit ebf952d4. Three changes:

1. `application.yml` — `admin-external-id` is commented out by default (line 71) with an explanatory note. No baseline value forces local users to override.
2. `BootstrapServiceImpl` — priority chain `configured → JVM user.name → "admin" sentinel` (lines 51-60). On `local` mode this lands on the OS username automatically.
3. Skipped `SecurityService`: bootstrap runs in an `ApplicationRunner` before any HTTP request, so there's no authenticated principal to read from. JVM `user.name` is the natural source.

Covered by 4 tests in `BootstrapServiceImplSpec` (`"falls back to JVM user.name…"`, `"explicit configured externalId wins…"`, `"falls back to 'admin' sentinel…"`, plus the empty-DB happy path).

vincent-audibert-whoz (approved):

PR #814, opened 2026-04-27
feat: publish agentos-service and plugins to GitHub Packages

mathieu-dearmey:
### Context

The `agentos/` sub-workspace contains `agentos-service`, `agentos-sdk`, and several plugins (`datetime`, `file`, `bash`, `tmux`). No CI was in place to build, test, or publish these artifacts automatically.

#### 1. Gradle version management via Nx release

Both `agentosSdk` and `agentosService` version keys in `agentos/gradle/libs.versions.toml` are now bumped automatically on every Nx release, keeping all JVM artifacts in sync with the workspace version. The `scripts/release.ts` loop makes adding future artifacts a one-liner.

#### 2. `maven-publish` for all new artifacts

Added `maven-publish` blocks to:
- `agentos-service`
- `agentos-datetime-plugin`
- `agentos-file-plugin`
- `agentos-bash-plugin`
- `agentos-tmux-plugin`

All publish to GitHub Packages under the `whoz-oss.agentos` group, mirroring the existing `agentos-sdk` setup.

#### 3. Local Nx plugin: `agentos-gradle`

A new local plugin (`tools/plugins/agentos-gradle`) uses `createNodesV2` to infer all common targets for every Gradle subproject under `agentos/*/build.gradle.kts`:

| Target | Description |
|---|---|
| `build` | `./gradlew :<name>:build`, cached, depends on `agentos-sdk:build` |
| `test` | `./gradlew :<name>:test`, cached |
| `clean` | `./gradlew :<name>:clean` |
| `nx-release-publish` | No-op echo — actual publish done by Gradle in CI |
| `publish` | `./gradlew :<name>:publish`, depends on `build` |
| `release` | Sets `manifestRootsToUpdate: []` — prevents Nx from looking for a `package.json` |

This means adding a new Gradle artifact requires zero changes to existing config files — just create the subproject with a `build.gradle.kts`, a `project.json`, and the `platform:jvm` tag.

Overrides applied where needed:
- `agentos-sdk/project.json`: removes the self-referential `agentos-sdk:build` dependency from its own `build` target
- `agentos-plugins-filesystem/project.json`: overrides `publish` with a no-op (legacy project, not published)

#### 4. `nx.json` improvements

- `release.projects` now uses `tag:platform:jvm` instead of an explicit list — new JVM projects are automatically included in the release group
- `agentos-plugins-filesystem` explicitly excluded via `!agentos-plugins-filesystem`
- Removed overly broad `"build": { "cache": false }` from `targetDefaults` — this was bleeding into JVM builds and disabling Nx caching for them. Cache settings are now handled per executor (`@nx/esbuild:esbuild`, `@nx/js:tsc`) or by the plugin (`cache: true` on JVM `build`)
- Removed `passWithNoTests` from the generic `"test"` targetDefault — it was being forwarded to Gradle as `--passWithNoTests=true`, causing build failures

#### 5. CI workflows

**`validate.yml`**: added a step to build and test all affected JVM projects (`tag:platform:jvm`) on every PR.
Uses `nx show projects --affected` to resolve affected JVM project names first, then passes them explicitly
to `nx run-many` — avoiding a flag-forwarding issue where `--projects="tag:platform:jvm"` would be passed
down to Gradle as an unknown CLI argument.

**`release.yml`**: added a single `publish-agentos-artifacts` job with a matrix strategy covering all 6 JVM
artifacts (`agentos-sdk`, `agentos-service`, `agentos-datetime-plugin`, `agentos-file-plugin`,
`agentos-bash-plugin`, `agentos-tmux-plugin`). Each matrix entry runs in parallel and publishes to GitHub
Packages via `pnpm nx publish ${{ matrix.project }}`.

#### 6. Documentation

- `agentos/AGENTOS.md`: new CI/CD section documenting the validate pipeline, release pipeline, published artifacts table, and a Gradle snippet for consuming the packages
- `tools/plugins/agentos-gradle/README.md`: documents the plugin, inference logic, override mechanism, and how to add new artifacts

### Published artifacts

All artifacts published to [GitHub Packages](https://github.com/orgs/whoz-oss/packages) under `whoz-oss.agentos`:

| Artifact | TOML version key |
|---|---|
| `agentos-sdk` | `agentosSdk` |
| `agentos-service` | `agentosService` |
| `agentos-datetime-plugin` | `agentosService` |
| `agentos-file-plugin` | `agentosService` |
| `agentos-bash-plugin` | `agentosService` |
| `agentos-tmux-plugin` | `agentosService` |


mathieu-dearmey:
## Review — feat: publish agentos-service and plugins to GitHub Packages

The PR is well-structured and the overall approach is sound. No security issues. Three things need attention before merge.

---

### `.github/workflows/release.yml` — Massive duplication, JDK 25 is EA

The five new publish jobs (`publish-agentos-service`, `publish-agentos-datetime-plugin`, `publish-agentos-file-plugin`, `publish-agentos-bash-plugin`, `publish-agentos-tmux-plugin`) are byte-for-byte identical except for the last `run:` line. This is 200+ lines of duplicated YAML that will drift the moment any step changes (e.g. JDK version, pnpm version, Node version). The existing `publish-agentos-sdk` job has exactly the same shape — five copies of it is a maintainability hazard. A single job with a matrix strategy (`matrix.project: [agentos-service, agentos-datetime-plugin, ...]`) would collapse this to one definition.

Additionally, all six publish jobs (including the pre-existing `publish-agentos-sdk`) set `java-version: '25'`. JDK 25 is an Early Access build, not a GA release; the current GA LTS is JDK 21. Using an EA JDK in a production publish pipeline is a reliability risk — EA builds can be pulled, have breaking changes, and are not supported by Temurin in the same way as GA releases. The `agentos/README.md` itself states "Java 17+". The validate job also inherits JDK 25.

---

### `tools/plugins/agentos-gradle/src/agentos-gradle.ts` — `passWithNoTests: null` is a leaky abstraction

```typescript
options: {
  command: `./gradlew :${projectName}:test`,
  cwd: 'agentos',
  passWithNoTests: null,
},
```

Setting `passWithNoTests: null` in the inferred `test` target options is the workaround documented in the PR description for preventing the flag from being forwarded to Gradle. However, the PR also removes `passWithNoTests` from the generic `"test"` `targetDefault` in `nx.json` — which means TypeScript/Jest projects now also lose `passWithNoTests: true` (previously inherited from the default). The `@nx/jest:jest` executor block still has it, so Jest projects are unaffected, but any TS project using `nx:run-commands` for tests that relied on the old generic default would silently break. The removal should be verified to have no unintended side-effects on non-JVM test targets.

---

### `nx.json` — `release.version.manifestRootsToUpdate` applies to JVM projects

```json
"version": {
  "manifestRootsToUpdate": ["apps/{projectName}"],
  ...
}
```

Now that JVM projects are included via `tag:platform:jvm`, the global `manifestRootsToUpdate: ["apps/{projectName}"]` will be evaluated for them too. The plugin overrides this per-project with `manifestRootsToUpdate: []`, which should win — but this is a fragile contract: any new JVM project that forgets to be tagged `platform:jvm` (and thus doesn't get the plugin inference) will have Nx try to find a `package.json` at `apps/<project-name>` and fail the release. Worth adding a note in the "Adding a new Gradle artifact" section of the plugin README about the `platform:jvm` tag being mandatory for this reason, not just for CI/release inclusion.

---

### Minor: `agentos-gradle` plugin registered by source path in `nx.json`

```json
{ "plugin": "./tools/plugins/agentos-gradle/src/index.ts" }
```

Registering the plugin by its TypeScript source path (rather than a compiled output or the package name) means Nx must transpile it on every cold start. This is acceptable for a local dev-only plugin, but it implies the plugin is never built/compiled as a standalone artifact. The `tsconfig.lib.json` and `package.json` suggest it could be compiled — if it grows in complexity, pointing to the compiled output would be preferable.

mathieu-dearmey:
> ## Review — feat: publish agentos-service and plugins to GitHub Packages
> The PR is well-structured and the overall approach is sound. No security issues. Three things need attention before merge.
> 
> ### `.github/workflows/release.yml` — Massive duplication, JDK 25 is EA
> The five new publish jobs (`publish-agentos-service`, `publish-agentos-datetime-plugin`, `publish-agentos-file-plugin`, `publish-agentos-bash-plugin`, `publish-agentos-tmux-plugin`) are byte-for-byte identical except for the last `run:` line. This is 200+ lines of duplicated YAML that will drift the moment any step changes (e.g. JDK version, pnpm version, Node version). The existing `publish-agentos-sdk` job has exactly the same shape — five copies of it is a maintainability hazard. A single job with a matrix strategy (`matrix.project: [agentos-service, agentos-datetime-plugin, ...]`) would collapse this to one definition.
> 
> Additionally, all six publish jobs (including the pre-existing `publish-agentos-sdk`) set `java-version: '25'`. JDK 25 is an Early Access build, not a GA release; the current GA LTS is JDK 21. Using an EA JDK in a production publish pipeline is a reliability risk — EA builds can be pulled, have breaking changes, and are not supported by Temurin in the same way as GA releases. The `agentos/README.md` itself states "Java 17+". The validate job also inherits JDK 25.
> 
> ### `tools/plugins/agentos-gradle/src/agentos-gradle.ts` — `passWithNoTests: null` is a leaky abstraction
> ```ts
> options: {
>   command: `./gradlew :${projectName}:test`,
>   cwd: 'agentos',
>   passWithNoTests: null,
> },
> ```
> 
> Setting `passWithNoTests: null` in the inferred `test` target options is the workaround documented in the PR description for preventing the flag from being forwarded to Gradle. However, the PR also removes `passWithNoTests` from the generic `"test"` `targetDefault` in `nx.json` — which means TypeScript/Jest projects now also lose `passWithNoTests: true` (previously inherited from the default). The `@nx/jest:jest` executor block still has it, so Jest projects are unaffected, but any TS project using `nx:run-commands` for tests that relied on the old generic default would silently break. The removal should be verified to have no unintended side-effects on non-JVM test targets.
> 
> ### `nx.json` — `release.version.manifestRootsToUpdate` applies to JVM projects
> ```json
> "version": {
>   "manifestRootsToUpdate": ["apps/{projectName}"],
>   ...
> }
> ```
> 
> Now that JVM projects are included via `tag:platform:jvm`, the global `manifestRootsToUpdate: ["apps/{projectName}"]` will be evaluated for them too. The plugin overrides this per-project with `manifestRootsToUpdate: []`, which should win — but this is a fragile contract: any new JVM project that forgets to be tagged `platform:jvm` (and thus doesn't get the plugin inference) will have Nx try to find a `package.json` at `apps/<project-name>` and fail the release. Worth adding a note in the "Adding a new Gradle artifact" section of the plugin README about the `platform:jvm` tag being mandatory for this reason, not just for CI/release inclusion.
> 
> ### Minor: `agentos-gradle` plugin registered by source path in `nx.json`
> ```json
> { "plugin": "./tools/plugins/agentos-gradle/src/index.ts" }
> ```
> 
> Registering the plugin by its TypeScript source path (rather than a compiled output or the package name) means Nx must transpile it on every cold start. This is acceptable for a local dev-only plugin, but it implies the plugin is never built/compiled as a standalone artifact. The `tsconfig.lib.json` and `package.json` suggest it could be compiled — if it grows in complexity, pointing to the compiled output would be preferable.

Point 1 — Duplication in release.yml

Addressed in the latest commit — collapsed the 6 publish jobs into a single publish-agentos-artifacts job with a matrix strategy. ~200 lines → ~40 lines, same parallelism.

Point 2 — JDK 25 EA

JDK 25 is intentional here. The agentos/ Gradle build targets JDK 25 via its toolchain configuration — downgrading to 21 would break the build. The CI logs confirm Java 25.0.2+10 is available on ubuntu-latest and has been running successfully. Happy to revisit if the Gradle build is ever retargeted to 21 LTS.

Point 3 — passWithNoTests removal side-effect

Audited the workspace — all TS test targets are inferred by @nx/jest/plugin and use the @nx/jest:jest executor, which has its own targetDefault block with passWithNoTests: true. There are no nx:run-commands test targets in the workspace, so no project is affected by removing passWithNoTests from the generic "test" targetDefault. The removal was necessary to prevent Gradle from receiving --passWithNoTests=true as an unknown CLI flag.

Point 4 — manifestRootsToUpdate fragility

Valid point — documented in the latest commit in tools/plugins/agentos-gradle/README.md with an explicit warning about the platform:jvm tag requirement and what breaks if it's missing. A future PR could add a validate step that cross-checks all agentos/*/build.gradle.kts projects against the tagged set to catch this at PR time.

Point 5 — .ts source path for plugin

Acknowledged. The plugin is loaded directly as TypeScript via jiti — no build step needed. If the plugin grows significantly or needs to be published, moving to a compiled artifact would be the right call at that point.

cetautomatix-coday:
## Review (from Reviewer agent)

Two issues worth addressing before merge, one potential reliability concern, and one minor flag.

~~**JDK 25 in CI** — both `validate.yml` and `release.yml` pin `java-version: '25'`. JDK 25 is not yet a GA release (current LTS is 21, current GA is 24). Using a pre-release JVM in CI is a reliability risk: builds can break on any JDK update without a corresponding change in this repo. This should be `'21'` (LTS) or at minimum `'24'`.~~
Well JDK 25 we do go with.

**`GITHUB_TOKEN` vs `RELEASE_PAT` in `publish-agentos-artifacts`** — the `publish-agentos-artifacts` matrix job uses `secrets.GITHUB_TOKEN` as `GITHUB_TOKEN` env var. The `release` job that precedes it uses `secrets.RELEASE_PAT` to push the release commit and tag. If the Gradle `maven-publish` configuration authenticates to GitHub Packages using the same `GITHUB_TOKEN` env variable, using the default ephemeral token is fine for publishing packages. However, if the Gradle publish task needs to write to a registry that requires the PAT (e.g. because the org policy restricts the default token's `packages: write` scope), this will silently fail at publish time. Verify that `secrets.GITHUB_TOKEN` with `packages: write` permission is sufficient for the GitHub Packages target, or switch to `RELEASE_PAT` for consistency.

**`passWithNoTests: null` in the Nx plugin** — in `agentos-gradle.ts`, the `test` target sets `passWithNoTests: null`. The PR description explains this is intentional to neutralize the forwarded flag, but `null` being serialized and forwarded to `nx:run-commands` is undocumented behavior. If a future Nx version changes how `null` options are handled, this silently breaks the suppression. A comment explaining the intent is present, which is good, but the fragility is worth noting.

**`--target` vs `--targets` in `validate.yml`** — the lint/test step uses `--target="lint,test"` (singular). The correct flag for running multiple targets in one `affected` call is `--targets` (plural). With the singular form, Nx may interpret `"lint,test"` as a single target name and silently skip both. Verify this actually runs both targets as intended, or split into two separate calls.

vincent-audibert-whoz (approved):
A few feedbacks from Reviewer agent ☝️ seem worthwhile ?

mathieu-dearmey:
> 
> **`GITHUB_TOKEN` vs `RELEASE_PAT` in `publish-agentos-artifacts`** — the `publish-agentos-artifacts` matrix job uses `secrets.GITHUB_TOKEN` as `GITHUB_TOKEN` env var. The `release` job that precedes it uses `secrets.RELEASE_PAT` to push the release commit and tag. If the Gradle `maven-publish` configuration authenticates to GitHub Packages using the same `GITHUB_TOKEN` env variable, using the default ephemeral token is fine for publishing packages. However, if the Gradle publish task needs to write to a registry that requires the PAT (e.g. because the org policy restricts the default token's `packages: write` scope), this will silently fail at publish time. Verify that `secrets.GITHUB_TOKEN` with `packages: write` permission is sufficient for the GitHub Packages target, or switch to `RELEASE_PAT` for consistency.
> 
> **`passWithNoTests: null` in the Nx plugin** — in `agentos-gradle.ts`, the `test` target sets `passWithNoTests: null`. The PR description explains this is intentional to neutralize the forwarded flag, but `null` being serialized and forwarded to `nx:run-commands` is undocumented behavior. If a future Nx version changes how `null` options are handled, this silently breaks the suppression. A comment explaining the intent is present, which is good, but the fragility is worth noting.
> 
> **`--target` vs `--targets` in `validate.yml`** — the lint/test step uses `--target="lint,test"` (singular). The correct flag for running multiple targets in one `affected` call is `--targets` (plural). With the singular form, Nx may interpret `"lint,test"` as a single target name and silently skip both. Verify this actually runs both targets as intended, or split into two separate calls.

GITHUB_TOKEN vs RELEASE_PAT — publish-agentos-artifacts already declares packages: write and passes the right env vars for Gradle's maven-publish. No change needed.

passWithNoTests: null — Fair point, the comment covers the intent. We'll watch it on Nx upgrades.

--target="lint,test" — Checked against Nx source: the value goes through a parseCSV coerce function that splits on commas, so "lint,test" correctly produces ["lint", "test"]. No issue here.

PR #815, opened 2026-04-28
feat: #wz-31170 add agentos-mcp-plugin with MCP stdio transport support

cetautomatix-coday:
## Summary

Adds the `agentos-mcp-plugin` module — a PF4J plugin that connects AgentOS to local MCP servers via stdio transport and exposes their tools as `StandardTool` instances in the `ToolRegistry`.

Closes WZ-31170 (partial — covers the local MCP stdio integration; remote MCP and other tool ports are tracked separately).

## What's in this PR

**New module: `agentos/agentos-mcp-plugin/`**

| File | Role |
|---|---|
| `McpPlugin.kt` | PF4J plugin lifecycle + `@Extension McpToolProvider : ToolPlugin` |
| `McpConnection.kt` | Wraps `McpSyncClient` (MCP Java SDK 0.17.0) — manages the child process lifetime |
| `McpConnectionPort.kt` | Interface over `McpConnection`, enabling unit testing without a live process |
| `McpConnectionPool.kt` | Shared pool of live connections, keyed by config hash — max 20, idle eviction every 2 min |
| `McpTool.kt` | `StandardTool<Input>` that delegates to a pooled `McpConnectionPort` |
| `McpServerConfig.kt` | Data class for server configuration (command, args, env, cwd, timeouts) |
| `McpConfigParser.kt` | Parses and validates a `JsonNode` into `McpServerConfig` |
| `McpConfigHash.kt` | SHA-256 hash of config for pool keying |

**Build changes**
- `agentos/settings.gradle.kts`: `includeBuild("agentos-mcp-plugin")` added
- `agentos/gradle/libs.versions.toml`: `mcpSdk = "0.17.0"` version + `mcp-sdk` library alias added
- `build.gradle.kts`: fat JAR bundles MCP SDK deps with `duplicatesStrategy = EXCLUDE`

**Tests** — 25 tests, all passing:
- `McpConfigParserUnitSpec` — 8 cases covering parse/validation
- `McpConfigHashUnitSpec` — 8 cases covering SHA-256 hash stability
- `McpToolUnitSpec` — 9 cases covering tool name, description, argument passing, result forwarding

## Integration type

The plugin registers `integrationType = "MCP_STDIO"`. An `IntegrationConfig` with this type and a `parameters` object like:

```json
{
  "command": "npx",
  "args": ["-y", "@modelcontextprotocol/server-github"],
  "env": { "GITHUB_PERSONAL_ACCESS_TOKEN": "..." }
}
```

will cause the plugin to spawn the server process, discover its tools, and expose them to the agent.

## Notes

- `McpConnectionPort` interface was introduced specifically to allow MockK mocking in tests — `McpConnection` has `lateinit var` fields that prevent direct mocking of the concrete class.
- The fat JAR excludes `kotlin/**` and `kotlinx/**` (already on the service classpath) and applies `duplicatesStrategy = EXCLUDE` to handle `META-INF/versions/9/module-info.class` duplicates from the MCP SDK's transitive dependencies.


cetautomatix-coday:
## Review — `deployPlugins` fix for `agentos-mcp-plugin`

The one-liner fix is correct and necessary. `agentos-mcp-plugin` is properly declared via `includeBuild` in `settings.gradle.kts` on this branch, so adding it to `pluginBuilds` in `build.gradle.kts` is the right and complete action to wire it into the `deployPlugins` task.

One structural issue is worth flagging, however: **`pluginBuilds` in `build.gradle.kts` is a manually maintained parallel list that must be kept in sync with the `includeBuild(...)` declarations in `settings.gradle.kts`**. There is no build-time enforcement of this invariant — a plugin can be registered as an included build (so Gradle knows about it) but silently omitted from `pluginBuilds` (so `deployPlugins` never builds or copies it), which is exactly the bug this PR fixes. This happened once; it will happen again the next time a plugin is added.

The structural improvement would be to derive `pluginBuilds` from `gradle.includedBuilds` at configuration time, filtering out non-plugin modules (`agentos-sdk`, `agentos-service`) by name prefix or an explicit exclude list. That would make `deployPlugins` self-maintaining and eliminate the category of omission entirely. This is not a blocker for this PR, but worth tracking as a follow-up.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHash.kt
```diff
@@ -0,0 +1,40 @@
...
+    val sb = StringBuilder()
+    when (transport) {
+        McpTransport.STDIO -> {
+            sb.append("stdio")
+            sb.append('|').append(command)
+            args.forEach { sb.append('|').append(it) }
+            env.entries.sortedBy { it.key }.forEach { sb.append('|').append(it.key).append('=').append(it.value) }
+            cwd?.let { sb.append('|').append(it) }
+        }
+        McpTransport.HTTP -> {
```
vincent-couturier-whoz, line 41: question : so the plugin also managed remote MCP ?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHash.kt
```diff
@@ -0,0 +1,40 @@
...
+            sb.append("http")
+            sb.append('|').append(url)
+            authToken?.let { sb.append('|').append(it) }
+        }
+    }
+    sb.append('|').append(timeoutSeconds)
+    sb.append('|').append(toolCallTimeoutSeconds)
+    val digest = MessageDigest.getInstance("SHA-256").digest(sb.toString().toByteArray())
+    return digest.joinToString("") { "%02x".format(it) }
```
vincent-couturier-whoz, line 50: suggestion: have constant instead of magic value

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
```diff
@@ -0,0 +1,120 @@
...
+ * **HTTP validation rules:**
+ * - `url` must be non-blank.
+ * - `authToken`, when present, must be non-blank.
+ *
+ * **Shared validation rules:**
+ * - Timeout values, when provided, must be positive.
+ *
+ * Throws [IllegalArgumentException] with a descriptive message on any violation.
+ */
+object McpConfigParser {
```
vincent-couturier-whoz, line 36: suggestion: perhaps sealed classes could have clarified the code and we could/should have StdIoMcpServerConfig and RemoteMcpServerConfig with only relevant information and not an all in one McpServerConfig with a "pick what you want" strategy

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    private fun parseEnv(config: JsonNode): Map<String, String> {
+        val envNode = config.get("env") ?: return emptyMap()
+        require(envNode.isObject) { "MCP integration config: 'env' must be an object" }
+        return envNode.properties().asSequence().associate { (k, v) -> k to v.asText() }
+    }
+
+    // ----- HTTP -----
+
+    private fun parseHttp(config: JsonNode, url: String): McpServerConfig {
```
vincent-couturier-whoz: suggestion: parseHttpConfig

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnection.kt
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * Wraps [McpSyncClient] from the MCP Java SDK 0.17.x. The underlying child process is
+ * started when [connect] is called and remains alive until [close] is called.
+ *
+ * [lastUsed] is updated on every [callTool] invocation. The [McpConnectionPool]
+ * uses it to evict idle connections.
+ *
+ * Instances are created and owned exclusively by [McpConnectionPool].
+ */
+class McpConnection(
```
vincent-couturier-whoz: suggestion: StdioMcpConnection

agentos/agentos-mcp-plugin/project.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "agentos-mcp-plugin",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "projectType": "library",
+  "sourceRoot": "agentos/agentos-mcp-plugin/src",
+  "targets": {
```
mathieu-dearmey: les targets build, clean, publish et test sont inférrés par le plugin local (récent) agentos-gradle; ça nous évite tout le boilerplate dans ce ficher.
Tu peux regarder le project json de agentos-tmux-plugin par exemple

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnectionPool.kt
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Returns a healthy [McpConnection] for [config], creating one if needed.
+     *
+     * If an existing pooled connection for the same config hash is found but its
+     * process has died, it is replaced with a fresh connection transparently.
+     *
+     * @throws McpConnectionException if a new connection cannot be established.
+     * @throws IllegalStateException if the pool is at capacity.
+     */
+    fun acquire(config: McpServerConfig): McpConnection {
```
vincent-couturier-whoz: issue : I think in case of multiple access to acquire at the same time we will strange behavior because of the registration of the connexion without using proper concurrent access method on the pool. Here we check the existence and then we create and store but even if the map is a ConcurrentHashMap it will behave as classical map if we split get and put. putIfAbsent should be used to really take advantage of thread safety of ConcurrentHashMap.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpPlugin.kt
```diff
@@ -0,0 +1,152 @@
...
+
+        val serverConfig = try {
+            McpConfigParser.parse(config)
+        } catch (e: IllegalArgumentException) {
+            logger.error { "MCP_STDIO integration '$configName': invalid config — ${e.message}" }
+            return emptyList()
+        }
+
+        val connection = try {
+            McpConnectionPoolHolder.pool.acquire(serverConfig)
```
vincent-couturier-whoz: question : How we ensure the McpToolProvider is executed after the load of the pool. It's strange because we have a singleton but a "dynamic" execution of the start/shutdown which won't necessarily be executed at the correct moment.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHash.kt
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Two configs with identical behaviour produce the same hash and will share
+ * a pooled [McpConnection].
+ */
+fun McpServerConfig.configHash(): String {
+    val sb = StringBuilder()
+    when (transport) {
+        McpTransport.STDIO -> {
+            sb.append("stdio")
+            sb.append('|').append(command)
```
thibault-drillien-whoz: Suggestion: mettre les chaines dans des constantes qui portent dans leur nom leur but. Ex HASH_SEPARATOR pour le "|"

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * **Shared validation rules:**
+ * - Timeout values, when provided, must be positive.
+ *
+ * Throws [IllegalArgumentException] with a descriptive message on any violation.
+ */
+object McpConfigParser {
+
+    fun parse(config: JsonNode): McpServerConfig {
+        val command = config.get("command")?.takeIf { !it.isNull }?.asText()?.trim()?.takeIf { it.isNotBlank() }
```
thibault-drillien-whoz: Plutôt que de parser le fichier à la "main" on pourrait définir l'objet attendu (McpServerConfig) et passer par la désérialisation JSon ?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpServerConfig.kt
```diff
@@ -0,0 +1,63 @@
...
+    val transport: McpTransport
+        get() = when {
+            command != null -> McpTransport.STDIO
+            url != null -> McpTransport.HTTP
+            else -> throw IllegalStateException("McpServerConfig has neither 'command' nor 'url' set")
+        }
+
+    /** Human-readable identifier for logging — command for stdio, url for HTTP. */
+    val label: String
+        get() = command ?: url ?: "(unconfigured)"
```
vincent-couturier-whoz, line 56: question : why the integration "name" is not used for logs ?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
```diff
@@ -0,0 +1,136 @@
...
+ */
+class McpConnection(
+    private val config: McpServerConfig,
+    val configHash: String,
+) : McpConnectionPort {
+    private lateinit var client: McpSyncClient
+    private lateinit var transport: StdioClientTransport
+
+    /** Tools discovered at connect time. Immutable after [connect]. */
+    override var tools: List<Tool> = emptyList()
```
thibault-drillien-whoz, line 35: Pourquoi ne pas rester en `val` ?

vincent-audibert-whoz:
Retours adressés dans le commit `f17b61ad` :

**`project.json` (Mathieu)** — Tu avais raison, j'ai manqué l'introduction du plugin Nx `agentos-gradle` sur master qui infère `build`, `test`, `clean`, `publish`. Le `project.json` est réduit aux seules métadonnées (`name`, `$schema`, `projectType`, `sourceRoot`, `tags`), conformément aux autres plugins.

**`McpConfigHash.kt` (Thibault)** — Les séparateurs `"|"` et `"="` sont extraits en constantes `HASH_SEPARATOR` et `HASH_KEY_VALUE_SEPARATOR`.

**`McpConfigParser.kt` (Thibault)** — Adopté. Jackson désérialise directement le `JsonNode` vers `McpServerConfig` via `treeToValue`, les validations métier (exclusivité command/url, non-blank, timeouts positifs) sont concentrées dans un `validate()` appelé en post-désérialisation. Le parser passe de 120 lignes à ~60, sans perte de couverture (les 8 tests existants passent tous, les messages d'erreur sont identiques).

**`McpConnection.tools` (Thibault)** — La propriété reste `override var tools` avec `private set`. Elle ne peut pas être `val` parce qu'elle est peuplée dans `connect()` après construction, et Kotlin n'a pas de `lateinit val`. La rendre `private var` et l'exposer via un getter `val` séparé serait équivalent mais plus verbeux. `private set` garantit qu'elle n'est modifiable que depuis `McpConnection` elle-même, ce qui est la contrainte réelle ici.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
vincent-audibert-whoz: Peuplé après.

mathieu-dearmey (dismissed):

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpTool.kt
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * The MCP server owns the schema and expects the raw argument map. Rather than
+     * deserializing into a typed class (which would require a class per tool), we
+     * wrap the raw JSON string in [Input] and re-serialize it inside [execute].
+     * This keeps deserialization inside the plugin classloader as required.
+     */
+    override fun executeWithJson(json: String?): String = execute(Input(args = json))
+
+    companion object : KLogging() {
+        private val objectMapper = jacksonObjectMapper()
```
vincent-couturier-whoz, line 95: thought : We should find a way to use injection instead of specific singleton for each plugin component.

agentos/agentos-mcp-plugin/src/test/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHashUnitSpec.kt
```diff
@@ -0,0 +1,94 @@
...
+        val a = McpServerConfig(url = "https://mcp.example.com/sse")
+        val b = McpServerConfig(url = "https://mcp.example.com/sse", authToken = "token")
+        a.configHash() shouldNotBe b.configHash()
+    }
+
+    // ----- transport isolation -----
+
+    "stdio and HTTP configs with same text do not collide" {
+        // Pathological case: command value equals url value
+        val stdio = McpServerConfig(command = "https://mcp.example.com/sse")
```
vincent-couturier-whoz, line 90: question : why this test ? it seems a little weird to have a url in a command

agentos/agentos-mcp-plugin/src/test/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHashUnitSpec.kt
```diff
@@ -0,0 +1,94 @@
...
+import io.kotest.matchers.shouldNotBe
+
+class McpConfigHashUnitSpec : StringSpec({
+
+    // ----- stdio -----
+
+    "identical stdio configs produce the same hash" {
+        val a = McpServerConfig(command = "docker", args = listOf("run", "--rm", "img"))
+        val b = McpServerConfig(command = "docker", args = listOf("run", "--rm", "img"))
+        a.configHash() shouldBe b.configHash()
```
vincent-couturier-whoz, line 14: issue: it's strange because if the integration name or namespaceId is different we should have a specific hash in order to avoid communication between both integration even if they have the same definition. For example if I have a configuration of an integration for a namespace I don't wan't to reuse for the connection of another integration for another namespace. But perhaps I'm wrong 🤔 .

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolPlugin.kt
```diff
@@ -44,8 +46,7 @@ import org.pf4j.ExtensionPoint
  *
  *     override fun provideTools(config: JsonNode?, configName: String?): List<StandardTool<*>> {
  *         val apiKey = config?.get("apiKey")?.asText() ?: ""
- *         val prefix = configName?.let { "${it}__" } ?: ""
- *         return listOf(MyCustomTool(name = "${prefix}MyTool", apiKey = apiKey))
+ *         return listOf(MyCustomTool(apiKey = apiKey))
```
vincent-couturier-whoz, line 49: question: why this change to remove the name of the tool ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/PluginConfiguration.kt
```diff
@@ -37,21 +44,61 @@ class PluginConfiguration {
...
+ * Only classes absent from the service classpath are loaded from the plugin JAR itself.
+ * This is the standard behaviour of managed runtimes (OSGi, JEE) and eliminates the
+ * duplicate-class problem without requiring plugins to carefully exclude every shared lib.
  */
 private class NullSafeSpringPluginManager(pluginsRoot: Path) : SpringPluginManager(pluginsRoot) {
     override fun createExtensionFactory(): ExtensionFactory =
         NullSafeSpringExtensionFactory(this)
+
+    override fun createPluginLoader(): PluginLoader =
+        CompoundPluginLoader()
```
vincent-couturier-whoz, line 71: question : Not sure of the impact of this change. It make the instance of classes available for the processing in the plugin but it will potentially create failure for the plugin with embedded jar which is in conflict with core agentos services. Perhaps we should ask coday to check the impact and ensure that this change won't introduce a regression on another plugin

agentos/build.gradle.kts
```diff
@@ -25,13 +25,12 @@ tasks.named("build") {
...
-    "agentos-plugins-filesystem",
-    "agentos-datetime-plugin",
-    "agentos-tmux-plugin",
-    "agentos-bash-plugin",
-    "agentos-file-plugin"
-)
+// Non-plugin included builds -- excluded from deployPlugins.
+// All other included builds are assumed to be plugins.
+val nonPluginBuilds = setOf("agentos-sdk", "agentos-service")
+val pluginBuilds: List<String> = gradle.includedBuilds
```
vincent-couturier-whoz: suggestion: it could be dangerous. We could just check the builds containing "plugin"

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHash.kt
vincent-audibert-whoz: Could be soon because the underlying sdk covers it too. However the credential/auth topic is not dry at all on AgentOS, so this integration plugin is aiming only at local MCP.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
vincent-audibert-whoz: This would mean different schemas for configuration and go with different plugins for remote and local, even if a good chunk of the code is the same... Will need to be discussed.

agentos/agentos-mcp-plugin/src/test/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHashUnitSpec.kt
vincent-audibert-whoz: This is debatable as same configs should lead exactly to the same tool and the namespace has no influence in that. For example the github local MCP has the same configuration for use on Coday repo or on another repo my user has access to => makes no sense then to spawn another docker image with exactly the same parameters. 

If 2 integrations from different users and namespaces are resolved as the same IntegrationConfig (in term of hash), both use the same tool.

There was in the design a `no-share` attribute to prevent sharing an instance if the MCP is stateful, will have to look at it again.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolPlugin.kt
vincent-audibert-whoz: Does not seem normal, abusive merge ?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHash.kt
vincent-audibert-whoz: constant extracted

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpPlugin.kt
vincent-audibert-whoz: PF4J guarantees that `Plugin.start()` is called before any `@Extension` on that plugin is resolved by the plugin manager. The sequence is: load JAR → instantiate `Plugin` subclass → call `start()` → resolve extensions. So `McpConnectionPoolHolder.start()` (and therefore the eviction executor) is always running before any `provideTools()` call. No ordering issue here.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpServerConfig.kt
vincent-audibert-whoz, line 56: The `label` property is intentionally based on `command`/`url` rather than the integration `configName`. The pool is shared across all namespaces and configs — its log entries describe the *server process* being managed, not the AgentOS integration that requested it. Using `configName` here would be misleading because the same pooled connection can be reused by multiple integrations with different names. The `configName` is already present in all `McpToolProvider.provideTools` log lines, which is where the integration-level identity is relevant.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
```diff
@@ -0,0 +1,98 @@
+package io.whozoss.agentos.plugins.mcp
```
vincent-audibert-whoz, line 1: Done — `McpConfigParser` now delegates structural mapping to Jackson (`mapper.treeToValue`) and only contains business-rule validation (`validate`, `validateStdio`, `validateHttp`, `validateTimeouts`). The mapper is configured with `KotlinModule` (NullToEmptyCollection + NullToEmptyMap) and string-to-null coercion for Maps/Collections so that lenient frontend inputs degrade gracefully to defaults.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpTool.kt
vincent-audibert-whoz, line 95: PF4J extensions are instantiated by reflection via `ExtensionFactory` — there is no IoC container inside a plugin JAR. Injecting the pool into `McpToolProvider` via a constructor would require either a custom `ExtensionFactory` on the service side (complex, couples the service to the plugin's internals) or passing it through the `ToolPlugin` interface (which would change the SDK contract for all plugins). The object singleton scoped to the plugin classloader is the standard PF4J pattern for shared plugin-internal state, and it is safe here because the classloader lifetime matches the plugin lifecycle. Worth tracking as a future architectural improvement if PF4J is ever replaced with a DI-aware loading mechanism.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolPlugin.kt
vincent-audibert-whoz: only docs.

agentos/agentos-mcp-plugin/src/test/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigHashUnitSpec.kt
vincent-audibert-whoz: Creative robustness test.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
```diff
@@ -0,0 +1,98 @@
...
+
+        if (hasCommand) validateStdio(config) else validateHttp(config)
+        validateTimeouts(config)
+
+        return config
+    }
+
+    private fun validateStdio(config: McpServerConfig) {
+        config.args.forEachIndexed { i, arg ->
+            require(arg.isNotBlank()) { "MCP integration config: args[$i] must not be blank" }
```
selimbensenouciep-prog, line 86: **Worth settling before merge.** `validateStdio` only checks that `args` are non-blank, but `command`, `args` and `env` come straight from the `IntegrationConfig` and are used to spawn an arbitrary child process in `StdioMcpConnection.connect()`. Configuring an `MCP_STDIO` integration is therefore effectively remote code execution on the server. This is inherent to MCP stdio, so the real question is a trust-boundary decision: who is allowed to create an `MCP_STDIO` integration? If that is not restricted to platform admins, any config author gets remote command execution. Could we add an allow-list of permitted executables (and/or a sandbox), and document the trust boundary explicitly?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnectionPool.kt
```diff
@@ -0,0 +1,142 @@
...
+                        .between(pooled.connection.lastUsed, now)
+                        .toSeconds()
+                val threshold = pooled.config.idleTimeoutMinutes * 60
+                logger.trace { "[McpPool] Connection ${hash.take(8)}: idle ${idleSeconds}s / threshold ${threshold}s" }
+                idleSeconds > threshold
+            }
+        toEvict.forEach { (hash, pooled) ->
+            logger.info { "[McpPool] Evicting idle connection ${hash.take(8)} for '${pooled.config.command}'" }
+            pool.remove(hash)
+            pooled.connection.close()
```
selimbensenouciep-prog: **Minor / latent robustness.** Idle eviction closes connections based only on `StdioMcpConnection.lastUsed`, with no coordination with in-use connections: `lastUsed` is set only at the start of `callTool()` (never refreshed at the end), and `acquire()` does not bump it when reusing a connection, so a connection can in principle be closed while a caller still holds it. In practice this rarely bites with the defaults (requestTimeout 60s is well below the 5 min idle threshold, and a fresh call refreshes `lastUsed` before the next eviction tick), and it is recoverable (surfaces as `McpConnectionException`, the next `provideTools` reconnects). It only becomes a real problem if `toolCallTimeoutSeconds` is raised above `idleTimeoutMinutes * 60`, where a long call can be evicted mid-flight. Cheap hardening if you want it: refresh `lastUsed` in a `finally` at call end, bump it on reuse in `acquire()`, and skip eviction while a call is in progress (a simple active-call flag or refcount).

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
```diff
@@ -0,0 +1,185 @@
...
+            // McpSyncClient.close() calls transport.close() which is non-blocking.
+            logger.warn { "[MCP] Graceful close failed, forcing shutdown (hash ${configHash.take(8)})" }
+            runCatching { client.close() }
+        }
+        val closeMs = System.currentTimeMillis() - closeStart
+        logger.debug { "[MCP] Close completed in ${closeMs}ms (hash ${configHash.take(8)})" }
+    }
+
+    private fun formatResult(result: McpSchema.CallToolResult): String {
+        val content = result.content() ?: return "(no output)"
```
selimbensenouciep-prog, line 185: In MCP, a tool reports an application-level failure via the `isError` flag on `CallToolResult`, not by throwing. `formatResult` only reads `result.content()` and ignores `isError`, so a tool that returns an error result (isError = true, error text in `content`) is surfaced by `McpTool.execute` as `ToolExecutionResult.success(...)`. The agent then cannot distinguish a genuine success from a tool-level error and may proceed as if the call succeeded. Suggestion: when the result is flagged as an error, return `ToolExecutionResult.error(...)` with the content text; otherwise `success`. (Worth double-checking the exact accessor name in MCP SDK 2.0, but `isError` is part of the `CallToolResult` schema.)

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnectionPool.kt
```diff
@@ -0,0 +1,142 @@
...
+                        createConnection(config, hash)
+                    }
+                }
+            PooledConnection(connection = result!!, config = config)
+        }
+
+        return result!!
+    }
+
+    private fun createConnection(
```
selimbensenouciep-prog, line 114: `McpConnectionPool` has no tests, and the design makes them hard to add: `createConnection()` constructs `StdioMcpConnection` directly, so the pool cannot be exercised without spawning real child processes. Yet this is where the risk-bearing logic lives (reuse, dead-connection reconnect, idle eviction, capacity cap). Suggestion: inject a connection factory `(McpServerConfig, String) -> McpConnectionPort` (defaulting to `::StdioMcpConnection`) so the pool's acquire / reuse / eviction / cap behaviour can be unit-tested against a fake `McpConnectionPort`, in line with the team's push for tests that exercise the actual risky paths (#1070).

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @throws McpConnectionException if the process fails to start or the MCP handshake times out.
+     */
+    fun connect() {
+        require(config.transport == McpTransport.STDIO) {
+            "McpConnection only supports stdio transport; HTTP transport is not yet implemented"
+        }
+        val command = config.command!!
+        val connectStart = System.currentTimeMillis()
+        logger.info { "[MCP] Connecting to server: $command ${config.args.joinToString(" ")}" }
```
selimbensenouciep-prog: `connect()` logs the full command and args at INFO: `logger.info { "[MCP] Connecting to server: $command ${config.args.joinToString(" ")}" }`. `env` is deliberately not logged (good), but args are logged verbatim. MCP servers are often configured with credentials passed as CLI flags (e.g. `--dsn postgres://user:secret@host/db`, `--api-key`, `--token`), which then land in INFO logs on every (re)connect. This is somewhat incremental (args are also visible in the host process table), but logs have wider audience and retention. Suggestion: log the command and arg count at INFO, and gate the full (or redacted) arg list behind debug/trace, mirroring the exclusion already applied to `env`.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
```diff
@@ -0,0 +1,185 @@
...
+
+        val paramsBuilder =
+            ServerParameters
+                .builder(command)
+                .args(config.args)
+        if (config.env.isNotEmpty()) {
+            paramsBuilder.env(config.env)
+        }
+        val params = paramsBuilder.build()
```
selimbensenouciep-prog, line 63: `connect()` builds `ServerParameters` from `command`, `args` and `env`, but never applies `config.cwd`, so the child process runs in the service JVM's working directory. Yet `cwd` is parsed, validated, hashed, and documented (schema + KDoc: "Optional working directory for the child process"), so the field currently does nothing. Suggestion: wire `cwd` into the process launch if the SDK's `ServerParameters` supports a working directory; otherwise drop the field, its schema entry, and its KDoc so it does not advertise behaviour that is not there.

selimbensenouciep-prog (dismissed):

agentos/build.gradle.kts
vincent-audibert-whoz: moved to an explicit list for now.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConfigParser.kt
vincent-audibert-whoz: MCP configuration ends up being RCE anyway, this is accepted as managed in practice by platform/namespace administrators in the deployed instance, and super-admin do whatever they want locally.
There is a gap however for simple members of a namespace that can add their overrides => #1104

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnectionPool.kt
vincent-audibert-whoz: lastUsed also updated on tool response.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpConnectionPool.kt
vincent-audibert-whoz: Tests were in a commit not pushed, pushed.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
vincent-audibert-whoz: tool error handled through its own exception.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
vincent-audibert-whoz: info cleaned

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/StdioMcpConnection.kt
vincent-audibert-whoz: `cwd` wired through `CwdAwareStdioClientTransport`

selimbensenouciep-prog (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/PluginConfiguration.kt
vincent-audibert-whoz: Coday (Archay) answer:

The concern is valid but has been addressed in the current state of the branch.

The problem APD solves: with the default PDA strategy, the MCP plugin bundles the MCP SDK which pulls Jackson transitively. PDA loads the plugin's copy first → two distinct Jackson class instances coexist → LinkageError whenever a Jackson type crosses the plugin/service boundary via SDK interfaces.

How APD fixes it: the plugin classloader delegates to the service first. Shared libs (Jackson) are resolved from the service classpath; plugin-specific libs (MCP SDK, Reactor) are loaded from the plugin JAR. This is the standard approach in managed runtimes (OSGi, JEE).

Build-level protection: the MCP plugin declares Jackson as compileOnly in its build.gradle.kts, so it's explicitly excluded from the fat JAR. Only MCP SDK and its transitive deps are bundled. This ensures the plugin never ships a conflicting copy of shared libraries.

Residual risk: APD applies globally to all plugins. A future plugin needing a different version of a service-provided lib would be forced to use the service's version. This is a known trade-off, mitigated by the fact that plugins are developed in-house and depend on the SDK via compileOnly, which naturally aligns versions.

Issue #817, opened 2026-04-30
Return on global mission control


benjamin-valdes-whoz:
-bugfix navbar responsive
-filter by favoris
-keep star display when it's a favorite thread
-add settings on home
-add mark as down button on thread
-add possibility to navigate to project since the modal

PR #818, opened 2026-04-30
fix: #817 return on global mission control

benjamin-valdes-whoz:


mathieu-dearmey:
## Code Review

No security issues. The PR correctly addresses all items from issue #817. Three issues to fix before merge.

---

### 1. `GlobalTaskControlComponent` injects `ConfigApiService` directly — violates service layer architecture

**File**: `apps/client/src/app/components/global-task-control/global-task-control.component.ts`

```typescript
private readonly configApi = inject(ConfigApiService)
```

The `openUserConfig()` method performs the full HTTP fetch + dialog orchestration + save directly in the component by calling `ConfigApiService`. Per the mandatory architecture, components must never call API services directly — all business logic and API coordination belongs in a state service. The same pattern is used in `GlobalMissionControlComponent` (not in this PR's diff but visibly the same issue here).

This violates the API service → state service → component layering rule. The `openUserConfig` flow (fetch config, open dialog, save on close) is business logic that belongs in a dedicated state service or at minimum in an existing config state service.

---

### 2. `ThreadSelectorComponent` injects `ProjectApiService` directly — same architectural violation

**File**: `apps/client/src/app/components/thread-selector/thread-selector.component.ts`

```typescript
private readonly projectApi = inject(ProjectApiService)
```

`markDone` and `markActive` call `projectApi.markThreadDone` / `projectApi.markThreadActive` directly from the component. These operations are also available via `TaskControlComponent` which correctly delegates to `projectApi` — but even there the pattern is borderline since `TaskControlComponent` has no state service intermediary for these new mark-done/active operations. In `ThreadSelectorComponent` the violation is clear: an API service is injected into a component.

---

### 3. Inconsistent starring logic between `GlobalTaskControlComponent` and `TaskControlComponent`

**File**: `apps/client/src/app/components/global-task-control/global-task-control.component.ts`, line ~110

```typescript
if (this.starredOnly()) {
  tasks = tasks.filter((t) => t.starring.length > 0)
}
```

**File**: `apps/client/src/app/components/task-control/task-control.component.ts`, line ~68

```typescript
if (this.starredOnly() && currentUsername) {
  tasks = tasks.filter((t) => t.starring.includes(currentUsername))
}
```

`TaskControlComponent` filters by `starring.includes(currentUsername)` — i.e. "starred by me". `GlobalTaskControlComponent` filters by `starring.length > 0` — i.e. "starred by anyone". These semantics are different. The global view should be consistent with the per-project view and filter by the current user's stars, not any star. The same inconsistency applies in `filterCounts`.

---

### Minor — PR title capitalisation

The PR title `Bugfix/benjamin/817 return on global mission control` does not follow the conventional commit format required by DEV_WORKFLOW.md. It should be `fix: #817 return on global mission control`.

mathieu-dearmey (approved):

PR #821, opened 2026-05-05
feat: #31644 Implement get and create user group

AlanMenantWhoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceServiceImpl.kt
```diff
@@ -19,16 +21,26 @@ class NamespaceServiceImpl(
     private val namespaceRepository: NamespaceRepository,
     private val permissionService: PermissionService,
 ) : NamespaceService {
-    override fun create(entity: Namespace): Namespace = namespaceRepository.save(entity)
-
-    override fun update(entity: Namespace): Namespace = namespaceRepository.save(entity)
+    override fun create(entity: Namespace): Namespace = try {
+        namespaceRepository.save(entity)
+    } catch (e: DataIntegrityViolationException) {
+        throw ConflictException("A namespace with externalId '${entity.externalId}' already exists", e)
```
vincent-audibert-whoz, line 27: Useless re-throw, exception should bubble up to spring controller for automatic status re-use ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/api/user-groups", produces = [MediaType.APPLICATION_JSON_VALUE])
+class UserGroupController(
```
vincent-audibert-whoz, line 16: Will be to rework to match the security pattern we are starting to adopt as in `NamespaceController.kt`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
```diff
@@ -0,0 +1,16 @@
+package io.whozoss.agentos.userGroup
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Size
+
+@Schema(name = "UserGroupCreateRequest")
+data class UserGroupCreateRequest(
```
vincent-audibert-whoz, line 9: Seems bundling too much, rather split:
- user group creation
- users addition to the group
- agents addition to the group

dev_tools/api/http-client.env.json
vincent-audibert-whoz, line 1: Why this one ?

vincent-audibert-whoz (dismissed):
Can move on, nice on externalId on namespace, some API design issues to adjust in next PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupServiceImpl.kt
```diff
@@ -0,0 +1,55 @@
...
+        val namespace =
+            namespaceService.findByExternalId(request.namespaceExternalId)
+                ?: throw UnprocessableEntityException("Namespace not found for externalId: ${request.namespaceExternalId}")
+
+        create(
+            UserGroup(
+                namespaceId = namespace.id,
+                name = request.name,
+            ),
+        )
```
vincent-audibert-whoz, line 48: Race condition: after `create()` succeeds, the returned `UserGroup` already contains the persisted entity with its ID. Re-querying by `namespaceExternalId + name` to find it back is unnecessary and introduces a TOCTOU window — a concurrent insert of a same-named group could match the filter. Use the return value of `create()` directly and convert it to `UserGroupSearchResult` without the extra query.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun initSchema() {
+        driver.session().use { session ->
+            session.run(
+                "CREATE CONSTRAINT user_group_name_namespace_unique IF NOT EXISTS " +
+                    "FOR (g:UserGroup) REQUIRE (g.name, g.namespaceId) IS UNIQUE",
+            )
+            logger.info { "[Neo4jSchemaInitializer] Constraint user_group_name_namespace_unique ensured" }
+
+            session.run(
+                "CREATE CONSTRAINT namespace_external_id_unique IF NOT EXISTS " +
```
vincent-audibert-whoz, line 27: `REQUIRE n.externalId IS UNIQUE` without a `WHERE n.externalId IS NOT NULL` guard may reject multiple `Namespace` nodes with `externalId = null` depending on the Neo4j version (4.x treats `null` as a distinct value for uniqueness; 5.x behaviour changed). The test "multiple namespaces with null externalId are allowed" will silently pass in embedded mode if the constraint is not actually enforced there. Verify the constraint syntax against the target Neo4j version, or use a node-key constraint scoped to non-null values.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
```diff
@@ -0,0 +1,16 @@
...
+    @field:NotBlank
+    val namespaceExternalId: String,
+    @field:NotBlank
+    @field:Size(max = 250)
+    val name: String,
```
vincent-audibert-whoz, line 14: `userIds` and `agentIds` are declared in the request body but are never read in `createFromRequest`. This is a misleading API contract: callers sending these fields will receive a 201 with no indication their data was ignored. Remove the fields until the feature is implemented, or explicitly reject non-empty values.

vincent-audibert-whoz (commented):
This PR lays a solid foundation for user groups. No security issues or structural blockers. Four concerns to address — two in this PR, two worth tracking for follow-up.

**Must fix now:**

1. **Race condition in `createFromRequest`** — the method calls `create()` then immediately re-queries by `namespaceExternalId + name` to retrieve the created entity. Between the insert and the re-query, another concurrent write could match the filter first, or the filter could theoretically match a different group with the same name. The created entity's ID is already known after `create()` returns; the round-trip query is unnecessary and fragile. Use the returned `UserGroup` directly.

2. **`namespace_external_id_unique` constraint is too strict** — the constraint `REQUIRE n.externalId IS UNIQUE` applies to all `Namespace` nodes including those with `externalId = null`. In Neo4j, a uniqueness constraint on a nullable property will reject a second `null` value (depending on the Neo4j version), which breaks the explicit test case "multiple namespaces with null externalId are allowed". The constraint should be `REQUIRE n.externalId IS UNIQUE` scoped only to non-null values, or use `IS NODE KEY` / a partial constraint. Verify the embedded-Neo4j test actually enforces this constraint before relying on it.

**Track for next PR (API design, as noted by Vincent):**

3. **`UserGroupController` does not extend `EntityController`** — other resources use the abstract `EntityController` base for consistent CRUD. `UserGroupController` is a standalone `@RestController` with only GET (search by `namespaceExternalId`) and POST. The GET endpoint returns a flat search result rather than a namespaced resource hierarchy (`/api/namespaces/:id/user-groups`). This is acceptable as a first step but needs alignment with the REST resource hierarchy in the next PR.

4. **`userIds` and `agentIds` in `UserGroupCreateRequest` are accepted but silently ignored** — `UserGroupCreateRequest` declares `userIds: List<String>` and `agentIds: List<String>` with defaults, but `createFromRequest` only uses `name` and `namespaceExternalId`. The ignored fields are confusing and misleading to API consumers. Either remove them from the request until implemented, or throw `NotImplementedException` when they are non-empty.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceServiceImpl.kt
mickael-barroux-whoz: Question is, do we always want to throw a 409 Conflict for any `DataIntegrityViolationException` ?

It's for our dedicated use-case only IMO

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
mickael-barroux-whoz: We decided to have a endpoint to be able to create the user group and attach the users and agents in a a single shot

dev_tools/api/http-client.env.json
mickael-barroux-whoz: To be able to select the env in intellij with HTTP requests

Capture d’écran 2026-05-05 à 14 08 19

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupServiceImpl.kt
mickael-barroux-whoz: There is a TODO but we will need to re-query the created group using it's ID in order to retrieve the user counter + list of agentIds (`UserGroupSearchResultResource`)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
mickael-barroux-whoz: Will be implemented later, let as is

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
mickael-barroux-whoz: We will add the security layer with `@PreAuthorize` later if it's effectively the concern of this remark ;)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun findByParent(parentId: UUID): List<UserGroup> =
+        neo4jRepository
+            .findActiveByNamespaceId(parentId.toString())
+            .map { it.toDomain() }
+
+    override fun findByNamespaceExternalId(externalId: String): List<UserGroupSearchResult> =
+        neo4jClient
+            .query(
+                "MATCH (g:UserGroup)-[:BELONGS_TO]->(ns:Namespace) " +
```
romain-petit-whoz: Maybe use triple " and $$ here

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroup.kt
```diff
@@ -0,0 +1,13 @@
+package io.whozoss.agentos.userGroup
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import io.whozoss.agentos.sdk.entity.Entity
+import io.whozoss.agentos.sdk.entity.EntityMetadata
+import java.util.UUID
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
romain-petit-whoz: Is it an input data class ? if it's not related to input or output it should not have a json annotation

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val userGroupService: UserGroupService,
+) {
+    @GetMapping
+    fun searchByNamespaceExternalId(
+        @RequestParam namespaceExternalId: String,
```
romain-petit-whoz, line 21: No validation on this field ? like a max size for example ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
```diff
@@ -0,0 +1,16 @@
+package io.whozoss.agentos.userGroup
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Size
+
+@Schema(name = "UserGroupCreateRequest")
```
romain-petit-whoz, line 8: why a @Schema annotation here ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
```diff
@@ -0,0 +1,16 @@
...
+data class UserGroupCreateRequest(
+    @field:NotBlank
+    val namespaceExternalId: String,
+    @field:NotBlank
+    @field:Size(max = 250)
+    val name: String,
+    val userIds: List<String> = emptyList(),
```
romain-petit-whoz, line 15: Validation ? not null ? valid uuid ? max size ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupNode.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.data.neo4j.core.schema.Relationship.Direction.OUTGOING
+import java.time.Instant
+import java.time.Instant.now
+import java.util.UUID
+
+@Node("UserGroup")
+data class UserGroupNode(
+    @Id
+    val id: String,
+    val namespaceId: String,
```
romain-petit-whoz, line 17: We put both namespaceId et namespace in the repo ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.data.neo4j.repository.Neo4jRepository
+import org.springframework.data.neo4j.repository.query.Query
+
+interface UserGroupNodeNeo4jRepository : Neo4jRepository<UserGroupNode, String> {
+
+    @Query(
+        $$"""
+            MATCH (g:UserGroup)
+            WHERE g.namespaceId = $namespaceId AND (g.removed IS NULL OR g.removed = false)
+            RETURN g ORDER BY g.name ASC
```
romain-petit-whoz, line 12: why an order by here ?

romain-petit-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
AlanMenantWhoz: Le problème est clair. La clause WHERE dans un CREATE CONSTRAINT n'est pas supportée par Neo4j Community Edition — c'est une fonctionnalité Enterprise Edition uniquement (les "property existence constraints" conditionnelles).

Dans Community Edition, une contrainte IS UNIQUE s'applique à tous les nœuds du label, y compris ceux où la propriété est absente. Mais l'absence de propriété (null) n'est pas indexée — Neo4j ne considère pas deux nœuds sans la propriété comme en conflit d'unicité. Autrement dit, la contrainte IS UNIQUE sans WHERE fait déjà ce que tu veux : elle garantit l'unicité des valeurs présentes tout en autorisant plusieurs nœuds sans externalId.

La solution est simplement de supprimer le WHERE n.externalId IS NOT NULL :

CREATE CONSTRAINT namespace_external_id_unique IF NOT EXISTS 
FOR (n:Namespace) REQUIRE n.externalId IS UNIQUE
Ce comportement est standard en Neo4j : les nœuds sans la propriété indexée sont transparents pour la contrainte d'unicité.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
AlanMenantWhoz: On a fait comme User pour l'instant. On verra après.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupCreateRequest.kt
AlanMenantWhoz: ça parait nécessaire pour générer la doc API

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupNodeNeo4jRepository.kt
AlanMenantWhoz: On s'est posé la question, mais sinon ça sera par ordre "naturel", je ne sais pas si c'est mieux. Pas de spec à ce sujet.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupNode.kt
AlanMenantWhoz: Il faut le laisser, notamment pour la contrainte d'unicité sur userGroup.name et namespaceId. Vois ça comme une clé étrangère.

vincent-audibert-whoz (approved):

PR #829, opened 2026-05-05
feat: tag-based project selection for release pipeline

mathieu-dearmey:
Replaces all hardcoded project lists in `nx.json` and `release.yml` with tag-driven selection.

### Changes
- Add `platform:npm` tag to the 5 npm-published projects
- Remove `platform:jvm` from `agentos-plugins-filesystem` (legacy, not published)
- `nx.json` release projects: `["tag:platform:npm", "tag:platform:jvm"]`
- `release.yml`: new `compute-jvm-projects` job resolves the JVM matrix dynamically via `nx show projects --json --quiet`
- Added `|| echo '[]'` fallback on the JVM resolver to avoid opaque failures on nx errors

Adding a new publishable project now only requires the right `platform:*` tag — no CI changes needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -270,7 +270,7 @@ jobs:
         id: resolve
         # --json --quiet mirrors validate.yml to avoid output pollution from the nx daemon/cache
         run: |
-          PROJECTS=$(npx nx show projects --projects="tag:platform:jvm" --json --quiet | jq -c '.')
+          PROJECTS=$(npx nx show projects --projects="tag:platform:jvm" --json --quiet | jq -c '.' || echo '[]')
           echo "projects=$PROJECTS" >> $GITHUB_OUTPUT
 
   publish-agentos-artifacts:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -242,10 +242,41 @@ jobs:
             echo "Warning: Coday Twin PKG not found"
           fi
 
-  publish-agentos-artifacts:
-    name: Publish ${{ matrix.project }} to GitHub Packages
+  # Job 4: Resolve the list of JVM projects to publish from the platform:jvm tag
+  compute-jvm-projects:
     runs-on: ubuntu-latest
     needs: release
+    if: needs.release.outputs.new_version != ''
+    outputs:
+      projects: ${{ steps.resolve.outputs.projects }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.release.outputs.release_sha }}
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 10
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Resolve JVM projects from tag
+        id: resolve
+        # --json --quiet mirrors validate.yml to avoid output pollution from the nx daemon/cache
+        run: |
+          PROJECTS=$(npx nx show projects --projects="tag:platform:jvm" --json --quiet | jq -c '.')
+          echo "projects=$PROJECTS" >> $GITHUB_OUTPUT
+
+  publish-agentos-artifacts:
+    name: publish ${{ matrix.project }} to github packages
+    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


apps/client/project.json
```diff
@@ -4,7 +4,7 @@
   "projectType": "application",
   "prefix": "app",
   "sourceRoot": "apps/client/src",
-  "tags": [],
+  "tags": ["platform:node"],
```
vincent-palita-whoz: client doesn't really run on "node". It uses npm packages though

vincent-audibert-whoz: `apps/client` is a browser Angular bundle, not a Node.js runtime. Tagging it `platform:node` includes it in the `nx release` project scope alongside server/desktop npm packages. If `client` has no `nx-release-publish` target that makes sense (or if it does publish to npm unintentionally), this could silently publish the wrong artifact. The tag semantics need to be clarified — either `platform:node` means "published via npm" (in which case `client` should be excluded or use a different tag), or `client` genuinely needs to be versioned and published.

.github/workflows/release.yml
```diff
@@ -242,25 +242,50 @@ jobs:
...
+        # --json --quiet mirrors validate.yml to avoid output pollution from the nx daemon/cache
+        run: |
+          PROJECTS=$(npx nx show projects --projects="tag:platform:jvm" --json --quiet | jq -c '.')
+          echo "projects=$PROJECTS" >> $GITHUB_OUTPUT
+
+  publish-agentos-artifacts:
+    name: publish ${{ matrix.project }} to github packages
+    runs-on: ubuntu-latest
+    needs: [release, compute-jvm-projects]
+    if: needs.release.outputs.new_version != ''
```
vincent-audibert-whoz, line 281: If `nx show projects` exits non-zero (daemon crash, workspace parse error), the `projects` output is never written to `$GITHUB_OUTPUT`, making the downstream `fromJson(needs.compute-jvm-projects.outputs.projects)` receive an empty string and causing an opaque workflow failure. A safe fallback: `npx nx show projects --projects="tag:platform:jvm" --json --quiet | jq -c '.' || echo '[]'`. This ensures the matrix job is simply skipped rather than failing in a hard-to-diagnose way.

docs/to-rework/RELEASE_PIPELINE.md
```diff
@@ -0,0 +1,17 @@
+# Release Pipeline
```
vincent-audibert-whoz, line 1: This document doesn't mention the `compute-jvm-projects` job at all, describing the pipeline as if it goes directly from `release` to `publish-agentos-artifacts`. The actual workflow now has 5 jobs: `check-release` → `release` → `desktop-release` + `compute-jvm-projects` → `publish-agentos-artifacts`. The omission creates an inconsistency with `docs/nxpert/01-coday-nx-knowledge.md` which correctly lists all 5 jobs, and will mislead anyone (human or agent) using this doc as the authoritative reference.

vincent-audibert-whoz (commented):
The approach is clean and the intent — making the release pipeline self-maintaining via tags — is solid. Two issues need attention before merge.

**`apps/client` tag accuracy**: `platform:node` implies a Node.js runtime artifact (npm-published server/desktop), but `client` is a browser bundle. It gets pulled into the `nx release` project scope via this tag, which may produce an unwanted npm publish step for what is essentially a build artifact, not a standalone package. The existing comment from @vincent-palita-whoz on the diff flags the same concern. Either a more precise tag (`platform:npm`?) or explicit exclusion is needed.

**Empty JVM project list edge case**: If no project carries `platform:jvm` at release time, `nx show projects --json --quiet` returns `[]`. `fromJson([])` produces an empty matrix, and GitHub Actions silently skips the `publish-agentos-artifacts` job without error — which is fine. However if the command itself fails (e.g. nx daemon error), the step exits non-zero and the output variable is never set, causing `fromJson` to receive an empty string and the whole workflow to fail in a confusing way. Consider adding `|| echo '[]'` as a fallback so failures in the resolver are surfaced cleanly rather than propagated opaquely.

**Documentation count discrepancy**: `docs/to-rework/RELEASE_PIPELINE.md` describes the release pipeline as having 4 jobs, but the updated `docs/nxpert/01-coday-nx-knowledge.md` correctly lists 5. The new doc is wrong on first read and will mislead agents using it as a reference.

vincent-audibert-whoz (approved):

PR #832, opened 2026-05-05
fix: relax super-admin creation/update via API with self-rule

selimbensenouciep-prog:
## Summary

- **Backend** — `POST /api/users` honors body `isAdmin` (gated `SUPER_ADMIN`) ; `PUT /api/users/{id}` honors body `isAdmin` only when caller ≠ target.
- **Self-rule** — a user can NEVER change their own `isAdmin` via the API. Combined with `@PreAuthorize`, this guarantees the DB cannot reach zero super-admins via API calls (proof in code comments).
- **Front** — admin user-form now includes an `isAdmin` checkbox, disabled in self-edit mode. Resolves the TS compile error from #811.

## Why

1. **Compile fix** — Since PR #811 (`WZ-29564 permissions v2`), the generated TS client requires `User.isAdmin` (boolean), breaking `user-form.component.ts:95` with `TS2345 — Property 'isAdmin' is missing`. The form was never updated.
2. **Operational** — bootstrap was the only path to create a super-admin. If the bootstrap admin leaves, no one can promote a replacement without direct DB manipulation.
3. **Contract honesty** — the OpenAPI schema declared `isAdmin: required` but the backend force-set `false`. Lying contract.

## Key design decisions

- **PUT replace-semantic** — `isAdmin: Boolean = false` (non-null, with default) is consistent with the rest of the API (e.g. `AiModelResource.priority: Int = 0`). Clients on PUT are expected to send the full state; an omitted field deserialises to the Kotlin default. A pinned regression test documents this behavior.
- **Self-rule subsumes "last super-admin" guard** — explicit decision documented in spec. Proof: only another super-admin can demote (non-admins are blocked by `@PreAuthorize`), so if there is ONE super-admin, no one can demote them. No counter needed.
- **No dedicated audit logger** — relies on the existing `userService.create/.update` info logs. Can be revisited in a follow-up if compliance requires.
- **`override` + `SecurityContextHolder`** — kept the parent's `update` signature to avoid Spring registering two competing `@PutMapping` handlers. `Authentication` is read from `SecurityContextHolder` inside the method.

## Test plan

Backend unit (`UserControllerSpec`) :
- [x] POST `isAdmin: true` (caller = super-admin) → persists `true`
- [x] POST `isAdmin: false` → persists `false`
- [x] PUT super-admin on OTHER user `isAdmin: true` → promote
- [x] PUT super-admin on OTHER user `isAdmin: false` → demote
- [x] PUT self with `isAdmin: false` (existing `true`) → **preserved** (self-rule)
- [x] PUT self with `isAdmin: true` (existing `false`) → **preserved** (self-rule)
- [x] PUT delegates and preserves isAdmin via self-rule (regression on existing test)

Backend MVC integration (`UserControllerSelfRuleMvcSpec` — new) :
- [x] All of the above through full Spring MVC stack
- [x] PUT partial body on OTHER super-admin (no `isAdmin` field) → **resets to false** (replace-semantic, pinning test)
- [x] PUT on OTHER user when caller is non-admin → 403

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
@@ -75,7 +75,7 @@ class UserController(
             firstname = resource.firstname,
             lastname = resource.lastname,
             bio = resource.bio,
-            isAdmin = resource.isAdmin ?: false,  // null in request body → default non-admin
+            isAdmin = resource.isAdmin,
         )
 
     @GetMapping("/{id}")
@@ -127,15 +127,12 @@ class UserController(
         // needs ANOTHER super-admin to be demoted). auth.name is the User's UUID
         // as a String (cf. AgentOsAuthentication.getName()).
         //
-        // Body-omitted isAdmin (resource.isAdmin == null) preserves existing.isAdmin too:
-        // protects against silent demotes when a client PUTs a partial body without isAdmin.
+        // PUT is replace-semantic: clients are expected to send the full state.
+        // A partial body that omits isAdmin will reset it to `false` (Kotlin default
+        // on the DTO) — same convention as other resource DTOs in this API.
         val callerName = SecurityContextHolder.getContext().authentication?.name
         val isSelfEdit = callerName != null && id.toString() == callerName
-        val newIsAdmin = when {
-            isSelfEdit -> existing.isAdmin
-            resource.isAdmin != null -> resource.isAdmin
-            else -> existing.isAdmin
-        }
+        val newIsAdmin = if (isSelfEdit) existing.isAdmin else resource.isAdmin
         val updated = toDomain(resource).copy(
             metadata = existing.metadata,
             externalId = existing.externalId,        // server-managed (IdP key)
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
@@ -11,6 +11,7 @@ import mu.KLogging
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
@@ -33,8 +34,10 @@ import java.util.UUID
  * - Super-admin only (list, create, delete, batch fetch): `hasRole('SUPER_ADMIN')`
  * - `/me`: any authenticated user — resolves their own record
  *
- * The `update` override preserves [User.externalId] and [User.isAdmin] from the persisted
- * entity (mass-assignment guard — these are server-managed fields).
+ * The `update` method preserves [User.externalId] always (server-managed IdP key).
+ * [User.isAdmin] is preserved only in self-edit mode (caller == target). When a
+ * super-admin updates another user, [User.isAdmin] from the request body is honored.
+ * Self-rule guarantees no API call can ever leave the DB with zero super-admins.
  */
 @RestController
 @RequestMapping(
@@ -72,7 +75,7 @@ class UserController(
             firstname = resource.firstname,
             lastname = resource.lastname,
             bio = resource.bio,
-            isAdmin = resource.isAdmin,
+            isAdmin = resource.isAdmin ?: false,  // null in request body → default non-admin
         )
 
     @GetMapping("/{id}")
@@ -106,10 +109,8 @@ class UserController(
     @ResponseStatus(HttpStatus.CREATED)
     @PreAuthorize("hasRole('SUPER_ADMIN')")
     override fun create(@Valid @RequestBody resource: UserResource): UserResource {
-        // Never allow isAdmin promotion from request body — only system/bootstrap can grant super-admin.
-        val newUser = toDomain(resource).copy(isAdmin = false)
-        val created = service.create(newUser)
-        logger.info { "Super-admin created user ${created.id}" }
+        val created = service.create(toDomain(resource))
+        logger.info { "Super-admin created user ${created.id} (isAdmin=${created.isAdmin})" }
         return toResource(created)
     }
 
@@ -121,10 +122,24 @@ class UserController(
     ): UserResource {
         val existing = userService.findById(id)
             ?: throw ResourceNotFoundException("Entity not found: $id")
+        // Self-rule: a user can never change their own isAdmin via the API.
+        // Guarantees the DB never reaches 0 super-admins via API (a super-admin
+        // needs ANOTHER super-admin to be demoted). auth.name is the User's UUID
+        // as a String (cf. AgentOsAuthentication.getName()).
+        //
+        // Body-omitted isAdmin (resource.isAdmin == null) preserves existing.isAdmin too:
+        // protects against silent demotes when a client PUTs a partial body without isAdmin.
+        val callerName = SecurityContextHolder.getContext().authentication?.name
+        val isSelfEdit = callerName != null && id.toString() == callerName
+        val newIsAdmin = when {
+            isSelfEdit -> existing.isAdmin
+            resource.isAdmin != null -> resource.isAdmin
+            else -> existing.isAdmin
+        }
         val updated = toDomain(resource).copy(
             metadata = existing.metadata,
             externalId = existing.externalId,        // server-managed (IdP key)
-            isAdmin = existing.isAdmin,              // server-managed (only system/bootstrap can set)
+            isAdmin = newIsAdmin,                    // self-rule applied above
         )
         logger.info { "User profile $id updated" }
         return toResource(userService.update(updated))
--- a/agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
+++ b/agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
@@ -194,30 +194,33 @@ class UserControllerSelfRuleMvcSpec : StringSpec() {
         }
 
         // -----------------------------------------------------------------
-        // PUT — partial body without isAdmin field preserves existing
-        // (regression cover for adversarial finding F12)
+        // PUT — replace semantic : a body omitting isAdmin resets it to false.
+        // Documents the API contract and aligns with the rest of the API
+        // (e.g. AiModelResource.priority defaulting to 0). Clients on PUT must
+        // send the full state.
         // -----------------------------------------------------------------
 
-        "PUT partial body (no isAdmin field) on OTHER super-admin preserves existing.isAdmin=true" {
+        "PUT partial body (no isAdmin field) on OTHER super-admin resets isAdmin to false (replace-semantic)" {
             val target = User(
                 metadata = EntityMetadata(id = targetUserId),
                 externalId = "other-admin@example.com",
                 email = "other-admin@example.com",
-                isAdmin = true,    // existing target IS a super-admin
+                isAdmin = true,
             )
             val captured = slot<User>()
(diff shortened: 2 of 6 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -121,10 +122,21 @@ class UserController(
...
             ?: throw ResourceNotFoundException("Entity not found: $id")
+        // Self-rule: a user can never change their own isAdmin via the API.
+        // Guarantees the DB never reaches 0 super-admins via API (a super-admin
+        // needs ANOTHER super-admin to be demoted). auth.name is the User's UUID
+        // as a String (cf. AgentOsAuthentication.getName()).
+        //
+        // PUT is replace-semantic: clients are expected to send the full state.
+        // A partial body that omits isAdmin will reset it to `false` (Kotlin default
+        // on the DTO) — same convention as other resource DTOs in this API.
+        val callerName = SecurityContextHolder.getContext().authentication?.name
```
vincent-audibert-whoz, line 133:
```kotlin
val callerName = SecurityContextHolder.getContext().authentication?.name
```

The `authentication` object is already bound by Spring Security before this method is entered (the `@PreAuthorize` on this method proves it). Reading it again from `SecurityContextHolder` is redundant coupling to the static thread-local and introduces a silent failure mode: if `authentication` is `null` (unauthenticated path, internal call, or test that forgets to seed the context), `isSelfEdit` evaluates to `false`, and a super-admin editing themselves would have their `isAdmin` flag overwritten by the request body without any error or log.

Prefer accepting `Authentication` as a method parameter (Spring MVC injects it automatically when declared) or extracting it from the already-available `@PreAuthorize` context. This makes the dependency explicit and eliminates the null-silent-failure path. Architectural violation: business logic that depends on security context should receive it as an explicit dependency, not pull it from a global static holder.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -106,10 +109,8 @@ class UserController(
...
     @PreAuthorize("hasRole('SUPER_ADMIN')")
     override fun create(@Valid @RequestBody resource: UserResource): UserResource {
-        // Never allow isAdmin promotion from request body — only system/bootstrap can grant super-admin.
-        val newUser = toDomain(resource).copy(isAdmin = false)
-        val created = service.create(newUser)
-        logger.info { "Super-admin created user ${created.id}" }
+        val created = service.create(toDomain(resource))
+        logger.info { "Super-admin created user ${created.id} (isAdmin=${created.isAdmin})" }
         return toResource(created)
     }
```
vincent-audibert-whoz, line 116:
```kotlin
val created = service.create(toDomain(resource))
logger.info { "Super-admin created user ${created.id} (isAdmin=${created.isAdmin})" }
```

The log now records `isAdmin` at creation, which is good. But there is no equivalent change to the `update` log on line 143 (`"User profile $id updated"`). A privilege change (promote/demote) on another user is a security-sensitive operation; the log should record both the caller identity and the new `isAdmin` value so the event is auditable. Inconsistency: `create` logs `isAdmin`, `update` does not.

libs/agentos-ui/src/lib/components/user-form/user-form.component.ts
```diff
@@ -50,6 +52,26 @@ export class UserFormComponent implements OnInit {
...
 
+    // Self-rule: the backend ignores isAdmin changes from the caller on their own
+    // record. Disable the checkbox in self-edit mode so the UX matches the contract.
+    // Fallback: if currentUser hasn't been loaded yet (e.g. direct landing), fetch it
+    // before deciding — without this, a self-edit could leave the checkbox enabled.
+    const applySelfEditDisable = () => {
+      if (this.userState.currentUser()?.id === this.userId) {
+        this.form.controls.isAdmin.disable()
+      }
+    }
```
vincent-audibert-whoz, line 63: Optional:

```typescript
this.userState
  .loadMe()
  .pipe(takeUntilDestroyed(this.destroyRef))
  .subscribe({
    next: () => applySelfEditDisable(),
  })
```

Missing `error` handler. If `loadMe()` fails (network error, 401 after session expiry, etc.), the `isAdmin` checkbox remains enabled in self-edit mode. The backend enforces the rule regardless, but the UI misleads the user into thinking they can change their own admin status. Add an `error` callback that disables the checkbox unconditionally (conservative fallback), or at minimum logs the failure.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
```diff
@@ -0,0 +1,241 @@
...
+ * Bean Validation and the basic happy-path stay covered by [UserControllerMvcIntegrationSpec].
+ */
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.MOCK)
+@AutoConfigureMockMvc
+@ActiveProfiles("test")
+class UserControllerSelfRuleMvcSpec : StringSpec() {
+    override fun extensions() = listOf(SpringExtension)
+
+    @Autowired lateinit var mockMvc: MockMvc
+
```
vincent-audibert-whoz, line 46:
```kotlin
@MockkBean(relaxed = true) lateinit var userService: UserService
```

With `relaxed = true`, any un-stubbed call on `userService` returns a default (null/false/empty). The test for `PUT on OTHER user when caller is non-admin returns 403` only stubs `getCurrentUser()` and relies on `@PreAuthorize` to block the call before `findById` is reached. This is correct, but if the `@PreAuthorize` annotation were accidentally removed, the relaxed mock would return `null` from `findById`, causing a 404 rather than a 403, and the test would fail for the right reason. This is acceptable, but worth noting that the 403 coverage depends entirely on the annotation being present — there is no explicit assertion that the security layer (not just the controller logic) is what blocks the call. Consider using `verify { userService.findById(any()) wasNot Called }` to assert that the controller body was never entered, making the test's intent explicit.

vincent-audibert-whoz (changes requested):
The PR correctly fixes the TS compile error and makes the `isAdmin` contract honest. The self-rule logic is sound, well-documented, and backed by thorough tests. However, there are two concerns that need to be addressed before merge: one security-relevant fragility in the backend and one functional bug in the frontend.

**Backend — `SecurityContextHolder` coupling in the controller.** The self-rule check reads the caller identity from `SecurityContextHolder` directly inside the controller method, while the `@PreAuthorize` SpEL expression on the same method already has `authentication` bound as a parameter. This duplication is fragile: if the method is ever called outside the filter chain (e.g. internal service-to-service call, or a future refactor that injects `Authentication` as a method parameter), `SecurityContextHolder.getContext().authentication` may be `null`, and `isSelfEdit` silently falls to `false`, meaning a super-admin editing themselves would lose their own admin flag without error. The `authentication` object is already available to `@PreAuthorize`; it should be passed into the method or resolved via an injected parameter rather than pulled from the static holder.

**Frontend — disabled form control value is lost on PUT.** `save()` uses `this.form.getRawValue()` which correctly includes disabled controls, so the `isAdmin` value sent to the backend is always the pre-loaded value from `syncForm`. This is fine. The subtle bug is in `syncForm` ordering: in the `else` branch (state not yet loaded), `applySelfEditDisable()` runs asynchronously *after* `loadAll()` resolves, but `syncForm` is also called inside that same `next` callback — meaning the disable fires *after* the form is populated, which is correct. However in the fast-path (state already loaded), `applySelfEditDisable()` is called *before* `syncForm` is ever called (which happens synchronously right after). This means `disable()` is called while the field still holds the default `false`, and then `syncForm` calls `setValue` which re-enables the control on Angular reactive forms (setValue on a FormGroup does not respect disabled state — it resets the control value but does not re-enable it, so this is actually safe). Worth verifying this explicitly with a test or a comment, because the ordering is non-obvious and a future reader could break it. More critically: there is no `error` handler on the `loadMe()` subscription — if the request fails, the checkbox stays enabled in self-edit mode, which is a UX inconsistency (the backend would still enforce the rule, but the UI would mislead the user).

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
vincent-audibert-whoz, line 1: convention: rename to `*IntegrationSpec.kt`.

vincent-audibert-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
selimbensenouciep-prog: Fair point on the architectural purity, but injecting `Authentication` as a method parameter is not implementable here without a parent refactor. `UserController.update` overrides `EntityController.update`, whose signature is `update(id, resource)` and which already carries `@PutMapping("/{id}")`. Two paths:

- **Keep `override`** → Kotlin forbids extending the signature, so I can't add `auth: Authentication`.
- **Drop `override`** → two `@PutMapping("/{id}")` handlers (parent + child) coexist on the same bean; Spring picks one non-deterministically, and in practice it routes to the parent and bypasses the self-rule. That's exactly the regression we hit during the first attempt (documented in the tech-spec frontmatter under `implementation_deviation`).

Refactoring `EntityController` to expose `update` without `@PutMapping` (or to take `Authentication?` everywhere) would touch every entity controller and is out of scope for this hotfix.

On the silent-failure concern itself: the `@PreAuthorize` SpEL on this method evaluates `authentication.name` — if `authentication` is `null`, that expression throws NPE before the body runs, so we cannot enter `update` with a null auth via the MVC stack. The only way to reach the `?.name` path with `null` is a direct in-process call bypassing Spring Security entirely, which isn't a realistic risk for a `@RestController` method.

Happy to open a follow-up to refactor `EntityController` if you want the cleaner shape across the codebase, but I'd rather keep this PR minimal.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
selimbensenouciep-prog: Good catch — fixed. `update` now logs caller identity, both isAdmin values (before/after), and the `selfEdit` flag:

```kotlin
logger.info {
    "User profile $id updated by caller=$callerName " +
        "(isAdmin: ${existing.isAdmin} -> ${persisted.isAdmin}, selfEdit=$isSelfEdit)"
}
```

Privilege changes are now traceable from the standard log stream without requiring a dedicated audit logger.

libs/agentos-ui/src/lib/components/user-form/user-form.component.ts
selimbensenouciep-prog: Fixed. Added the missing `error` callback as a conservative fallback — disables the `isAdmin` checkbox unconditionally if `loadMe()` fails, so the UI never advertises a permission the backend will reject:

```typescript
this.userState
  .loadMe()
  .pipe(takeUntilDestroyed(this.destroyRef))
  .subscribe({
    next: () => applySelfEditDisable(),
    error: () => this.form.controls.isAdmin.disable(),
  })
```

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
selimbensenouciep-prog: Fixed. Added an explicit `verify(exactly = 0) { userService.findById(targetUserId) }` to the 403 test, with a comment pinning the intent: the 403 must come from `@PreAuthorize`, not from a relaxed-mock `findById` returning `null` (which would yield 404). If the annotation were ever removed, the assertion would fail.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/user/UserControllerSelfRuleMvcSpec.kt
selimbensenouciep-prog: Done — `git mv` to `UserControllerSelfRuleIntegrationSpec.kt`, class renamed accordingly.

PR #837, opened 2026-05-05
feat: user-level overlays for Tools + LLM (#wz-31189)

selimbensenouciep-prog:
## What this PR does

Lets users personalize their Tools (IntegrationConfig) and LLM providers (AiProvider)
configurations without admin intervention, in 2 modes:

- **User × namespace** — override only on a specific namespace
- **User-global** — override on every namespace the user has access to

Each resource page (`/integrations`, `/ai-providers`) is now a
3-section view:

1. Configurations du namespace (NS-shared, admin-only writes)
2. Mes overrides sur ce namespace
3. Mes overrides globaux

Every card exposes a "Duplicate" action that opens a pre-filled form letting the
user retarget the destination scope via a radio. The user profile (`/me`) gains
a collapsable recap of all user-global overrides across the 2 resources, kept in
sync with mutations from any namespace page.

## Runtime behavior

When an agent runs, the backend resolves Tools and LLM provider configs through a 3-tier
reconciliation that merges, in precedence order: `user × namespace` >
`user-global` > `namespace shared`. Resolution is applied per-name, per-resource,
and uses a per-run cache to avoid duplicate lookups.

## Permissions

- Anyone can CRUD their own overrides (user-scope).
- Edit/delete on the namespace section is gated by either super-admin
  (`User.isAdmin = true`) or namespace ADMIN by relation (`[:ADMIN]` to the
  namespace). The Create button defaults to `userOnNs` for non-admins so they
  don't hit a silent 403 on submit.
- Ownership-based access: users can always read/write/delete entities they own,
  enforced via `OwnershipAware` / `OwnershipResolver` (OCP-compliant, auto-collected
  by Spring).

## Security

- `apiKey` is never displayed on a card, never carried into a Duplicate, and
  never re-sent on update if the user didn't touch the field. An empty value
  on submit means "explicit clear" and is sent as JSON null. Pattern aligned
  with PR #811 (super-admin self-rule).
- User-to-user isolation enforced backend-side (story-level integration tests
  cover the 8 cases of the cross-user matrix per controller).

## Backend changes

- `IntegrationConfig` aligned on the same triple-mode pattern as `AiProvider`:
  `namespaceId: UUID?` and `userId: UUID?` both nullable, with the
  invariant `(namespaceId != null) OR (userId != null)`.
- `ConfigMergeService<T>` introduced as a generic 3-tier merge primitive
  (used by Tools and AiProvider; ready for other entities in a follow-up).
- `ToolResolverService` extracted from `ToolRegistryService` for tool resolution.
- `OverlayKeyEncoding` encodes the `(namespaceId, userId, name)` triple into a
  single discriminator backing Neo4j UNIQUE CONSTRAINTs.
- `OwnershipAware` interface + `OwnershipResolver`: OCP-compliant ownership
  resolution for the permission evaluator.
- Pagination removed from IntegrationConfig and AiProvider list endpoints
  (low-cardinality, user-scoped entities).
- Domain exceptions (`BadRequestException`, `ResourceNotFoundException`) used
  consistently instead of `ResponseStatusException`.

## Frontend additions

- State services (`IntegrationConfigStateService`,
  `AiProviderConfigStateService`) that hide the
  `?namespaceId=none` sentinel, dispatch on scope, and keep a `vm$` multicast
  view of the 3 sources with refresh fan-out.
- New `NamespaceRoleStateService` resolving the current user's effective role
  on a namespace (super-admin short-circuit, otherwise checks the namespace
  user list, default-safe to false on error).
- New components: `IntegrationsAllScopesComponent`, `AiProvidersAllScopesComponent`.
  Existing `*-form` and `*-item` components modernized to signal-based inputs
  and scope-aware behavior.

## Migration / compatibility

- Existing `IntegrationConfig` rows are unchanged: they keep `userId IS NULL`
  and `namespaceId` non-null. All current `findByParent(namespaceId)` queries
  return the same rowset as before. No data migration required.
- AiModel is excluded from the overlay mechanism in this PR (managed at
  namespace level only, standard CRUD unchanged).

## Test plan

- [ ] Login as a non-admin user, no overrides — sections render, only the
      `namespace` section's items show; create button opens the form on
      `?scope=userOnNs`; submit creates a userOnNs override that appears in
      "Mes overrides sur ce namespace".
- [ ] Click Duplicate on a namespace card → form pre-fills (without apiKey for
      providers) → toggle radio to userGlobal → submit → override appears in
      "Mes overrides globaux".
- [ ] As super-admin, edit a namespace-shared item → save → list reflects the
      change.
- [ ] Edit an AiProvider override, leave apiKey untouched, change name only,
      save → network shows PUT without the `apiKey` key in the body.
- [ ] Edit an AiProvider override, clear apiKey, save → PUT sends `apiKey: null`
      → DB row has no credential.
- [ ] Open an agent in chat with the namespace DATETIME integration set to
      `Asia/Tokyo`, then create a `userOnNs` override of DATETIME with
      `Europe/Paris` and ask the agent the time → the agent gets Paris.
- [ ] Visit `/me`, expand the recap, delete a user-global override → the row
      disappears (and is also gone from the namespace pages).


vincent-audibert-whoz:
Less intimidating when looking into it but some general directions should be corrected or discussed:
- Level priorities: it should rather be `user` < `namespace` < `namespace & user`
- Existing APIs should be changed to incorporate this feature rather than adding new controllers and resources

...and these directions seem to come from subtle ambiguous cues in the initial ticket that were missed at write and review time of the ticket 😭 .

Regarding the levels, we could have a `none` level where no userId nor namespaceId is referenced, that would serve as "platform-wide" configuration element and prove useful to administrate multi-tier setup.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelController.kt
```diff
@@ -102,18 +127,176 @@ class AiModelController(
...
+     * would be invisible through this batch endpoint even though `GET /{id}` returns them via
+     * the evaluator's ownership branch. The override fetches all rows once and unions
+     * membership-visible ids with rows the caller owns — bounded by input size.
+     */
+    @PostMapping(
+        "/by-ids",
+        consumes = [MediaType.APPLICATION_JSON_VALUE],
+        produces = [MediaType.APPLICATION_JSON_VALUE],
+    )
+    @PreAuthorize("isAuthenticated()")
```
vincent-couturier-whoz: question: why this Preauthorize since we can define prefix like '/api' to be filtered in security configuration to ensure the call is authenticated ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigReconciliationService.kt
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * Open class with explicit constructor injection: per-entity beans are wired by
+ * [io.whozoss.agentos.reconciliation.ReconciliationConfiguration]. Spring cannot
+ * auto-resolve `ConfigLookup<T>` / `MergeStrategy<T>` from a single
+ * `@Service ConfigReconciliationService<T>` declaration because Kotlin's type erasure leaves
+ * the runtime container with no way to choose between competing parameterised beans — RFC Q7.
+ *
+ * @param T the entity type being reconciled
+ */
+open class ConfigReconciliationService<T : Entity>(
```
vincent-couturier-whoz: suggestion : we should avoid having open classes and use inheritance. It seems here that the class is not explicitly specialized in subclass. Perhaps we could ad a warning to avoid somebody from creating unwanted subclasses

vincent-couturier-whoz: suggestion: we talk about MergeStrategy so we could name the ConfigReconciliationService  just ConfigMergeService

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ReconciliationConfiguration.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Wires per-entity beans of [ConfigReconciliationService].
+ *
+ * Spring cannot derive parameterised reconciliation beans from a single `@Service` declaration
+ * (Kotlin type erasure: the container has no way to disambiguate `ConfigLookup<IntegrationConfig>`
+ * from `ConfigLookup<AiProvider>` at injection time). Each entity therefore registers
+ * its bean explicitly here.
+ */
+@Configuration
+class ReconciliationConfiguration {
```
vincent-couturier-whoz: suggestion: MergeConfiguration

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ReconciliationSchemaInitializer.kt
```diff
@@ -0,0 +1,60 @@
...
+ * Ordered after [io.whozoss.agentos.integrationConfig.IntegrationConfigSchemaInitializer]
+ * (which runs at the default order) so any Neo4j session is already established.
+ */
+@Component
+@Order(20)
+@ConditionalOnExpression(
+    "'\${agentos.persistence.mode:in-memory}' == 'neo4j' " +
+        "or '\${agentos.persistence.mode:in-memory}' == 'embedded-neo4j'",
+)
+class ReconciliationSchemaInitializer(
```
vincent-couturier-whoz: suggestion : we should follow the init process we already have on Neo4jSchemaInitializer. We could improve the structure of Neo4jSchemaInitializer to ensure Open/Close principe but I don't want a new system for each new feature.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
```diff
@@ -19,14 +23,25 @@ import java.util.UUID
 @RequestMapping("/api/user-groups", produces = [MediaType.APPLICATION_JSON_VALUE])
 class UserGroupController(
     private val userGroupService: UserGroupService,
+    private val guard: UserGroupGuard,
 ) {
     @GetMapping
+    @PreAuthorize("isAuthenticated()")
```
vincent-couturier-whoz: suggestion: remove this preauthorize and use SecurityManager with a proper configuration

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
```diff
@@ -19,14 +23,25 @@ import java.util.UUID
...
+    @HideOnAccessDenied
     fun searchByNamespaceExternalId(
         @RequestParam namespaceExternalId: String,
-    ): List<UserGroupSearchResultResource> =
-        userGroupService
+        auth: Authentication,
+    ): List<UserGroupSearchResultResource> {
+        if (!guard.canRead(namespaceExternalId, auth)) {
```
vincent-couturier-whoz: suggestion: have a dedicated PR for such security concern which is not directly related to this PR

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupGuard.kt
```diff
@@ -0,0 +1,44 @@
...
+ * - Write (create) requires WRITE on the resolved namespace (= ADMIN).
+ *
+ * The HTTP API exposes `namespaceExternalId` (federation identifier) rather than the internal
+ * `namespaceId` UUID, so this guard resolves the external id to a namespace before the
+ * permission check. A namespace that does not exist (or has no row at all under that external
+ * id) yields `false` regardless of the caller — we never leak existence via a different status
+ * code.
+ */
+@Component
+class UserGroupGuard(
```
vincent-couturier-whoz: suggestion: add this in another PR for another ticket about security enhancement.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolRegistryService.kt
```diff
@@ -153,6 +158,128 @@ class ToolRegistryService(
...
+     *   `findByNamespaceShared`), it is a **dormant user override** — the user persisted an
+     *   override that no longer matches any namespace-shared config. Swallow the exception,
+     *   warn, and skip the name (FR30: "remains dormant without raising an error").
+     * - When the failing `name` comes from `findByNamespaceShared` (the namespace itself
+     *   declares the integration), reconciliation MUST succeed. **Fail-closed**: rethrow so
+     *   the run aborts rather than silently producing a partial toolset (NFR-REL-1).
+     *
+     * [resolveToolsForNamespace] is preserved unchanged for legacy callers without userId (AC1).
+     */
+    fun resolveToolsForRun(
```
vincent-couturier-whoz: suggestion: Create a ToolResolver service to avoid adding more reponsibility to ToolRegistry.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/declarative/OwnershipResolver.kt
```diff
@@ -0,0 +1,35 @@
...
+ * evaluator is wired early in Spring Security ; the entity services arrive later.
+ * Keeping the dispatcher in a separate `@Component` lets Spring break the wiring
+ * chain cleanly. If `@Lazy` is needed in the evaluator's injection of this
+ * resolver, it can be added there without touching this class.
+ *
+ * Supports AI_PROVIDER, INTEGRATION_CONFIG, AI_MODEL — owners are the row's userId field,
+ * denormalized for AiModel from the parent provider at create time.
+ */
+@Component
+class OwnershipResolver(
```
vincent-couturier-whoz, line 15: suggestion: avoid this design which is not really compliant with Open/Close principle.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/RunReconciliationCache.kt
```diff
@@ -0,0 +1,38 @@
...
+ * orchestration layer. If we ever execute tool resolution concurrently within a run,
+ * wrap the [cache] map with ConcurrentHashMap.
+ *
+ * Defence-in-depth: the cache key includes `(namespaceId, userId)` even though a single
+ * cache instance lives for one `(namespaceId, userId)` run by construction. Should the
+ * lifecycle ever be broken (singleton bean, refactor, accidental sharing across runs),
+ * the scoped key prevents a cross-tenant config leak. Cost is one extra `UUID` per entry.
+ */
+class RunReconciliationCache {
+    private val cache = mutableMapOf<CacheKey, Any>()
```
vincent-couturier-whoz: suggestion : not sure if mutableMapOf will produce ConcurrentHashMap. Perhaps we could explicitly create a ConcurrentHashMap. Furthermore we already have "custom" cache mecanism. Why we couldn't use standard spring mecanism which could be monitored and managed easierly ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigReconciliationService.kt
selimbensenouciep-prog: Done in d4b7f6d3 — renamed `ConfigReconciliationService` → `ConfigMergeService` (class + file + 33 references across callers, tests, and docs).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ReconciliationConfiguration.kt
selimbensenouciep-prog: Done in d4b7f6d3 — renamed `ReconciliationConfiguration` → `MergeConfiguration` (class + file + 4 references).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
selimbensenouciep-prog: The whole UserGroup authz hardening (including this `@PreAuthorize`) is reverted in fd7cd93f, following your comments #7 and #8 — moved out of this PR's scope. The broader suggestion to use a global SecurityManager filter chain rather than per-endpoint `@PreAuthorize` is noted as a transverse refactor: it deserves a dedicated security ticket that arbitrates the choice across all controllers in the codebase (this PR's AiProvider/AiModel/IntegrationConfig included). Not blocking here.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
selimbensenouciep-prog: Done in fd7cd93f — UserGroup authz hardening reverted from this PR. `git diff origin/master..HEAD` on `UserGroupController.kt` is now empty. The dedicated security ticket will land on Alan's UserGroup epic (#31424 / #31644) and may reuse the `UserGroupGuard` pattern as inspiration if useful.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupGuard.kt
selimbensenouciep-prog: Done in fd7cd93f — `UserGroupGuard.kt` deleted from this PR. Will be reintroduced in a dedicated security ticket on the UserGroup epic, where the architectural choice between per-endpoint guards and a global SecurityManager filter chain (your comment #6 above) can be arbitrated.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/RunReconciliationCache.kt
selimbensenouciep-prog: Partial fix in d4b7f6d3 — switched the backing map from `mutableMapOf` (which returns a `LinkedHashMap`, not thread-safe) to `ConcurrentHashMap<CacheKey, Any>()`. The KDoc was updated accordingly: the warning *"NOT thread-safe by design — single-threaded by Spring AI"* is gone, and the cache is now safe by construction regardless of the execution model. On the broader question of using a standard Spring cache mechanism (monitoring, eviction): agreed it's a better long-term direction, but that's an architectural refactor with its own surface that I'd prefer to handle in a dedicated ticket rather than in this PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -82,20 +142,194 @@ class IntegrationConfigController(
...
+            "`userId` accepts ONLY the literal sentinel `me` — a UUID returns 400 (cross-user " +
+            "listing is not exposed). `namespaceId=none` is the sentinel for `namespaceId IS NULL`. " +
+            "Pagination defaults to page=0, size=20 ; size is capped at 100.",
+    )
+    @GetMapping
+    @PreAuthorize("isAuthenticated()")
+    fun list(
+        @RequestParam(required = false) namespaceId: String?,
+        @RequestParam(required = false) userId: String?,
+        @RequestParam(defaultValue = "0") page: Int,
```
vincent-audibert-whoz: Pagination is probably overkill for IntegrationConfig. Also, this brings some heterogenous apis, we should rather start paginating in the CaseEvent and Case controllers that will handle hundreds/thousands of entities.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigMergeService.kt
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Thrown by [ConfigMergeService.resolve] when none of the three precedence layers
+ * contains a configuration for the requested triple.
+ *
+ * Carries the failing triple in both the structured fields (for typed handling upstream) and
+ * the message (for log readability). The fail-closed posture (NFR-REL-1) requires this typed
+ * exception over a `null` return — callers MUST observe an explicit signal.
+ */
+class ConfigNotFoundException(
```
vincent-audibert-whoz: To move in exception package and kind of already exists (`ResourceNotFoundException`).

agentos/docs/user-level-overlays.md
vincent-audibert-whoz, line 1: Document waaaay too detailed for this folder (injected in Dev agent for agentos context), it should rather be around 10-20 lines, no graph or minimal.

agentos/agentos-service/build.gradle.kts
```diff
@@ -346,7 +346,7 @@ openApi {
     waitTimeInSeconds.set(60)
     // Activate the openapi profile so the app starts without real AI API keys
     customBootRun {
-        args.set(listOf("--spring.profiles.active=openapi", "--server.port=$agentosPort"))
+        args.set(listOf("--spring.profiles.active=openapi,embedded-neo4j", "--server.port=$agentosPort"))
```
vincent-audibert-whoz, line 352: Really necessary ? Don’t we have the embedded neo4j by default ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -82,20 +142,194 @@ class IntegrationConfigController(
...
+     * [io.whozoss.agentos.aiProvider.AiProviderController.getByIds]).
+     */
+    @PostMapping(
+        "/by-ids",
+        consumes = [MediaType.APPLICATION_JSON_VALUE],
+        produces = [MediaType.APPLICATION_JSON_VALUE],
+    )
+    @PreAuthorize("isAuthenticated()")
+    override fun getByIds(@RequestBody ids: List<UUID>): List<IntegrationConfigResource> {
+        if (ids.size > MAX_BATCH_SIZE) {
```
vincent-audibert-whoz: overkill on IntegrationConfig, the API limits should be setup first on caseevents and cases to establish the pattern, another story.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -82,20 +142,194 @@ class IntegrationConfigController(
...
+            throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Batch size ${ids.size} exceeds maximum of $MAX_BATCH_SIZE")
+        }
+        if (ids.isEmpty()) return emptyList()
+
+        val currentUser = userService.getCurrentUser()
+        val membershipVisibleIds: Set<UUID> = if (currentUser.isAdmin) {
+            ids.toSet()
+        } else {
+            val rawVisible = permissionService.filterVisibleIds(
```
vincent-audibert-whoz, line 160: Side note: permissionService should handle UUIDs as that is the basics of id in AgentOS, no ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -82,20 +142,194 @@ class IntegrationConfigController(
...
+    fun list(
+        @RequestParam(required = false) namespaceId: String?,
+        @RequestParam(required = false) userId: String?,
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "20") size: Int,
+        auth: Authentication,
+    ): IntegrationConfigPage {
+        val safeSize = size.coerceIn(1, MAX_PAGE_SIZE)
+        val safePage = page.coerceAtLeast(0)
+        val nsFilter = parseNamespaceFilter(namespaceId)
```
vincent-audibert-whoz: The whole interpretation of the list request should be in the service layer, controller just there to validate, pass on and convert output.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -82,20 +142,194 @@ class IntegrationConfigController(
...
-    override fun create(@Valid @RequestBody resource: IntegrationConfigResource): IntegrationConfigResource =
-        super.create(resource)
+    @PreAuthorize("isAuthenticated()")
+    @ResponseStatus(HttpStatus.CREATED)
+    override fun create(@Valid @RequestBody resource: IntegrationConfigResource): IntegrationConfigResource {
+        val me = currentUserId(currentAuth())
+
+        // Phase 1 — mass-assignment guard
+        if (resource.userId != null && resource.userId != me) {
+            throw ResponseStatusException(
```
vincent-audibert-whoz: Needs to be in the exception package, `BadRequestException` ?

Same for other exceptions below.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -107,7 +341,86 @@ class IntegrationConfigController(
...
-    companion object : KLogging()
+    private fun callerCanReadNamespace(auth: Authentication, namespaceId: UUID): Boolean =
+        permissionService.hasPermission(
+            userId = currentUserId(auth).toString(),
+            entityType = EntityType.NAMESPACE,
+            entityId = namespaceId.toString(),
+            action = Action.READ,
+        )
+
+    private fun currentUserId(auth: Authentication): UUID {
```
vincent-audibert-whoz: Don’t we have already a service for that ? Should not be repeated in every controller.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -107,7 +341,86 @@ class IntegrationConfigController(
...
+    private fun currentAuth(): Authentication =
+        SecurityContextHolder.getContext().authentication
+            ?: throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "Missing authentication")
+
+    private fun parseNamespaceFilter(raw: String?): NamespaceFilter = when {
+        raw == null -> NamespaceFilter.Any
+        raw.equals(NONE_SENTINEL, ignoreCase = true) -> NamespaceFilter.UserGlobalOnly
+        else -> {
+            val parsed = runCatching { UUID.fromString(raw) }
+                .getOrElse { throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid namespaceId: '$raw'") }
```
vincent-audibert-whoz: UUID not validated upstream by Spring ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/TripleKeyEncoding.kt
```diff
@@ -0,0 +1,72 @@
...
+ * - lets `delete()` rewrite the key to a per-id [tombstone] form so the unique slot is
+ *   freed and the user can immediately recreate `(ns, user, name)` after a soft-delete;
+ * - turns concurrent creates with the same triple into a database-level 409 instead of
+ *   silently producing duplicates.
+ *
+ * Cf. RFC `rfc-user-level-overlays.md` §D11 (chunk 1a).
+ *
+ * NOT thread-unsafe — pure functions, no state.
+ */
+object TripleKeyEncoding {
```
vincent-audibert-whoz: I really like the idea, much less the name 😅

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
```diff
@@ -40,8 +63,143 @@ class IntegrationConfigServiceImpl(
...
+                            it.namespaceId != null &&
+                            it.name == name &&
+                            it.integrationType != type
+                    }
+                    ?.let(::reject)
+            }
+        }
+    }
+
+    private fun saveOrConflict(entity: IntegrationConfig): IntegrationConfig =
```
vincent-audibert-whoz, line 185: I may have missed something, but the triplekey unicity should be enforced in the neo4j constraint and not managed here by the service layer, no ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigMergeService.kt
selimbensenouciep-prog: Moved to `exception/` package. Kept as a distinct class, it carries the resolution triple `(namespaceId, userId, name)` for fail-closed diagnostics, unlike the generic `ResourceNotFoundException`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: Agreed, removed from all 3 controllers. Will establish the pattern on high-volume endpoints first (caseevents/cases) in a dedicated story.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: True, the whole PermissionService interface uses String ids. Every caller in the codebase does `.id.toString()` (CaseController, CaseEventGuard, EntityController, etc.). Refactoring the interface to accept UUID is a cross-cutting change, tracked in #869.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: Done. Extracted filter dispatch into `findFiltered()` on all 3 service interfaces (AiProvider, AiModel, IntegrationConfig). Controllers now only validate params, call the service, paginate and convert.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: Created `BadRequestException` in the exception package and replaced the inline `ResponseStatusException(BAD_REQUEST)` throws across all 3 controllers.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: Replaced with `userService.getCurrentUser().id` which is already available via EntityController. Removed the private `currentUserId()` and `currentAuth()` helpers from AiProvider and AiModel controllers.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: The `namespaceId` param is typed `String?` because it also accepts the `none` sentinel (user-global scope). Typing it as `UUID?` would make Spring reject `?namespaceId=none` with a 400 before we even reach the controller body. So the manual parse after ruling out the sentinel is intentional.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/TripleKeyEncoding.kt
selimbensenouciep-prog: Ha, open to suggestions! Candidates: `ScopeKeyEncoder`, `OverlayKeyEncoder`, `CompositeKeyEncoding`? The 'triple' refers to the (namespaceId, userId, name) tuple but happy to rename if you have a preference.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
selimbensenouciep-prog: The DB constraint enforces uniqueness of the exact triple `(namespaceId, userId, name)`. The check at L158 (`assertConsistentIntegrationTypeAcrossLayers`) is a different rule: it validates cross-layer consistency, e.g. if `(ns=X, user=null, name=FOO)` has `integrationType=A`, then `(ns=X, user=Y, name=FOO)` can't have `integrationType=B` because the merge would be incoherent. That cross-layer rule spans multiple nodes with different nullable components, which Neo4j can't enforce as a native constraint. The lookups are indexed (tripleKey unique = index seek), so the cost is bounded. Will add a comment to clarify the intent.

agentos/docs/user-level-overlays.md
selimbensenouciep-prog: trimmed to ~13 lines: removed both tables, kept only the 3-tier precedence list and scope inference summary. This folder is injected into the dev agent context so brevity matters.

agentos/agentos-service/build.gradle.kts
selimbensenouciep-prog: Yes, it is needed. There are two distinct concepts: the **property** `agentos.persistence.mode=embedded-neo4j` (which defaults correctly) and the **Spring profile** `embedded-neo4j` (which loads `application-embedded-neo4j.yml` providing `spring.neo4j.uri`). Without the profile, Spring Data Neo4j cannot construct `Neo4jConnectionDetails` and the forked bootRun fails. This is pre-existing tech debt from `59875f83` (April 2026, "neoj repositories") when Spring Data Neo4j was introduced. The `openapi` profile alone does not provide the required connection config. Discovered and fixed here during the OpenAPI regen task.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/declarative/OwnershipResolver.kt
selimbensenouciep-prog: Agreed. Refactored in `008e4135`: introduced `OwnershipAware` interface that entity services implement. `OwnershipResolver` now auto-collects all `OwnershipAware` beans via Spring injection, removing the hardcoded `when` block. Unit tests added in `1fde3dce`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
selimbensenouciep-prog: Agreed. Pagination removed from IntegrationConfig and AiProvider list endpoints in `32696978`. These are low-cardinality, user-scoped entities (single-digit counts per user per namespace). The pattern should be established first on Case/CaseEvent as you suggested.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolRegistryService.kt
selimbensenouciep-prog: Done in `604b3fdf`: extracted `ToolResolverService` from `ToolRegistryService`, keeping resolution logic separate from registration.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ReconciliationSchemaInitializer.kt
selimbensenouciep-prog: Done in `0b4f17d9`: merged `ReconciliationSchemaInitializer` into the per-entity initializers (`AiProviderSchemaInitializer`, `IntegrationConfigSchemaInitializer`), following the existing `Neo4jSchemaInitializer` pattern rather than introducing a separate initializer.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelController.kt
selimbensenouciep-prog: The `@PreAuthorize` annotations here go beyond simple authentication: they use `hasPermission(#id, 'AiModel', 'READ')` and custom guards like `@aiModelGuard.canCreate(#resource)` for entity-level authorization. A global `/api` prefix filter only covers authentication (is the caller logged in?), not fine-grained permission checks (can this user access this specific entity?). Both layers complement each other.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigReconciliationService.kt
selimbensenouciep-prog: Good catch. The `open` modifier was removed in `0ef3768f`. `ConfigMergeService` is now a final class (Kotlin default) with a generic type parameter `<T : Entity>`, not subclassed anywhere.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/TripleKeyEncoding.kt
selimbensenouciep-prog: Renamed to `OverlayKeyEncoding` in `0cba9224` (class, file, and all 5 references). Ties directly to the "user-level overlay" domain concept used across the codebase.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    fun resolveToolsForRun(
+        namespaceId: UUID,
+        userId: UUID,
+        cache: RunReconciliationCache? = null,
+        agentIntegrations: Map<String, List<String>?>? = null,
+    ): Collection<StandardTool<*>> {
+        val resolved = mutableMapOf<String, StandardTool<*>>()
+        val pluginsByType = toolRegistryService.getPluginsByType()
```
vincent-audibert-whoz: Side note for later: ToolRegistryService should expose a `findPlugin(type: String)` instead of this reference to the map ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -75,23 +78,15 @@ class AgentServiceImpl(
...
-                    val defaultModel =
-                        findDefaultModelConfig(context.namespaceId)
-                            ?: throw IllegalArgumentException(
-                                "AgentConfig '${config.name}' has no modelName and no default AiModel is configured " +
-                                    "for namespace ${context.namespaceId}.",
-                            )
-                    defaultModel to aiProviderService.getById(defaultModel.aiProviderId)
-                }
-            }
+        val cache = if (context.userId != null) RunReconciliationCache() else null
```
vincent-audibert-whoz: Cache is not useful in my understanding, as scoped to the agent instantiation for AiProvider and IntegrationConfigs, but both types are queried only once per agent instantiation and retrieval key.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderLookup.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * [ConfigLookup] adapter for [AiProvider].
+ *
+ * Trivial wrapper around [AiProviderRepository.findByTriple] — kept as a separate bean
+ * so the generic [io.whozoss.agentos.reconciliation.ConfigMergeService] can be wired
+ * by [io.whozoss.agentos.reconciliation.MergeConfiguration] without leaking the
+ * repository contract into the reconciliation package.
+ */
+@Component
+class AiProviderLookup(
```
vincent-audibert-whoz: [Optional]: Why not have `AiProviderServiceImpl` implement the `ConfigLookup<AiProvider>` ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderMergeStrategy.kt
```diff
@@ -0,0 +1,32 @@
...
+ * - `apiKey`: override wins when non-null and non-blank; otherwise `base.apiKey` is preserved.
+ *   This mirrors [AiProviderController]'s `resolveApiKey` 4-way contract — a blank or null override must not
+ *   accidentally zero-out the effective API key (would cause a 401 at the provider, story 6.4 AC8).
+ * - `baseUrl`: override wins when non-null and non-blank.
+ * - `description`: override wins when non-null.
+ * - `apiType`: override wins (in practice all layers share the same apiType).
+ */
+@Component
+class AiProviderMergeStrategy : MergeStrategy<AiProvider> {
+    override fun merge(
```
vincent-audibert-whoz, line 22: [Optional]: have mergeable entities implement a `Mergeable` interface directly ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderResource.kt
```diff
@@ -9,21 +9,30 @@ import java.util.UUID
...
- * entities via their own endpoints — they are not embedded in this resource.
+ * Models are managed as independent `AiModel` entities via their own endpoints — they
+ * are not embedded in this resource.
  */
 @Schema(name = "AiProvider")
 data class AiProviderResource(
     val id: UUID? = null,
+    // OpenAPI 3.1 nullable encoding : `type: [string, "null"]`. The simpler `nullable = true`
+    // attribute is OpenAPI 3.0-only and silently dropped by SpringDoc 2.x in 3.1 mode, leaving
+    // generated SDKs typing the field as `string | undefined` instead of `string | null` (F3).
```
vincent-audibert-whoz: [Optional] ok for handling that but the JS side won’t see the difference (`undefined` is as falsy as `null`), neither the KT side after Jackson, really useful ? Was it a problem ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigLookup.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * [ConfigLookup] adapter for [IntegrationConfig].
+ *
+ * Trivial wrapper around [IntegrationConfigRepository.findByTriple] — kept as a separate bean
+ * so the generic [io.whozoss.agentos.reconciliation.ConfigMergeService] can be wired
+ * by [io.whozoss.agentos.reconciliation.MergeConfiguration] without leaking the
+ * repository contract into the reconciliation package.
+ */
+@Component
+class IntegrationConfigLookup(
```
vincent-audibert-whoz: Same as `AiProviderLookup`, weird to have it not as an interface implemented by the existing service ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        pluginsByType.values
+            .filter { it.configSchema == null }
+            .forEach { plugin ->
+                try {
+                    val tools = plugin.provideTools(null)
```
vincent-audibert-whoz: For later : This instantiation is problematic because it tries to instantiate all tools regardless of their usage by the agent.

The whole process should probably reversed:
- start from the agent integration names
- retrieve the integrationConfigs by names and scope for merge (have the merge done in the IntegrationService ?)
- then get the plugins from the resolved integrationConfigs
- instantiate and get the filtered tools.

=> create a ticket/issue for that.

vincent-audibert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
selimbensenouciep-prog: You're right, in the current code each reconciled name resolves exactly once per agent: applyOverlaysToModel handles one provider per agent, and resolveToolsForRun iterates a Set of distinct names. The cache was defence-in-depth for cross-agent sharing that doesn't happen, so it just added indirection. Removed in 8151c107.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderLookup.kt
selimbensenouciep-prog: Good point, the wrapper just duplicated findByNamespaceAndUserAndName from the service. AiProviderService now implements ConfigLookup<AiProvider> directly, the method is renamed to findByTriple to match the interface name, and MergeConfiguration injects the service. Same change applied to IntegrationConfigLookup. Done in 8e1d0b8d.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigLookup.kt
selimbensenouciep-prog: Handled together with the AiProvider one: IntegrationConfigService now implements ConfigLookup<IntegrationConfig> directly, method renamed to findByTriple, wrapper deleted, MergeConfiguration wires the service. Done in 8e1d0b8d.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderMergeStrategy.kt
selimbensenouciep-prog: Fair alternative stylistically, but I'd lean against it here. AiProvider lives in agentos-sdk which is consumed by plugin modules (agentos-plugins-filesystem, agentos-datetime-plugin, etc.), so adding a Mergeable interface on the entity leaks a reconciliation concern into the plugin-facing SDK. The current MergeStrategy<T> keeps that orchestration concern inside the service. IntegrationConfig is service-side so it doesn't have that constraint, but keeping a symmetric pattern across both entities means one contract to reason about. Happy to revisit if AiProvider ever moves out of the SDK.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiProvider/AiProviderResource.kt
selimbensenouciep-prog: Just tested removing them. SpringDoc 2.x in OpenAPI 3.1 mode then generates a broken schema for IntegrationConfig — userId disappears entirely from the schema and namespaceId becomes required, even though both are nullable Kotlin fields with default null. So the annotations are a SpringDoc workaround, not just type-honesty cosmetics. Updated the code comment to reflect the real reason in f8f84427.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
selimbensenouciep-prog: Agreed on the shape. Same pattern applies to resolveToolsForNamespace (the legacy anonymous-run path), which is master's existing enumerate-then-filter. The new bit our branch added is resolveToolsForRun's upfront config-less plugin loop, which makes config-less plugins auto-available in the user-aware path (master required an IntegrationConfig of the plugin's type to exist first). I'll track the reversal as a follow-up covering both methods together since the symmetric refactor is cleaner. The config-less question (always-on vs opt-in) deserves its own design call before refactoring.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
selimbensenouciep-prog: Done in 839c5d41. ToolRegistryService now exposes findPlugin(integrationType) for point lookups and findConfigLessPlugins() for the iteration used in resolveToolsForRun. The internal pluginsByType map is no longer leaked.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupController.kt
selimbensenouciep-prog: Agreed — handled in a dedicated security-config PR (your own suggestion below).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelController.kt
selimbensenouciep-prog: The @PreAuthorize calls here aren't URL-pattern auth, they're fine-grained per-resource permission checks (hasPermission(#id, 'AiModel', 'READ'), aiModelGuard.canCreate(#resource), aiModelGuard.canListByProvider(#parentId)) that depend on the resource ID extracted from the path. A SecurityConfig-level /api/** filter can only enforce authentication or role-based URL gates, not "user X can perform action A on resource Y where Y comes from the path". The two layers are complementary rather than substitutable.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigReconciliationService.kt
selimbensenouciep-prog: The file was renamed to ConfigMergeService and the class is now `class` (final), no longer `open`.

selimbensenouciep-prog: Done, renamed to ConfigMergeService.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -0,0 +1,155 @@
...
+                    tools.forEach { tool ->
+                        if (resolved.containsKey(tool.name)) {
+                            logger.warn { "[ToolResolver] Tool name conflict: '${tool.name}' from config-less plugin" }
+                        }
+                        resolved[tool.name] = tool
+                    }
+                } catch (e: Exception) {
+                    logger.error(e) { "[ToolResolver] Error instantiating config-less tools for plugin '${plugin.integrationType}': ${e.message}" }
+                }
+            }
```
vincent-audibert-whoz: I missed that on previous reviews 😢 , this block gives to all agents, all the tools that belong to configless plugins. This must not happen.

We need to be intentional in our allowance of tools to the agents, that goes strictly through:
- agent config referencing an integration name
- integration config existing for the said name, even if the integration comes from a plugin with null config schema.

vincent-audibert-whoz (changes requested):

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/tool/ToolResolverServiceSpec.kt
```diff
@@ -309,4 +304,166 @@ class ToolRegistryServiceSpec : StringSpec({
...
+        every { reconciliationService.resolve(any(), any(), any()) } answers {
+            val name = thirdArg<String>()
+            reconciledConfigs[name]
+                ?: throw ConfigNotFoundException(firstArg(), secondArg(), name)
+        }
+
+        return ToolResolverService(registry, integrationConfigService, reconciliationService)
+    }
+
+    "resolveToolsForRun returns config-less tools when no shared or user configs exist" {
```
vincent-audibert-whoz: Change test to expect no tools when no configuration exists.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
selimbensenouciep-prog: Fixed in 993d41b3. Removed the upfront config-less loop in resolveToolsForRun. All tools, config-less or configured, now flow through IntegrationConfig only, matching resolveToolsForNamespace and master. Two specs adapted: one asserts the new empty-when-no-IntegrationConfig behaviour, the other adds a DATETIME IntegrationConfig so the combined-tools test holds for the right reason.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/tool/ToolResolverServiceSpec.kt
selimbensenouciep-prog: Done in 993d41b3, the test at this line was rewritten to assert tools.shouldBeEmpty() when no IntegrationConfig exists, even with a config-less plugin loaded. Title updated accordingly.

vincent-audibert-whoz (approved):

PR #851, opened 2026-05-07
feat: #31644 Implement DELETE user group endpoint

AlanMenantWhoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
```diff
@@ -33,19 +33,19 @@ class Neo4jSchemaInitializer(private val driver: Driver) {
         driver.session().use { session ->
             session.run(
                 "CREATE CONSTRAINT user_group_name_namespace_unique IF NOT EXISTS " +
-                    "FOR (g:UserGroup) REQUIRE (g.name, g.namespaceId) IS UNIQUE",
+                    "FOR (g:ActiveUserGroup) REQUIRE (g.name, g.namespaceId) IS UNIQUE",
```
vincent-couturier-whoz, line 36: suggestion: I just realized that we could have a AllUser label (which contains User node and RemovedUser - which could stay implicit - nodes) and keep User label to avoid changing the common request on User and have the unicity constraint on User so the soft delete is just the set or unset of the User label on an AllUser node. The same could be applied on the other main label like Namespace and Group

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
```diff
@@ -33,19 +33,19 @@ class Neo4jSchemaInitializer(private val driver: Driver) {
...
                 "CREATE CONSTRAINT namespace_external_id_unique IF NOT EXISTS " +
-                    "FOR (n:Namespace) REQUIRE n.externalId IS UNIQUE",
+                    "FOR (n:ActiveNamespace) REQUIRE n.externalId IS UNIQUE",
             )
             logger.info { "[Neo4jSchemaInitializer] Constraint namespace_external_id_unique ensured" }
 
             session.run(
                 "CREATE CONSTRAINT user_external_id_unique IF NOT EXISTS " +
-                    "FOR (u:User) REQUIRE u.externalId IS UNIQUE",
+                    "FOR (u:ActiveUser) REQUIRE u.externalId IS UNIQUE",
```
vincent-couturier-whoz, line 48: issue: if we do that we should certainly change the create service in order to set the newly created node "active". In case the previous suggestion is applied it would be the same thing knowing that we have two labels instead of just one

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceServiceImpl.kt
```diff
@@ -41,8 +44,10 @@ class NamespaceServiceImpl(
 
     override fun findByExternalId(externalId: String): Namespace? = namespaceRepository.findByExternalId(externalId)
 
+    @Transactional
     override fun delete(id: UUID): Boolean = namespaceRepository.delete(id)
```
vincent-couturier-whoz, line 48: suggestion: it's not a real delete now so perhaps we should change the name to softDelete to be more explicit on the real behavior.

vincent-couturier-whoz (approved):
Not sure of the Active/Inactive wording because soft delete is more around removed or deleted status.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceServiceImpl.kt
AlanMenantWhoz: Vu ensemble, on repassera là dessus après avoir livré les user groups

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
AlanMenantWhoz: Vu ensemble, on repassera là dessus après avoir livré les user groups

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
AlanMenantWhoz: Vu ensemble, on repassera là dessus après avoir livré les user groups

PR #852, opened 2026-05-07
feat: #31681 REDIRECT internal integration tool

vincent-audibert-whoz:
## Summary

Adds the REDIRECT internal integration: an agent can now delegate the current request to another agent by calling the `redirect` tool.

### Feature
- `RedirectTool` — `StandardTool` that throws `RedirectRequestException` as a control-flow signal; the eligible agent list is built at resolution time from namespace configs filtered by glob patterns
- `RedirectToolPlugin` — Spring-managed `ToolPlugin` (no PF4J) wired with an `AgentConfigService`-backed resolver lambda
- `RedirectConfiguration` — `@Configuration` that breaks the circular dependency between `AgentServiceImpl` → `ToolRegistryService` → `RedirectToolPlugin` → `AgentConfigService`
- `AgentSimple` — catches `RedirectRequestException` and emits `AgentFinishedEvent` then `AgentSelectedEvent` (in that order, so `CaseRuntime.processNextStep` finds the selection as the newest event and launches the target agent)

### Bug fixes
- `AgentConfigController.toDomainForUpdate` — `integrations` field was missing from the `copy()` call, causing any PUT to silently discard integration changes
- `AgentSimple` redirect emission order — `AgentSelectedEvent` was emitted before `AgentFinishedEvent`, causing `processNextStep` to stop at `AgentFinishedEvent` before seeing the selection

### Tests
- `CaseRuntimeSpec` — "runAgent is called twice when agent A redirects to agent B" (regression for the orchestration fix)
- `AgentSimpleToolCallbackUnitSpec` — "on redirect, AgentFinishedEvent is emitted before AgentSelectedEvent" (pins the emission order contract)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
```diff
@@ -0,0 +1,87 @@
...
+              "required": ["agentName"]
+            }
+        """.trimIndent())
+    }
+
+    override val version: String = "1.0.0"
+    override val paramType: Class<Input> = Input::class.java
+
+    /**
+     * Always throws [RedirectRequestException] — this is the intended behaviour.
```
vincent-audibert-whoz: **Silent fallback on null input is surprising and hard to diagnose.**

```kotlin
override fun execute(input: Input?, context: ToolContext): String {
    val target = input?.agentName
        ?: eligibleAgents.firstOrNull()?.name
        ?: error("RedirectTool: no eligible agents configured")
    throw RedirectRequestException(target)
}
```

When the LLM sends a well-formed `agentName` argument the schema is satisfied. But when `input` is `null` (deserialisation failure, or LLM sends `{}` despite `required`), the tool silently redirects to the first eligible agent instead of surfacing the error. This is an implicit default that the LLM never requested and that will be invisible in logs — the redirect succeeds, but to an agent the user didn't intend.

The `eligibleAgents.firstOrNull()?.name` branch should be removed. A null `input` here is always a contract violation (the schema marks `agentName` as required); it should `error()` or throw a descriptive exception so the misconfiguration is visible.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolRegistryService.kt
```diff
@@ -135,7 +154,13 @@ class ToolRegistryService(
             }
             try {
                 val allowedToolNames = agentIntegrations?.get(config.name)
-                val providedTools = plugin.provideTools(config.parameters, config.name)
+                val toolContext = ToolContext(
+                    namespaceId = namespaceId,
+                    userId = null,
+                    userExternalId = null,
+                    caseEvents = emptyList(),
+                )
```
vincent-audibert-whoz, line 162: **`ToolContext` passed to `RedirectToolPlugin.provideTools` has `namespaceId` but it is constructed without a real namespace — this will cause `RedirectToolPlugin` to always return empty tools at resolution time.**

```kotlin
val toolContext = ToolContext(
    namespaceId = namespaceId,   // ← correct, passed in from the outer scope
    userId = null,
    userExternalId = null,
    caseEvents = emptyList(),
)
val providedTools = plugin.provideTools(config.parameters, config.name, toolContext)
```

Actually re-reading the full method, `namespaceId` *is* in scope here (it is the parameter of `resolveToolsForNamespace`), so the `ToolContext` is correctly populated. **However**, `RedirectToolPlugin.provideTools` uses `context?.namespaceId` to call `agentResolver(namespaceId, patterns)` which calls `agentConfigService.findByParent(namespaceId)`. This is called at agent *instantiation* time (startup of an agent run), not at tool *execution* time. The eligible agent list is therefore **fixed at the moment the tool set is built** and will not reflect agents added to the namespace after that point. This is documented nowhere and is likely to surprise operators.

This is a design choice more than a bug, but it should be explicitly documented on `RedirectToolPlugin.provideTools` (or `RedirectTool`) so it is clear that the whitelist is snapshot-at-instantiation, not live.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectRequestException.kt
```diff
@@ -0,0 +1,16 @@
...
+ * This is NOT an error condition. The tool emits a successful [ToolResponseEvent]
+ * before throwing so the event traces are complete and no failure is recorded.
+ * [io.whozoss.agentos.agent.AgentSimple] catches this exception before the generic
+ * error handler and emits [io.whozoss.agentos.sdk.caseEvent.AgentSelectedEvent] +
+ * [io.whozoss.agentos.sdk.caseEvent.AgentFinishedEvent] to trigger the next agent.
+ *
+ * @param targetAgentName The exact name of the agent to redirect to, as it appears
+ *   in the [io.whozoss.agentos.agentConfig.AgentConfig] of the namespace.
+ */
+class RedirectRequestException(val targetAgentName: String) : Exception("Redirect to '$targetAgentName'")
```
vincent-audibert-whoz: **`RedirectRequestException` extends `Exception` rather than a non-checked, non-`RuntimeException` signal type.**

```kotlin
class RedirectRequestException(val targetAgentName: String) : Exception("Redirect to '$targetAgentName'")
```

Extending `Exception` (checked in Java) means any future Java interop code, or any generic `catch (e: Exception)` added between the throw site and `AgentSimple`'s catch block, will inadvertently swallow the redirect signal. The existing generic error handler in `AgentSimple` is ordered *after* the `RedirectRequestException` catch, which is correct today, but the ordering is a fragile contract.

A control-flow signal like this should extend `RuntimeException` (or a dedicated sealed marker) to make the intent explicit and reduce the risk of accidental interception. The KDoc already says "NOT an error condition" — the class hierarchy should reflect that.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectConfiguration.kt
```diff
@@ -0,0 +1,68 @@
...
+        /**
+         * Converts a glob pattern (only `*` as wildcard) to a case-insensitive [Regex].
+         * All regex metacharacters in the pattern are escaped before substituting `*`.
+         */
+        internal fun globToRegex(pattern: String): Regex {
+            val escaped = pattern
+                .replace("\\", "\\\\")
+                .replace(".", "\\.")
+                .replace("[", "\\[")
+                .replace("]", "\\]")
```
vincent-audibert-whoz: **`?` wildcard in glob is mapped to `.` (any single character) but the KDoc only mentions `*`.**

```kotlin
.replace("?", ".")
.replace("*", ".*")
```

The `CONFIG_SCHEMA` description in `RedirectToolPlugin` says only `"*"` is the wildcard (`"Use \"*\" for all agents"`), but the implementation also silently supports `?` as a single-character wildcard. Either document `?` in the schema description, or remove its special handling to match the declared contract.

vincent-audibert-whoz (commented):
The overall design is sound and the two bug fixes are correct. Three issues deserve attention before merge: a semantic inconsistency in the `RedirectTool` fallback path, a `ToolContext` data mismatch at resolution time that will silently break `RedirectToolPlugin` in its most common use, and a `RedirectRequestException` class hierarchy choice that could interact poorly with future catch-all handlers. A minor glob escaping gap and a stale comment in `RedirectToolPlugin` are flagged inline.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -0,0 +1,102 @@
...
+import io.whozoss.agentos.sdk.tool.ToolPlugin
+import mu.KLogging
+import java.util.UUID
+
+/**
+ * Internal Spring-managed [ToolPlugin] that provides the REDIRECT integration.
+ *
+ * Unlike PF4J plugins, this class is NOT annotated with `@Component` — it is
+ * instantiated as a `@Bean` in [io.whozoss.agentos.agent.AgentServiceImpl] so that
+ * the [agentResolver] lambda can be wired without creating a circular Spring dependency.
```
vincent-audibert-whoz: **Stale KDoc: the class is no longer instantiated as a `@Bean` inside `AgentServiceImpl`.**

```
 * Unlike PF4J plugins, this class is NOT annotated with `@Component` — it is
 * instantiated as a `@Bean` in [io.whozoss.agentos.agent.AgentServiceImpl] so that
```

The `@Bean` was moved to `RedirectConfiguration` (which is the whole point of that class). The reference to `AgentServiceImpl` here is incorrect and will confuse future readers.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
vincent-audibert-whoz: null case removed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolRegistryService.kt
vincent-audibert-whoz: Documented.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectRequestException.kt
vincent-audibert-whoz: used RuntimeException

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectConfiguration.kt
```diff
@@ -0,0 +1,67 @@
...
```
vincent-audibert-whoz: Resolved by removal: the `?` handling has been dropped from `globToRegex`. Only `*` is now supported, which matches the declared contract in `CONFIG_SCHEMA`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectConfiguration.kt
```diff
@@ -0,0 +1,67 @@
...
+                .findByParent(namespaceId)
+                .filter { config -> regexes.any { it.matches(config.name) } }
+        }
+
+    companion object {
+        /**
+         * Converts a glob pattern (only `*` as wildcard) to a case-insensitive [Regex].
+         * All regex metacharacters in the pattern are escaped before substituting `*`.
+         */
+        internal fun globToRegex(pattern: String): Regex {
```
vincent-couturier-whoz: suggestion: Perhaps we should extract to another class / package to be able to reuse because it seems not specific to this context.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectRequestException.kt
```diff
@@ -0,0 +1,16 @@
...
+ * This is NOT an error condition. The tool emits a successful [ToolResponseEvent]
+ * before throwing so the event traces are complete and no failure is recorded.
+ * [io.whozoss.agentos.agent.AgentSimple] catches this exception before the generic
+ * error handler and emits [io.whozoss.agentos.sdk.caseEvent.AgentSelectedEvent] +
+ * [io.whozoss.agentos.sdk.caseEvent.AgentFinishedEvent] to trigger the next agent.
+ *
+ * @param targetAgentName The exact name of the agent to redirect to, as it appears
+ *   in the [io.whozoss.agentos.agentConfig.AgentConfig] of the namespace.
+ */
+class RedirectRequestException(val targetAgentName: String) : RuntimeException("Redirect to '$targetAgentName'")
```
vincent-couturier-whoz: suggestion: it is not really optimal to manage business rules by using exception. Perhaps we should manage it by a dedicated mecanism with a clear result which could "redirect" to a specific agent. Otherwise we could have unpredictable side effect which will be hard to debug. Furthermore we should be able to manage "clean tool result handling".... so it seems to be the correct moement to do it.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
```diff
@@ -0,0 +1,86 @@
...
+    data class EligibleAgent(val name: String, val description: String?)
+
+    override val name: String = configName?.let { "${it}__redirect" } ?: "redirect"
+
+    override val description: String = buildString {
+        appendLine("Route the current request to another agent. Use this when the request is better handled by a specialised agent.")
+        appendLine("Available agents:")
+        eligibleAgents.forEach { agent ->
+            val desc = agent.description?.takeIf { it.isNotBlank() }
+            if (desc != null) appendLine("  - ${agent.name}: $desc") else appendLine("  - ${agent.name}")
```
vincent-couturier-whoz: suggestion: `agent.description?.takeIf { it.isNotBlank() }?.also {}` instead of `if (desc != null)`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -0,0 +1,104 @@
...
+ * actually configured in the namespace via [agentResolver]. This is a snapshot: agents
+ * added to the namespace after the run starts will not appear in the redirect list for
+ * that run. Agents that do not exist in the namespace are silently excluded — the LLM
+ * never receives a stale or inaccessible name.
+ *
+ * @param agentResolver Lambda injected by [io.whozoss.agentos.agent.AgentServiceImpl]
+ *   to avoid a circular Spring dependency. Given a namespace UUID and a list of glob
+ *   patterns, returns the matching [AgentConfig]s from that namespace.
+ */
+class RedirectToolPlugin(
```
vincent-couturier-whoz, line 43: suggestion: perhaps we could/should have an InternalToolsPlugin to avoid creating plugin each we have to integrate tool for internal management....

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        val patterns = config
+            ?.get("agents")
+            ?.takeIf { it.isArray }
+            ?.map { it.asText() }
+            ?.filter { it.isNotBlank() }
+            ?.takeIf { it.isNotEmpty() }
+            ?: listOf("*")
+
+        val eligible = agentResolver(namespaceId, patterns)
```
vincent-couturier-whoz: suggestion: eligibleAgents

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolRegistryService.kt
```diff
@@ -39,6 +40,12 @@ class ToolRegistryService(
     private val pluginManager: PluginManager,
     private val integrationConfigService: IntegrationConfigService,
     private val integrationTypeRegistry: IntegrationTypeRegistry,
+    /**
+     * Spring-managed [ToolPlugin] beans (internal integrations such as REDIRECT).
+     * Registered alongside PF4J-loaded plugins; Spring injects all implementations
+     * automatically via the list constructor parameter.
+     */
+    private val springToolPlugins: List<ToolPlugin> = emptyList(),
```
vincent-couturier-whoz, line 48: question: If we know the plugin at build time perhaps we don't need to use plugin perhaps it's only a "ToolProvider" and so ToolPlugin would be only a kind of ToolProvider and SpringToolProvider would be the injected ones which are managed internally without the boilerplate needed by plugin mecanism

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentSimpleToolCallbackUnitSpec.kt
```diff
@@ -242,6 +245,52 @@ class AgentSimpleToolCallbackUnitSpec :
...
+        "on redirect, AgentFinishedEvent is emitted before AgentSelectedEvent" {
+            // CaseRuntime.processNextStep scans events newest-first. For it to launch
+            // the redirect target rather than stop, AgentFinishedEvent (current agent)
+            // must appear BEFORE AgentSelectedEvent (target agent) in the event list.
+            // This test pins that emission order as a contract.
+            val namespaceId = UUID.randomUUID()
+            val caseId = UUID.randomUUID()
+            val agentId = UUID.randomUUID()
+
+            val redirectTool = RedirectTool(
```
vincent-couturier-whoz, line 260: question:  Not sure to understand if we have a real configuration needed by redirect tool but it seems necessary to have an integration to be able to activate redirect on a specific agent. What would be the configuration of such integration ? It could a good candidate to be a platform integration to avoid too much configuration on the namespace to have such essential feature easily.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
vincent-couturier-whoz, line 1: question: I didn't see anything on the AgentAdvanced.kt. I don't know if you will do it on another PR or if it should be achieved in my story about Advanced execution. Who does the implementation of agent advanced ?

agents/Dev.yaml
```diff
@@ -8,13 +8,6 @@ integrations:
   AI:
   MEMORY:
   GITHUB:
```
vincent-couturier-whoz, line 10: suggestion: perhaps this change is not for this PR

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectRequestException.kt
vincent-audibert-whoz: Indeed but exceptions are used here to stop spring-ai from running again the agent after a redirect has been expressed. The observed behavior on Coday that I want to avoid on AgentOS is to have the agent still outputting all his thoughts and keep control of the conversation, before ending his run and letting the framework move on the redirected agent.

We will have some cases where the agent run is interrupted in the middle of a tool run (sync delegation, confirmation) and we then have to stop it. Looking into it again...

vincent-audibert-whoz: No better option found => let’s keep that, moving to `AgentInterrupt`.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentSimpleToolCallbackUnitSpec.kt
vincent-audibert-whoz: We will need a way to specify that agent A can redirect to every one ("*"), agent B to a select sub-agents, and that per-namespace to benefit from the IntegrationConfig scopes in the pipes. An integrationConfig to allow targeting every agent available to the user, is a good candidate for a platform-level definition.

agents/Dev.yaml
vincent-audibert-whoz: Tired of toolnames being tricky to resolve on the whitelist => allow all for now. Will be better later when Coday will use AgentOS and an upcoming capability to test the MCPs properly...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectConfiguration.kt
vincent-audibert-whoz: moved to its own utils file.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
vincent-audibert-whoz: Maybe, but for now as this integration does need a configuration, it was quite simple to implement it as a PF4J plugin and benefit from all the current plumbing.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
vincent-audibert-whoz: Added in this PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
```diff
@@ -209,6 +211,35 @@ class AgentSimple(
                         agentName = name,
                     ),
                 )
+            } catch (e: AgentInterrupt) {
```
vincent-couturier-whoz, line 209: suggestion: in another PR we should use a classical response to manage this business case which should not use exception as a way to "simplify" response "interception".

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
```diff
@@ -452,6 +483,25 @@ class AgentSimple(
                         result =
                             try {
                                 tool.executeWithJson(toolInput, context)
+                            } catch (e: AgentInterrupt) {
+                                // Interrupt is not an error: emit a successful response so traces
+                                // are complete, then re-throw the signal for the flow catch block.
+                                val message = when (e) {
+                                    is AgentInterrupt.Redirect -> "Redirecting to agent '${e.targetAgentName}'."
+                                }
+                                runBlocking {
```
vincent-couturier-whoz: suggestion: I think we should precise the coroutine context here to avoid bad surprises. (you could use : `CoroutineScope(Dispatchers.Default).launch {  runBlocking {...} }`)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterrupt.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * ## Why exceptions?
+ *
+ * [io.whozoss.agentos.sdk.tool.StandardTool] is called through Spring AI's
+ * [org.springframework.ai.tool.ToolCallback] interface, whose `call` contract returns
+ * a plain `String`. There is no way to signal an out-of-band interruption via a return
+ * value without leaking control-flow semantics into the tool result string. Throwing an
+ * exception is the only mechanism that exits `ToolCallback.call` without returning a
+ * result and propagates cleanly through Spring AI's internal tool-calling loop up to the
+ * [AgentSimple] catch block.
```
vincent-couturier-whoz: suggestion: [AgentSimple] or [AgentAdvanced]

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -163,6 +164,29 @@ class AgentAdvanced(
                         agentName = name,
                     ),
                 )
+            } catch (e: AgentInterrupt) {
+                // Not an error: a tool requested a structured interruption of this agent run.
+                when (e) {
```
vincent-couturier-whoz: suggestion: factorize this block and the one in AgentSimple in a function to avoid duplicated code

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
vincent-audibert-whoz: Yep, good idea.

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
vincent-audibert-whoz: moved to trySend that remains blocking and does not mess with asynchronicity for events.

vincent-couturier-whoz (approved):

PR #853, opened 2026-05-07
feat: #WZ-31782 LLM alias dropdown and field renames in AgentOS forms

vincent-audibert-whoz:
#### Field renames
- **AI model form**: `Max tokens` → `Max output tokens` (label only, `maxTokens` form control unchanged)
- **Agent config form**: `Model name` → `LLM model` (label rename, field converted to dropdown)

#### LLM model dropdown (`agent-config-form`)
- Replaced the free-text `Model name` input with an optional `<select>` dropdown
- Dropdown is populated from a new hand-written API service (`AiModelAliasService`) that calls `GET /api/ai-models/aliases-by-namespaceId/{namespaceId}`
- The first option is always `— default —` (maps to `null` in the form control), letting the backend apply the namespace's default alias when no alias is selected
- If the backend endpoint is unavailable or returns an error, `catchError` degrades gracefully to an empty alias list — only the default option is shown, no crash
- Existing agents with a previously saved `modelName` value load correctly in edit mode — the value is preserved in the form control and submitted as-is

#### New hand-written API service
- `libs/agentos-api-client/src/custom/ai-model-alias.service.ts` — follows the same `src/custom/` pattern as `CaseEventSseService`, safe from OpenAPI regeneration
- Exported from `libs/agentos-api-client/src/index.ts`

### Files changed
| File | Change |
|---|---|
| `libs/agentos-api-client/src/custom/ai-model-alias.service.ts` | **new** — hand-written HTTP service for aliases endpoint |
| `libs/agentos-api-client/src/index.ts` | export new service |
| `libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.ts` | inject `AiModelAliasService`, add `availableAliases` signal, load aliases in `forkJoin` |
| `libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.html` | replace text input with `<select>` dropdown, rename label |
| `libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.scss` | add `__select` and `__field-hint` styles |
| `libs/agentos-ui/src/lib/components/ai-model-form/ai-model-form.component.html` | rename `Max tokens` → `Max output tokens` label |

### Verified
- `pnpm nx lint agentos-api-client` ✅
- `pnpm nx lint agentos-ui` ✅
- `pnpm nx build client --configuration=development` ✅

### Note on backend endpoint
The `GET /api/ai-models/aliases-by-namespaceId/{namespaceId}` endpoint is not yet in the OpenAPI spec. `AiModelAliasService` is written by hand in `src/custom/` (safe from regeneration). The frontend degrades gracefully if the endpoint is not yet deployed — `catchError(() => of([]))` ensures the dropdown shows only the default option rather than crashing.


libs/agentos-api-client/src/custom/ai-model-alias.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Fetch the list of distinct AI model alias strings for a namespace.
+   *
+   * Endpoint: GET /api/ai-models/aliases-by-namespaceId/{namespaceId}
+   *
+   * Returns an empty array when the namespace has no aliases configured,
+   * or when the endpoint is not yet available (via catchError at call-site).
+   */
+  listAliasesByNamespace(namespaceId: string): Observable<string[]> {
```
vincent-audibert-whoz: URL mismatch with the backend route. The controller maps to `/aliases/by-namespaceId/{namespaceId}` under `/api/ai-models`, so the full path is `/api/ai-models/aliases/by-namespaceId/${namespaceId}`. This service calls `/api/ai-models/aliases-by-namespaceId/${namespaceId}` (hyphen instead of a path segment). The OpenAPI spec in this same PR uses the correct form. At runtime every call returns 404 and the dropdown silently falls back to the degraded state.

libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.ts
```diff
@@ -217,9 +246,9 @@ export class AgentConfigFormComponent implements OnInit {
       ...(this.existingConfig ?? {}),
       namespaceId: this.namespaceId,
       name: this.nameControl.value.trim(),
```
vincent-audibert-whoz, line 254: `||` was replaced with `??` for `description`, `modelName`, and `instructions`. For `description` and `instructions` this is a semantic change: `??` does not coerce an empty string `""` to `undefined`, so a user who clears one of these fields and saves now sends `""` instead of omitting the field. Verify that the backend handles `""` and `undefined`/absent identically for these fields; if not, this is a regression. `modelName` is fine because the control is `null` (not `""`) when the default option is selected.

libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.html
```diff
@@ -36,15 +36,16 @@ <h1 class="agent-config-form__title">{{ isEditMode() ? 'Edit agent config' : 'Ne
...
-          [formControl]="modelNameControl"
-          autocomplete="off"
-        />
+        <label class="agent-config-form__label" for="agent-config-model-name">LLM model</label>
+        <select id="agent-config-model-name" class="agent-config-form__select" [formControl]="modelNameControl">
+          <option [ngValue]="null">&mdash; default &mdash;</option>
+          @for (alias of availableAliases(); track alias) {
+            <option [value]="alias">{{ alias }}</option>
+          }
+        </select>
```
vincent-audibert-whoz: Binding inconsistency: the default option uses `[ngValue]="null"` (Angular value binding, sets control to `null`) while the alias options use `[value]="alias"` (DOM attribute binding, sets control to the string value of the attribute). For a `FormControl<string | null>`, mixing these two binding strategies causes inconsistent option-matching behaviour when Angular tries to determine the selected option. All options should use `[ngValue]` for uniform comparison semantics: `[ngValue]="alias"` on the alias options.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelController.kt
```diff
@@ -139,5 +139,41 @@ class AiModelController(
...
+                schema = io.swagger.v3.oas.annotations.media.Schema(type = "string"),
+            ),
+        )],
+    )
+    fun listAliasesByNamespaceId(
+        @PathVariable namespaceId: UUID,
+    ): List<String> =
+        aiModelService
+            .findByNamespaceId(namespaceId)
+            .mapNotNull { it.alias }
```
vincent-audibert-whoz, line 174: The Swagger annotations are written with fully-qualified class names (`@io.swagger.v3.oas.annotations.Operation`, `@io.swagger.v3.oas.annotations.responses.ApiResponse`, etc.) instead of adding the corresponding `import` statements. Every other annotation in this file uses the short imported form. Add the missing imports and use the short names for consistency with the file's established style.

libs/agentos-ui/src/lib/components/agent-config-form/agent-config-form.component.scss
```diff
@@ -66,6 +66,32 @@
...
+
+    &:focus {
+      border-color: var(--color-primary, #007aff);
+    }
+  }
+
+  &__field-hint {
+    font-size: 0.8rem;
+    color: var(--color-text-secondary, #6e6e73);
+    line-height: 1.4;
```
vincent-audibert-whoz: The `__select` block duplicates the entire `__input`/`__textarea` block (padding, border-radius, border, background, color, font-size, font-family, outline, box-sizing, transition, `:focus` rule). The only additions specific to `__select` are `cursor: pointer` and `appearance: auto`. Consider extending the existing grouped selector to include `__select` and declaring only the two distinct properties separately, to avoid the duplication and future drift between the two blocks.

vincent-audibert-whoz (commented):
## Review — feat: #WZ-31782 LLM alias dropdown and field renames

The overall approach is sound: the backend endpoint is well-designed and properly secured, the graceful degradation via `catchError` is correct, and the Angular conventions (standalone, `inject()`, signals, `@for`, `takeUntilDestroyed`) are all respected. A few concrete issues need attention before merge.

---

**URL mismatch between backend and frontend client (bug)**

The backend registers the route at `/aliases/by-namespaceId/{namespaceId}` under the `/api/ai-models` controller, giving a full path of `/api/ai-models/aliases/by-namespaceId/{namespaceId}`. The hand-written `AiModelAliasService` calls `/api/ai-models/aliases-by-namespaceId/${namespaceId}` (hyphen instead of `/aliases/`). The OpenAPI spec also uses the correct `/aliases/by-namespaceId/` path. This mismatch means every alias fetch returns 404 at runtime, leaving the dropdown permanently in degraded mode without any visible error.

**`||` → `??` semantic change in submit payload (regression risk)**

The description acknowledges this but frames it as a possible intentional change. It is not: for `description` and `instructions` the form controls are `FormControl<string | null>`, so their `value` can be a non-empty string, an empty string, or `null`. With `|| undefined`, an empty-string value would be coerced to `undefined` (field omitted from payload). With `?? undefined`, an empty string now passes through as `""` in the payload. Whether the backend treats `""` and `undefined`/absent identically for those fields needs to be confirmed. If the backend interprets `""` as "clear the field" rather than "leave unchanged", this is a silent regression for any user who blanks out a description or instructions and saves.

**`[value]` vs `[ngValue]` inconsistency on alias options (binding bug)**

The default option correctly uses `[ngValue]="null"` so Angular's reactive form sees a true `null`. The alias options inside `@for` use `[value]="alias"`, which binds a string attribute. For a `FormControl<string | null>`, mixing `[ngValue]` (on the default) with `[value]` (on the alias options) can produce inconsistent behaviour: when Angular compares the control's current value to the option list to determine which option is "selected", it uses reference equality for `[ngValue]` bindings and string coercion for `[value]` bindings. All options should use `[ngValue]="alias"` for consistent comparison semantics.

**Swagger annotations imported fully-qualified (style)**

`AiModelController.kt` already imports `MediaType` and all Spring/Jakarta annotations at the top of the file. The new `@Operation` and `@ApiResponse` annotations are written with full package qualifiers (`@io.swagger.v3.oas.annotations.Operation(...)`) rather than adding the missing imports. Every other existing annotation in the file uses the short imported form. This is inconsistent with the file's established style and reduces readability.

**`__select` styles duplicate `__input` / `__textarea` block**

The new `__select` block in the SCSS is an almost exact copy of the `__input`/`__textarea` block (same padding, border-radius, border, background, color, font-size, font-family, outline, box-sizing, transition, `:focus` rule). The only difference is `cursor: pointer` and `appearance: auto`. Extending the existing selector group (`&__input, &__textarea, &__select`) and adding the two distinct rules as overrides would eliminate the duplication and keep the file maintainable.

libs/agentos-api-client/src/custom/ai-model-alias.service.ts
```diff
@@ -0,0 +1,37 @@
+import { inject, Injectable } from '@angular/core'
+import { HttpClient } from '@angular/common/http'
+import { Observable } from 'rxjs'
+import { Configuration } from '../lib/configuration'
+
+/**
+ * AiModelAliasService — hand-written client for the AI model aliases endpoint.
+ *
+ * The endpoint `GET /api/ai-models/aliases-by-namespaceId/{namespaceId}` is not yet
+ * captured in the generated OpenAPI client. This service lives in `src/custom/` so
```
vincent-audibert-whoz: **[BLOCKER]** The URL here is `/api/ai-models/aliases-by-namespaceId/{namespaceId}` (flat hyphenated form). The backend controller maps the endpoint to `/api/ai-models/aliases/by-namespaceId/{namespaceId}` (with `aliases` as a distinct sub-path segment). These are different URLs — this call will 404 against the deployed backend. One side must be corrected before merge.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/aiModel/AiModelController.kt
vincent-audibert-whoz, line 174: The `mapNotNull { it.alias }.distinct().sorted()` projection is controller logic. Per the AgentOS service-layer architecture, the controller should be a thin HTTP adapter — data projection and business logic belong in `AiModelService`. Additionally, `findByNamespaceId` loads full entity graphs only to discard all fields except `alias`; a dedicated repository query returning only alias strings would avoid materialising unnecessary data.

vincent-audibert-whoz (commented):
## Review — WZ-31782 backend + frontend

One hard blocker (URL mismatch that will produce a 404 in production), one architectural concern, and one minor note.

**[BLOCKER] URL mismatch between backend and frontend.** The backend controller registers `GET /api/ai-models/aliases/by-namespaceId/{namespaceId}` (with `/aliases/` as a sub-path segment). The frontend `AiModelAliasService` calls `/api/ai-models/aliases-by-namespaceId/{namespaceId}` (flat, hyphenated, no sub-path). These are two different URLs. Every call from the frontend will produce a 404 against the actual backend. The two sides must agree on a single URL before merge.

**[Architectural concern] Projection logic in the controller.** The `mapNotNull { it.alias }.distinct().sorted()` chain lives directly in the controller method body. In AgentOS (production-grade, per architecture docs) the controller is a thin HTTP adapter; data projection belongs in the service layer. Beyond the layer violation, this approach loads full `AiModel` entity graphs into memory before discarding everything except the alias string — a dedicated Cypher projection (`MATCH (m:AiModel {namespaceId:$ns}) WHERE m.alias IS NOT NULL RETURN DISTINCT m.alias ORDER BY m.alias`) would be far cheaper at scale. The projection should move to `AiModelService` (and ideally to a repository query).

**[Minor] Test package inconsistency.** Both test files sit in `io.whozoss.agentos.aiModelConfig` while the production code lives in `io.whozoss.agentos.aiModel`. The new test cases inherit this pre-existing mismatch; flagging it here for eventual cleanup.

vincent-audibert-whoz:
## Review fixes applied (commit 9bab654)

All four issues from the review have been addressed:

### 🔴 URL mismatch — fixed
`AiModelAliasService` was calling `/api/ai-models/aliases-by-namespaceId/{namespaceId}`. Corrected to `/api/ai-models/aliases/by-namespaceId/{namespaceId}` to match the actual backend route in `AiModelController.kt`.

### 🟡 `[value]` vs `[ngValue]` — fixed
All `<option>` elements in the LLM model `<select>` now use `[ngValue]` consistently (both the `null` default and the alias options). This ensures Angular's reactive forms correctly match the control value for all options.

### 🟡 `||` → `??` regression — reverted
`description` and `instructions` fields in the submit payload are back to `|| undefined` (empty string coerced to `undefined`), matching the original behaviour. Only `modelName` uses `?? undefined` — intentionally, since `null` is the meaningful "no alias selected" state for that field.

### ⚪ SCSS duplication — refactored
`__select` is no longer a full copy of `__input`/`__textarea`. It is now part of the grouped `&__input, &__select, &__textarea { ... }` selector, with only the two differing rules (`cursor: pointer`, `appearance: auto`) in a dedicated `&__select { }` override block.

---

The Kotlin style note (fully-qualified Swagger annotations) is backend-side and out of scope for this frontend PR.

vincent-audibert-whoz:
## API client regenerated + hand-written service removed (commit 5b156de)

The OpenAPI spec at `agentos/openapi/agentos-openapi.yaml` already contained the new endpoint (added by the backend PR). Regenerating the client produced `listAliasesByNamespaceIdAiModel()` directly on the existing `AiModelControllerService`.

**Changes in this commit:**

- **Removed** `libs/agentos-api-client/src/custom/ai-model-alias.service.ts` — now redundant
- **Cleaned up** `libs/agentos-api-client/src/index.ts` — removed the export of the deleted service
- **Updated** `agent-config-form.component.ts` — now injects `AiModelControllerService` and calls `.listAliasesByNamespaceIdAiModel()` instead of the hand-written wrapper
- **Regenerated** `libs/agentos-api-client/src/lib/` — also picks up other backend additions present in the spec (`UserGroup` entity, `Namespace` model updates)

All checks pass: `pnpm nx lint agentos-api-client` ✅ · `pnpm nx lint agentos-ui` ✅ · `pnpm nx build client --configuration=development` ✅

PR #854, opened 2026-05-07
feat: #31783 call default agent - namespace defaultAgentName + routing fallback

vincent-audibert-whoz:
### What

Implements automatic default agent routing in AgentOS: when a user sends a message without an `@mention`, the namespace's configured default agent is called automatically. When the last active agent becomes unavailable, the runtime falls back to the default rather than crashing.

### Changes

**`Namespace` model** — new `defaultAgentName: String?` field, propagated through `NamespaceNode`, `NamespaceResource`, `NamespaceListItem`, and `NamespaceController`. Nullable by design: a namespace without a default configured is a valid explicit state that produces a clear error message rather than a silent failure.

**`CaseServiceImpl.selectAgent`** — rewritten as a `when` expression covering three cases:
1. Explicit `@mention` → resolve by name; on miss, `WarnEvent` + fallback to default
2. Sticky agent (last selected) → verify still available via `resolveAgentName`; if gone, `WarnEvent` + fallback to default
3. No mention, no history → delegate to `selectDefaultAgent`

**`selectDefaultAgent`** (new private method) — reads `namespace.defaultAgentName`, resolves it, returns `AgentSelectedEvent` on success or a `WarnEvent` + clean IDLE stop on null/unresolvable. No ERROR status: a missing default is a configuration issue, not a runtime crash.

**Cleanup** — removed `AgentConfigService.findDefault()`, `AgentConfigServiceImpl.DEFAULT_AGENT_CONFIG`, `AgentService.getDefaultAgent()`, `AgentService.getDefaultAgentName()`, `AgentConfig.hasNoRealNamespace()`, `AgentConfig.NO_NAMESPACE_ID`. The "default agent" concept now lives entirely on the namespace, not scattered across multiple service layers.

**OpenAPI** — regenerated `agentos-openapi.yaml` with `defaultAgentName` on both `Namespace` and `NamespaceListItem` schemas.

### Acceptance criteria coverage

| Criterion | Status |
|---|---|
| First message with no `@mention` → default agent called | ✅ |
| First message with `@mention` → named agent called (unchanged) | ✅ |
| Last active agent still available → routed to it | ✅ |
| Last active agent unavailable → silent fallback to default | ✅ |
| Clear error if no default configured in namespace | ✅ `WarnEvent` + IDLE |

### Design note

`defaultAgentName` is a soft reference by name, not a FK to an `AgentConfig` UUID. This is intentional: namespace admins may not control which `AgentConfig` versions are deployed on their namespace (platform-level concern). The name indirection decouples namespace configuration from `AgentConfig` versioning.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
@@ -4,6 +4,7 @@ import io.whozoss.agentos.agent.AgentExecutionContext
 import io.whozoss.agentos.agent.AgentService
 import io.whozoss.agentos.caseEvent.CaseEventService
 import io.whozoss.agentos.exception.ResourceNotFoundException
+import io.whozoss.agentos.namespace.NamespaceService
 import io.whozoss.agentos.sdk.actor.Actor
 import io.whozoss.agentos.sdk.caseEvent.AgentSelectedEvent
 import io.whozoss.agentos.sdk.caseEvent.CaseEvent
@@ -33,6 +34,7 @@ class CaseServiceImpl(
     private val caseRepository: CaseRepository,
     private val caseEventService: CaseEventService,
     private val userService: UserService,
+    private val namespaceService: NamespaceService,
 ) : CaseService {
     /**
      * Coroutine scope used to run case execution loops in the background.
@@ -153,16 +155,18 @@ class CaseServiceImpl(
      * events to store+emit on the runtime.
      *
      * Resolution order:
-     * 1. Explicit `@mention` in the message content.
-     * 2. Last agent selected in this case (from [pastEvents]) — preserves continuity
-     *    across turns when the user does not re-mention an agent.
-     * 3. Namespace default agent — fallback when no prior selection exists.
+     * 1. Explicit `@mention` in the message content — resolved by name; on miss, falls
+     *    back to the namespace default with a [WarnEvent].
+     * 2. Last selected agent in this case (from [pastEvents]) — preserves continuity
+     *    across turns. If the agent is no longer available (deleted, disabled), falls
+     *    back to the namespace default with a [WarnEvent].
+     * 3. Namespace default agent — [Namespace.defaultAgentName] resolved by name.
      *
-     * Returns an empty list when no agent is configured (signals the runtime to stop).
+     * Returns a single [WarnEvent] (no [AgentSelectedEvent]) when no default agent
+     * is configured on the namespace, signalling the runtime to stop cleanly.
      *
      * @param content the message content to inspect for @mention syntax.
-     * @param pastEvents the full event history of the case at the time of this call,
-     *   used to recover the last [AgentSelectedEvent] for sticky-agent behaviour.
+     * @param pastEvents the full event history of the case at the time of this call.
      */
     private fun selectAgent(
         content: List<MessageContent>,
@@ -184,13 +188,10 @@ class CaseServiceImpl(
             if (resolvedName != null) {
                 logger.info { "[CaseService] Agent mention resolved: @$mentionedName -> $resolvedName" }
                 return listOf(agentSelectedEvent(resolvedName, namespaceId, caseId))
-            } else {
-                logger.warn { "[CaseService] Agent '@$mentionedName' not found, falling back to default" }
-                val warn =
-                    WarnEvent(namespaceId = namespaceId, caseId = caseId, message = "Agent '$mentionedName' not found")
-                val defaultName = agentService.getDefaultAgentName(namespaceId)
-                return listOf(warn, agentSelectedEvent(defaultName, namespaceId, caseId))
             }
+            logger.warn { "[CaseService] Agent '@$mentionedName' not found, falling back to default" }
+            val warn = WarnEvent(namespaceId = namespaceId, caseId = caseId, message = "Agent '$mentionedName' not found")
+            return listOf(warn) + selectDefaultAgent(namespaceId, caseId)
         }
 
         // No explicit mention: re-use the last selected agent so the conversation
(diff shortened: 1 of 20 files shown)


libs/agentos-api-client/src/lib/api/user-group-controller.service.ts
```diff
@@ -0,0 +1,349 @@
+/**
```
vincent-audibert-whoz, line 1: This file, along with `user-group-create-request.ts`, `user-group-search-result.ts`, and `user-group-update-request.ts`, is an auto-generated Angular API client for a `UserGroup` resource that is unrelated to this PR's scope (namespace `defaultAgentName` + default-agent routing). Scope violation: unrelated generated code with no feature driving it in this PR, no tests consuming it, and no corresponding backend changes included here. These four files should be removed from this branch or moved to their own PR.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
```diff
@@ -512,6 +532,108 @@ class CaseServiceImplSpec :
...
+            val awaiter = scope.expectCaseStatus(runtime, CaseStatus.IDLE, CaseStatus.ERROR)
+            service.addMessage(
+                caseId = case.id,
+                actor = userActor,
+                content = listOf(MessageContent.Text("hello")),
+            )
+            awaiter.join()
+
+            // Case must be IDLE (not ERROR): no default is a clean stop, not a crash
+            service.getById(case.id).status shouldBe CaseStatus.IDLE
```
vincent-audibert-whoz, line 588: The assertion block is dead code: a new `caseEventService` is instantiated locally but never injected into `service`; the `events` variable is assigned `Unit` through a no-op lambda chain; and the only real assertion (`status == IDLE`) duplicates what the preceding test already verifies. The comment at the end acknowledges that the key assertion — that a `WarnEvent` was persisted — is missing. The `WarnEvent` emission path is therefore untested. The test should read events from the `caseEventService` that is actually wired into the service under test, then assert `events.filterIsInstance<WarnEvent>().shouldHaveSize(1)` (or equivalent).

vincent-audibert-whoz (commented):
The core backend logic is clean and the architectural direction is right: moving the default-agent concept from a scattered implicit fallback into an explicit nullable field on `Namespace` is a clear improvement. The `when` refactor in `selectAgent`, the `WarnEvent`-then-IDLE stop path, and the test coverage for the new routing branches are all well-executed.

Two issues need addressing before merge.

**Unrelated generated files in the diff.** Four new TypeScript files are included — `user-group-controller.service.ts`, `user-group-create-request.ts`, `user-group-search-result.ts`, `user-group-update-request.ts` — that have nothing to do with the default-agent feature. These are auto-generated `agentos-api-client` files for a `UserGroup` API that doesn't exist in this PR. Including unrelated generated code in a feature PR is a scope violation: it makes the diff harder to review and risks introducing dead/untested client code. They should be removed from this branch or moved to a dedicated PR.

**Incomplete test assertion.** In `CaseServiceImplSpec`, the test `"first message without @mention produces WarnEvent and stops when namespace has no default agent"` creates a local `caseEventService` that is never wired into the service under test, then assigns `events` to `Unit` through a dead-code block, and ends with only a comment where an assertion should be. The test asserts `status == IDLE` (which the preceding test already covers) but never verifies that a `WarnEvent` was actually persisted — the stated goal of the test. This leaves the `WarnEvent` emission path effectively untested and gives false confidence in the coverage.

Two additional notes, non-blocking:

- `selectDefaultAgent` calls `namespaceService.findById(namespaceId)` on every agent-selection invocation, including sticky-agent re-use turns. The namespace (and `defaultAgentName`) doesn't change during a case's lifetime; caching it on `CaseRuntime` construction would avoid a per-message service call.
- The `lastSelectedName != null` branch calls `resolveAgentName` to check availability on every turn, even the happy path where the agent is still present. This is an extra resolution call per message in the most common multi-turn scenario.

libs/agentos-api-client/src/lib/api/user-group-controller.service.ts
vincent-audibert-whoz, line 1: Auto-generated Angular API client for `UserGroup`. Unrelated to this PR's scope (namespace `defaultAgentName` + default-agent routing): no corresponding backend changes, no feature consuming it, no tests. Scope violation — remove from this branch or move to its own PR. Same applies to `user-group-create-request.ts`, `user-group-search-result.ts`, and `user-group-update-request.ts`.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
```diff
@@ -512,6 +532,118 @@ class CaseServiceImplSpec :
...
+            val awaiter = scope.expectCaseStatus(runtime, CaseStatus.IDLE, CaseStatus.ERROR)
+            service.addMessage(
+                caseId = case.id,
+                actor = userActor,
+                content = listOf(MessageContent.Text("hello")),
+            )
+            awaiter.join()
+
+            // Case must be IDLE (not ERROR): no default is a configuration issue, not a crash
+            service.getById(case.id).status shouldBe CaseStatus.IDLE
```
vincent-audibert-whoz, line 588: The warn-path for the unavailable-sticky-agent fallback is untested. The test verifies the final `IDLE` status but never asserts that a `WarnEvent` was persisted during the second turn. Add a check analogous to the null-default test: retain a reference to the `caseEventService`, read `findByParent(case.id)` after the second turn, and assert `filterIsInstance<WarnEvent>().shouldHaveAtLeastSize(1)`.

Also verify the MockK consecutive-answer setup: the second `every { resolveAgentName(unavailableAgentName, namespaceId) } returns unavailableAgentName andThen null` block re-declares the same stub — confirm MockK sequences these as intended rather than the second block silently replacing the first.

vincent-audibert-whoz (commented):
Two issues should be resolved before merge. The author already self-flagged both; this review confirms and adds detail.

**Unrelated generated files** — four files in `libs/agentos-api-client/` (`user-group-controller.service.ts`, `user-group-create-request.ts`, `user-group-search-result.ts`, `user-group-update-request.ts`) are auto-generated Angular client code for a `UserGroup` resource that has no corresponding backend changes in this PR, no feature consuming them, and no tests exercising them. Scope violation — remove from this branch or move to their own PR.

**Incomplete test coverage on the warn-path** — the `"last active agent unavailable falls back to namespace default"` test verifies the final `IDLE` status but never asserts that a `WarnEvent` was persisted during the second turn. The warn-path is therefore not exercised by the assertion. Additionally, the MockK consecutive-answer chain (`returns unavailableAgentName andThen null`) is set up on a re-declared `every` block; verify that the second `every { resolveAgentName(unavailableAgentName, namespaceId) }` sequences correctly in MockK rather than silently replacing the first stub.

**Minor / design note** — `selectDefaultAgent` calls `namespaceService.findById(namespaceId)` on every fallback invocation inside the hot coroutine path. If `NamespaceService.findById` ever involves a non-trivial lookup this adds per-turn latency. Worth considering resolving `defaultAgentName` once at `CaseRuntime` construction time and threading it through the `selectAgent` callback, rather than fetching it on each invocation.

vincent-audibert-whoz (commented):
Both concerns from the two test-related review comments are addressed in the current HEAD (2aa1b03):

**WarnEvent assertion (null-default test):** The `buildService(defaultAgentName = null)` shorthand was replaced with a manual wire-up that retains a `caseEventService` reference. The test now asserts `persistedEvents.filterIsInstance<WarnEvent>() shouldHaveAtLeastSize 1` and also confirms `filterIsInstance<AgentSelectedEvent>() shouldBe emptyList()` — the routing stopped at the warn.

**WarnEvent assertion (unavailable-sticky-agent test):** The `"last active agent unavailable falls back to namespace default"` test is now wired with an explicit `caseEventService`, and after `secondIdle.join()` it asserts both `filterIsInstance<WarnEvent>() shouldHaveAtLeastSize 1` and `filterIsInstance<AgentSelectedEvent>().last().agentName shouldBe agentName`.

**MockK consecutive-answer concern:** The double `every { resolveAgentName(unavailableAgentName, namespaceId) }` block was replaced with a single `answers { when (resolveCallCount.incrementAndGet()) { 1 -> unavailableAgentName; else -> null } }` using an `AtomicInteger`. This is unambiguous — there is no second `every` block to silently replace the first.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -153,49 +155,33 @@ class CaseServiceImpl(
...
         caseId: UUID,
     ): List<CaseEvent> {
-        val firstText =
+        val mentionedName =
             content
                 .filterIsInstance<MessageContent.Text>()
                 .firstOrNull()
                 ?.content
                 ?.trim()
+                ?.let { """^@(\S+)""".toRegex().find(it)?.groupValues?.get(1) }
```
vincent-couturier-whoz, line 183: suggestion: extract the regexp as a constant

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -204,14 +190,78 @@ class CaseServiceImpl(
...
+     * [AgentSelectedEvent] following it.
+     */
+    private fun selectDefaultAgent(
+        namespaceId: UUID,
+        caseId: UUID,
+    ): List<CaseEvent> {
+        val defaultAgentName = namespaceService.findById(namespaceId)?.defaultAgentName
+        if (defaultAgentName == null) {
+            logger.warn { "[CaseService] No default agent configured for namespace $namespaceId" }
+            return listOf(
```
vincent-couturier-whoz, line 244: suggestion: avoid early return as you can use return if () else a develop in a more functional way

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceListItem.kt
```diff
@@ -25,6 +25,8 @@ data class NamespaceListItem(
     val configPath: String? = null,
     @Schema(description = "Optional external identifier for this namespace, e.g. a federation id from an external system")
     val externalId: String? = null,
+    @Schema(description = "Logical name of the default agent for this namespace. Null means no default is configured.")
```
vincent-couturier-whoz, line 28: suggestion: perhaps we could precise the meaning of "logical name". For example is it case sensitive or not.

vincent-couturier-whoz (approved):

PR #860, opened 2026-05-11
feat(agentos): add retry mechanism on resolveIntentionAndTool (WZ-31597)

vincent-couturier-whoz:


vincent-audibert-whoz (changes requested):
Coday review: ~~Two~~ One issues to address before merging, both in `AgentAdvanced.kt`.

---

**1 — Implicit control flow via mutable state in the retry loop** (`resolveIntentionAndTool`, `repeat` block)

The loop catches all exceptions under a single `catch (e: Exception)` branch, then relies on whether `lastResponse` was set to silently distinguish between a parsing failure and a hard LLM exception. The fallback at the end of the function:

```kotlin
intention = lastResponse?.trim() ?: "Unable to generate intention"
```

branches on this incidental state rather than on an explicit decision. If the loop body is ever extended (e.g. `lastResponse` set before parsing), the two failure modes will silently conflate. Consider separating the two cases with distinct catch clauses (`IntentionParsingException` vs `Exception`) or by capturing the failure mode explicitly, so the fallback path is an intentional choice rather than a side-effect of execution order.

---

~~**2 — `MAX_INTENTION_ATTEMPTS = 2` is too thin for a production service** (`companion object`)~~

=> ok for intended usage.

---

**Minor — `IntentionParsingException` visibility** (`internal class IntentionParsingException`)

The exception is `internal` solely so that `AgentAdvancedSpec` can reference it with `shouldThrow<AgentAdvanced.IntentionParsingException>`. The exception is thrown and caught entirely within `AgentAdvanced` — it is not part of any module-level contract. The `internal` visibility leaks an implementation detail across the module boundary. Consider whether the tests should assert on observable outcomes (the fallback `IntentionGeneratedEvent` with `toolName == "Answer"`) rather than coupling to the internal exception type, which would allow the exception to be `private`.

vincent-audibert-whoz (approved):

PR #861, opened 2026-05-11
feat(agentos): add repetition detection mechanism on AgentAdvanced (WZ-31597)

vincent-couturier-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -89,7 +89,18 @@ class AgentAdvanced(
                     // iteration to complete.
                     if (!shouldContinue()) break
                     emit(ThinkingEvent(namespaceId = namespaceId, caseId = caseId))
-                    val intention = resolveIntentionAndTool(accumulatedEvents, namespaceId, caseId)
+
+                    // Detect repetitive tool usage
+                    val repeatedTool = detectRepetitionLoop(accumulatedEvents)
+                    val repetitionWarning = repeatedTool?.let { toolName ->
+                        val msg = "You have called `$toolName` $REPETITION_DETECTION_WINDOW times consecutively with no progress. " +
+                            "Stop and use the $ANSWER_TOOL tool to explain the situation or ask the user for more information."
```
vincent-audibert-whoz, line 99: The warning is emitted unconditionally every iteration once the loop is detected. `detectRepetitionLoop` looks at the last N events, so if the agent calls the same tool on iteration 4 as well, the check still returns the tool name and another `WarnEvent` is emitted. The LLM already received the warning in the previous prompt; re-emitting it on every subsequent iteration pollutes the event stream without adding signal. Consider tracking whether a warning was already emitted for the current streak (e.g. a flag or a comparison against the last emitted warn) and suppressing duplicates.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -176,6 +187,21 @@ class AgentAdvanced(
             }
         }
 
+    /**
+     * Detects when the last N tool calls all targeted the same tool,
+     * indicating a potential infinite loop.
+     *
+     * @return the repeated tool name if a loop is detected, null otherwise
+     */
+    internal fun detectRepetitionLoop(events: List<CaseEvent>): String? =
```
vincent-audibert-whoz, line 201: The detection filters on `ToolResponseEvent` without distinguishing successful from failed executions. A tool that fails N times consecutively (e.g. due to bad parameters or a transient error) will be flagged as a repetition loop even though the agent hasn't necessarily stalled — it may be retrying a legitimately failing call. The `success` field is available on `ToolResponseEvent`; consider restricting detection to successful responses, or at minimum documenting explicitly why failed responses are intentionally included.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentAdvancedSpec.kt
```diff
@@ -314,6 +318,137 @@ class AgentAdvancedSpec :
...
+                mockChatClient.prompt(any<Prompt>()).call().content()
+            } returnsMany
+                listOf(
+                    "<intention>Reading the file.</intention><toolName>FILES__ReadFile</toolName>", // iter 1 intention
+                    "{}", // iter 1 params
+                    "<intention>Reading the file again.</intention><toolName>FILES__ReadFile</toolName>", // iter 2 intention
+                    "{}", // iter 2 params
+                    "<intention>Reading the file yet again.</intention><toolName>FILES__ReadFile</toolName>", // iter 3 intention
+                    "{}", // iter 3 params
+                    "<intention>Stopping due to loop.</intention><toolName>Answer</toolName>", // iter 4 intention (after WarnEvent)
```
vincent-audibert-whoz, line 440: The assertion `warnEvents.any { it.message.contains("consecutively") || it.message.contains("FILES__ReadFile") }` is too permissive: it passes as long as any single warn event contains either keyword. It would not catch a bug where the wrong tool name is embedded, or where multiple spurious warn events are emitted. A tighter assertion — `warnEvents shouldHaveSize 1` and a direct `shouldContain` on both expected substrings — would better pin down the contract.

vincent-audibert-whoz (changes requested):
The repetition detection logic is sound and the test coverage is solid, but there are three issues worth addressing before merge.

The warning fires on every iteration once the threshold is crossed, not just once. After 3 consecutive same-tool calls the `WarnEvent` is emitted on iteration 4, but if the agent ignores the warning and keeps calling the same tool, iterations 5, 6, … will each emit another `WarnEvent` with identical text. This is both noisy on the event stream and redundant as a signal to the LLM — the warning is already in the prompt, repeating it adds no information.

The `detectRepetitionLoop` function has a semantic blind spot: it looks at the last N `ToolResponseEvent`s across the entire accumulated history, regardless of success. A tool that legitimately fails N times in a row (e.g. transient errors) will trigger the repetition warning even though the agent may be retrying correctly. The `success` field on `ToolResponseEvent` is available and should be considered.

The integration test `"emits WarnEvent when repetition loop is detected…"` asserts the warn condition with a loose `any { … }` predicate — it would pass even if the tool name in the message were wrong or the word "consecutively" were missing from one of multiple warn events. A tighter assertion (e.g. checking `warnEvents shouldHaveSize 1` and asserting the exact message content) would better pin down the contract and catch regressions.

vincent-audibert-whoz (approved):

PR #863, opened 2026-05-12
feat: #31424 Implement search users with user groups by external id

AlanMenantWhoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -70,7 +73,7 @@ class UserController(
     override fun toDomain(resource: UserResource): User =
         User(
             metadata = EntityMetadata(id = resource.id ?: UUID.randomUUID()),
-            externalId = "",  // server-managed — never sourced from the request body
+            externalId = "", // server-managed — never sourced from the request body
```
vincent-couturier-whoz, line 76: suggestion: very strange mapping. Not convinced by the comment. We should provide a classical mapping with the resource because externalId should be updatable by an external system

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -133,11 +143,12 @@ class UserController(
...
         val newIsAdmin = if (isSelfEdit) existing.isAdmin else resource.isAdmin
-        val updated = toDomain(resource).copy(
-            metadata = existing.metadata,
-            externalId = existing.externalId,        // server-managed (IdP key)
-            isAdmin = newIsAdmin,                    // self-rule applied above
-        )
+        val updated =
+            toDomain(resource).copy(
+                metadata = existing.metadata,
+                externalId = existing.externalId, // server-managed (IdP key)
```
vincent-couturier-whoz, line 149: suggestion: I know that it's not a change in this PR but we should let an external system change an externalId of a user

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
```diff
@@ -174,5 +187,18 @@ class UserController(
     @Operation(summary = "Get the current user's profile")
     fun getMe(): UserResource = toResource(userService.getCurrentUser())
 
+    /** POST /api/users/groups-by-external-ids — return groups per user. Super-admin only. */
+    @PostMapping("/groups-by-external-ids")
+    @ResponseStatus(HttpStatus.OK)
+    @PreAuthorize("hasRole('SUPER_ADMIN')")
```
vincent-couturier-whoz, line 193: suggestion: perhaps we could have /api/admin and protect using SecurityManager configuration the /api/admin to avoid each endpoint to precise a PreAuthorize. Even if we don't put in place the Security Manager we could have the common api prefix to centralize critical API

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
```diff
@@ -100,6 +106,34 @@ open class Neo4jUserGroupRepository(
         neo4jRepository.removeUsers(userGroupId.toString(), userExternalIds.toList())
     }
 
+    override fun findGroupsByUserExternalIds(externalIds: Collection<String>): Map<String, List<UserGroupSummary>> {
+        if (externalIds.isEmpty()) return emptyMap()
+        return neo4jClient
+            .query(
+                """
+                    UNWIND $externalIds AS extId
```
vincent-couturier-whoz: suggestion: use the in operator instead of UNWIND

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserController.kt
vincent-audibert-whoz: Agreed, that was a design error.

vincent-audibert-whoz (approved):

PR #876, opened 2026-05-18
feat: wz-31609 add support of vllm with headers

yannick-bride-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/chat/ChatModelFactory.kt
```diff
@@ -42,6 +44,17 @@ class ChatModelFactory {
                 )
             }
 
+            AiApiType.vLLM -> {
+                createVllmModel(
+                    baseUrl = baseUrl ?: OPENAI_DEFAULT_BASE_URL,
```
vincent-audibert-whoz: Weird default, would not be better without and force baseUrl not null ?

vincent-audibert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/chat/ChatModelFactory.kt
vincent-couturier-whoz: suggestion: Perhaps we could use "localhost" and default vllm port as the default value is we really need a default value

agentos/aiprovider/vllm.yaml.example
```diff
@@ -3,4 +3,7 @@ description: provider of infrastructure is ovhai, api provider is vllm with api
 apiType: OpenAI
 apiKey: "****"
```
vincent-couturier-whoz, line 4: we could just indicate - here to indicate that apiKey is not really useful anymore

vincent-couturier-whoz (dismissed):

vincent-couturier-whoz (approved):

PR #878, opened 2026-05-18
feat: #WZ-31875: Deploy agent into namespace

AlanMenantWhoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespaceController.kt
```diff
@@ -227,6 +227,38 @@ class NamespaceController(
...
     }
 
+    /**
+     * POST /api/namespaces/{namespaceId}/deploy-agents
+     *
+     * Deploys the given agents on the namespace by creating a DEPLOYED_TO relation.
+     */
+    @PostMapping("/{namespaceId}/deploy-agents", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    @PreAuthorize("hasRole('SUPER_ADMIN')")
```
vincent-couturier-whoz: suggestion: remove the PreAuthorize and use the same rights as the one on agent update

vincent-couturier-whoz (dismissed):
issue(non blocking for now): we have to add an information about the deployed status of the agent on the agent resource but it could be achieved in another PR. We should also clarify the meaning of deployed agent in a namespace vs in a user group and identify the way we want to achieve draft vs published agent (the existing status in whoz copilot)

mickael-barroux-whoz (dismissed):

mickael-barroux-whoz (approved):

PR #890, opened 2026-05-20
feat: wz-31608: add endpoints for role provisioning

thomasmartin-whoz:


thomas-martin-whoz:
Review of PR #890 (updated patch)
The scope is now clean: all changes are in agentos/ (production-grade zone), covering authentication simplification, two new namespace permission endpoints, and a new user lookup endpoint. No sensitive files, no dependency changes, no version downgrade.

Issues Found
NamespacePermissionEndpoints.kt — listNamespaceRolesForUser self-service authorization is permanently dead

@PreAuthorize("hasRole('SUPER_ADMIN') or #userId.toString() == authentication.name")
fun listNamespaceRolesForUser(@PathVariable userId: UUID): List<NamespaceRole>
authentication.name is set by AgentOsAuthenticationFilter to the value of X-External-User-Id, which is an opaque external identity key (e.g. alice@example.com). #userId is an internal UUID. These two values can never be equal — the self-service branch #userId.toString() == authentication.name is permanently false. The docstring says "Authorization: SUPER_ADMIN only", which contradicts the expression. Either the self-service clause is dead code that should be removed, or the comparison must be against the user's externalId resolved from the path variable. As written, the expression silently degrades to SUPER_ADMIN-only while misleading future readers.

NamespacePermissionEndpoints.kt — updateRolesByExternalId uses full-list membership checks instead of point lookups

val hasAdmin = permissionService
    .listUsersWithPermission(EntityType.NAMESPACE, namespaceIdStr, PermissionRelation.ADMIN)
    .contains(userIdStr)
val hasMember = permissionService
    .listUsersWithPermission(EntityType.NAMESPACE, namespaceIdStr, PermissionRelation.MEMBER)
    .contains(userIdStr)
For each assignment, two calls each return the full user list for the namespace relation, then scan it for a single user. For namespaces with many members this loads unbounded data into memory for a containment check. This pattern is repeated per assignment in the loop, producing O(assignments × namespace-size) graph reads. If permissionService exposes a direct hasPermission(userId, entity, relation) method, it must be used here. If it does not, the absence of such a method is an architectural gap that this PR is the right moment to surface.

AuthSecurityService.kt — unconditional header trust requires deployment pre-condition not enforced in code

val externalUserId = request.getHeader(X_EXTERNAL_USER_ID_HEADER)
// ...
return externalUserId  // trusted unconditionally
The previous implementation verified a Cloudflare-signed JWT. The new implementation trusts X-External-User-Id with no signature or shared-secret check, relying entirely on network-level controls to prevent spoofing. This is an acceptable architectural choice only if the deployment topology guarantees AgentOS is unreachable except through the authenticating gateway. This pre-condition is stated in the docstring but not enforced anywhere in the codebase. If the service is accidentally exposed directly (misconfigured ingress, local mode: auth testing against a real graph), full identity impersonation is trivially achievable with a crafted header. The risk should be acknowledged: either add an integration test that asserts mode: auth rejects requests missing the header at the HTTP layer (already tested), or add a startup assertion that warns loudly when mode=auth and no additional network binding is configured. At minimum, the shouldNotFilter bypass for /management/ should be verified to not be reachable from outside the cluster, since that path skips identity resolution entirely.

NamespaceController.kt — listByExternalIds logs the full list of external IDs at INFO level

logger.info { "listByExternalIds called with $externalIds external id(s)" }
$externalIds interpolates the full List<String> into the log line. In production this emits every external identity key in the request body to the log on every call, which is a data exposure risk (external IDs may be emails or other PII) and a log-volume concern for large lists. The log should record the count, not the content.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/namespace/NamespacePermissionServiceImpl.kt
```diff
@@ -0,0 +1,114 @@
...
+}
+
+@Service
+class NamespacePermissionServiceImpl(
+    private val namespaceService: NamespaceService,
+    private val userService: UserService,
+    private val permissionService: PermissionService,
+) : NamespacePermissionService {
+
+    override fun syncUserRoles(request: SyncUserRolesRequest) {
```
vincent-couturier-whoz, line 42: issue: missed `@Transaction`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/AuthSecurityService.kt
```diff
@@ -1,152 +1,44 @@
...
- * Throws 401 when no identity can be resolved.
+ * Throws 401 when the header is absent or blank.
  *
  * User persistence (lookup / auto-create) is handled upstream by
  * [io.whozoss.agentos.user.UserService.resolveOrCreateByExternalId].
  */
-class AuthSecurityService(
-    private val objectMapper: ObjectMapper,
-) : SecurityService {
+class AuthSecurityService : SecurityService {
```
vincent-couturier-whoz: issue: you cannot remove the existing CF_AUTHORIZATION_HEADER translation because it's needed by coday when deployed behind CF

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/SecurityConfiguration.kt
```diff
@@ -10,7 +9,7 @@ import org.springframework.context.annotation.Configuration
  * Registers the correct [SecurityService] bean based on [SecurityConfigProperties.mode].
  *
  * - `local` (default): [LocalSecurityService] — OS username, no external dependencies.
- * - `auth`: [AuthSecurityService] — Cloudflare JWT / x-forwarded-email.
+ * - `auth`: [AuthSecurityService] — trusts the `X-External-User-Id` header set by the upstream gateway.
```
vincent-couturier-whoz: issue(blocking): we need to keep mecanism needed by coday

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/security/AuthSecurityServiceUnitSpec.kt
```diff
@@ -1,166 +1,58 @@
...
 import org.springframework.mock.web.MockHttpServletRequest
 import org.springframework.web.context.request.RequestContextHolder
 import org.springframework.web.context.request.ServletRequestAttributes
 import org.springframework.web.server.ResponseStatusException
-import java.util.Base64
 
 class AuthSecurityServiceUnitSpec : StringSpec({
     timeout = 5000
 
-    val objectMapper = ObjectMapper().registerKotlinModule()
```
vincent-couturier-whoz, line 17: issue: test on CF header and email should be restored

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/AuthSecurityService.kt
```diff
@@ -13,69 +14,73 @@ import java.util.Base64
...
         emailHeader: String?,
     ): String? {
-        val emailFromToken =
-            when {
-                !cfHeader.isNullOrBlank() -> {
-                    extractEmailFromJwt(cfHeader)
-                }
+        if (!externalUserIdHeader.isNullOrBlank()) {
+            logger.info { "[Security/auth] Resolved identity from $X_EXTERNAL_USER_ID_HEADER: $externalUserIdHeader" }
+            return externalUserIdHeader
```
vincent-couturier-whoz: suggestion(non blocking): avoid early return

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/security/AuthSecurityService.kt
```diff
@@ -13,69 +14,73 @@ import java.util.Base64
...
+            logger.warn { "[Security/auth] No identity header found — returning 401" }
+            throw ResponseStatusException(
+                HttpStatus.UNAUTHORIZED,
+                "No identity header found. Expected '$X_EXTERNAL_USER_ID_HEADER', " +
+                    "'$CF_AUTHORIZATION_HEADER' or '$X_FORWARDED_EMAIL_HEADER'.",
+            )
+        }
     }
 
     private fun resolveEmail(
```
vincent-couturier-whoz: suggestion: change the name to be more coherent with the real responsibility

vincent-couturier-whoz (dismissed):

vincent-couturier-whoz (approved):

PR #894, opened 2026-05-21
fix: #WZ-32044 pass ToolContext with namespaceId to provideTools in ToolResolverService

vincent-audibert-whoz:
## Root cause

`ToolResolverService` called `plugin.provideTools(config.parameters, config.name)` in both `resolveToolsForNamespace` and `resolveToolsForRun` without passing a `ToolContext`. The third parameter defaulted to `null`.

`RedirectToolPlugin.provideTools` starts by checking `context?.namespaceId` — when `null`, it immediately returns `emptyList()` and logs a warning. This meant the redirect tool was never instantiated regardless of the configured patterns (`\"*\"` or explicit names).

## Fix

Pass a `ToolContext` at both callsites:
- `resolveToolsForNamespace`: `ToolContext(namespaceId, userId=null, userExternalId=null, caseEvents=emptyList())`
- `resolveToolsForRun`: `ToolContext(namespaceId, userId=userId, userExternalId=null, caseEvents=emptyList())`

`userExternalId` and `caseEvents` are not available at tool-resolution time (they belong to the per-invocation runtime context), so they are passed as `null`/`emptyList()`. `RedirectToolPlugin` only needs `namespaceId` to resolve eligible agents.

## Tests

Two new unit tests added to `ToolResolverServiceSpec`:
- `resolveToolsForNamespace passes namespaceId via ToolContext to plugins that need it`
- `resolveToolsForRun passes namespaceId via ToolContext to plugins that need it`

Both used a context-capturing plugin that returns `emptyList()` when `context?.namespaceId == null`, mirroring the exact guard in `RedirectToolPlugin`. Both failed before the fix, both pass after."

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
@@ -5,6 +5,7 @@ import io.whozoss.agentos.integrationConfig.IntegrationConfigService
 import io.whozoss.agentos.exception.ConfigNotFoundException
 import io.whozoss.agentos.reconciliation.ConfigMergeService
 import io.whozoss.agentos.sdk.tool.StandardTool
+import io.whozoss.agentos.sdk.tool.ToolContext
 import mu.KLogging
 import org.springframework.stereotype.Service
 import java.util.UUID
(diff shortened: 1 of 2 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -37,8 +38,11 @@ class ToolResolverService(
             }
             try {
                 val allowedToolNames = agentIntegrations?.get(config.name)
-                val providedTools = plugin.provideTools(config.parameters, config.name)
-                    .filter { tool -> isToolAllowed(tool.name, config.name, allowedToolNames) }
+                val providedTools = plugin.provideTools(
```
thomasmartin-whoz: named params

thomasmartin-whoz (dismissed):

thomasmartin-whoz (approved):

PR #908, opened 2026-05-25
fix(agentos): fix redirect tools in advanced mode (WZ-32112)

vincent-couturier-whoz:
Emit ToolResponseEvent before re-throwing AgentInterrupt to maintain well-formed event history improve orphan tool request event conversion to avoid error 400 on openai

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
@@ -176,9 +176,36 @@ class AgentAdvanced(
         accumulatedEvents.add(parameters)
 
         if (!shouldContinue()) return
-        val response = executeTool(parameters, namespaceId, caseId)
+        // executeTool always returns a ToolResponseEvent, even on AgentInterrupt.
+        // We must emit and accumulate the response before re-throwing so the event
+        // history stays well-formed (every ToolRequestEvent has a matching response).
+        var interrupt: AgentInterrupt? = null
+        val response =
+            try {
+                executeTool(parameters, namespaceId, caseId)
+            } catch (e: AgentInterrupt) {
+                interrupt = e
+                // executeTool already built the ToolResponseEvent before throwing;
+                // the .also { throw e } pattern means the value was created but never
+                // returned. Re-create the response here so it can be emitted.
+                ToolResponseEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    toolRequestId = parameters.toolRequestId,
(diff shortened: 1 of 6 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -176,9 +176,36 @@ class AgentAdvanced(
...
+                // the .also { throw e } pattern means the value was created but never
+                // returned. Re-create the response here so it can be emitted.
+                ToolResponseEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    toolRequestId = parameters.toolRequestId,
+                    toolName = parameters.toolName,
+                    output =
+                        MessageContent.Text(
+                            when (e) {
```
thomasmartin-whoz, line 198: bizarre de faire un when ici

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
```diff
@@ -332,22 +332,23 @@ class AgentSimple(
...
-                        val output =
+                    val output =
+                        response?.let {
                             when (val content = it.output) {
                                 is MessageContent.Text -> content.content
                                 else -> content.toString()
                             }
-                        ToolResponseMessage.ToolResponse(toolCall.id(), toolCall.name(), output)
-                    }
+                        } ?: "[No response recorded]"
```
thomasmartin-whoz, line 347: const

thomasmartin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
vincent-couturier-whoz: It follows the same pattern as the other AgentInterrupt type exceptions. This way, if another exception of the same type is added, an error will be raised here, indicating the expected behavior. It's not the prettiest approach, but it's been seen used this way in several places. The design should be revisited, but that goes beyond the scope of this PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
vincent-couturier-whoz: wright but I will change this in another PR to be able to unblock situation and to do something better with a real json response. Here it will only be compliant with openai and vllm but it will fail with gemini and potentially unstable globally.

PR #910, opened 2026-05-26
feat: #WZ-31597 ToolExecutionResult with metadata propagated to ToolResponseEvent

vincent-audibert-whoz:
## Context

WZ-31597 — tools need to return structured metadata alongside their text output so that subsequent tool calls in the same integration can perform coherence checks (e.g. verifying a `getEntity` was performed before an `update`).

### SDK (`agentos-sdk`)

**`ToolExecutionResult`** restructured:
- Adds `metadata: Map<String, Any?>` — opaque per-call state at the tool's discretion
- Adds factory methods `success(output, metadata)` and `error(output, errorType, errorMessage)` for the common cases
- Removes `toolName` and `executionDurationMs` (redundant with call context; duration is now owned by the agent layer)

**`StandardTool`**: `execute()` and `executeWithJson()` now return `ToolExecutionResult` instead of `String`. The default `executeWithJson()` implementation wraps `execute()` transparently — plugins that don't need metadata only need to update their `execute()` return type.

**`ToolResponseEvent`**: adds `toolMetadata: Map<String, Any?> = emptyMap()` (persisted, readable by downstream tools via `ToolContext.caseEvents`) and `durationMs: Long?` which was already measured in the agents but never stored.

### Service (`agentos-service`)

- `ToolResponseEventNode` gains `metadataJson: String?` (nullable, backward-compatible with existing nodes) and `durationMs: Long?`
- `CaseEventNodeMapper` and `MessageContentSerializer` updated accordingly
- `AgentSimple` and `AgentAdvanced` propagate `ToolExecutionResult.metadata` and `durationMs` into the emitted `ToolResponseEvent`
- `RedirectTool` returns typed `ToolExecutionResult.error()` on invalid input instead of plain strings
- All affected tests updated


libs/agentos-api-client/src/lib/api/namespace-permission-endpoints.service.ts
```diff
@@ -409,4 +411,83 @@ export class NamespacePermissionEndpointsService extends BaseService {
...
+      } else if (this.configuration.isJsonMime(localVarHttpHeaderAcceptSelected)) {
+        responseType_ = 'json'
+      } else {
+        responseType_ = 'blob'
+      }
+    }
+
+    let localVarPath = `/api/namespaces/update-roles-by-external-id`
+    const { basePath, withCredentials } = this.configuration
+    return this.httpClient.request<SyncUserRolesRequest>('post', `${basePath}${localVarPath}`, {
```
thomasmartin-whoz, line 482: je vois pas le rapport avec la PR

thomasmartin-whoz (dismissed):

libs/agentos-api-client/src/lib/api/namespace-permission-endpoints.service.ts
vincent-audibert-whoz: Left over from a previous change, made it to the client generation. Changes in `agentos-api-client` are completely generated, so do not consider them in the PR.

thomasmartin-whoz (approved):

PR #911, opened 2026-05-26
fix: wz-32167: prevent admins from accessing cases of other users even if in namespace

thomasmartin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
vincent-couturier-whoz, line 1: suggestion: perhaps this PR solve an issue about security but we mixed filtering for purpose/feature vs filtering for security reason. The admin could access cases but it should have like anybody the list of it's own cases in order. If I go beyond that we should be able to filter the accessible cases by owner of the cases and at the end by participant knowing that owner and participant will have different access rights. So even if here we enhance the way we filter for security reason it seems to cover only security and not owner filtering which is the main issue we have in astra.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -26,24 +26,22 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
-     * on the parent namespace.
      */
     @Query(
         $$"""MATCH (c:Case)-[r:BELONGS_TO]->(ns:Namespace {id: $namespaceId})
             WHERE (c.removed IS NULL OR c.removed = false)
-              AND (
-                EXISTS { MATCH (:User {id: $userId})-[:ADMIN|MEMBER]->(c) }
-                OR EXISTS { MATCH (:User {id: $userId})-[:ADMIN]->(ns) }
-              )
+              AND EXISTS { MATCH (:User {id: $userId})-[:ADMIN|MEMBER]->(c) }
```
vincent-couturier-whoz, line 44: Issue: we won't be at the same level as copilot in terms of rights. This choice has an operational impact. The list of cases in Astra should show the cases of the users but not all cases it can access in the namespace.

vincent-couturier-whoz (approved):
Technically valid but not convinced by the way we address the issue encountered in Astra.

vincent-audibert-whoz (approved):
Nice simplification.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
vincent-audibert-whoz: That is not the new behavior of this query that removes the admin on namespace as a valid path to reach a case: only cases the user is member|admin are kept = the ones he owns or can explicitly view.

PR #917, opened 2026-05-28
feat: user group member list and role management API

vincent-audibert-whoz:
## What

Extends the UserGroup API to support proper internal administration.

### GET `/api/user-groups/{id}`
Now returns the full member list with roles instead of a bare count:
```json
{
  "members": [
    { "userId": "...", "externalId": "alice@example.com", "email": "alice@example.com", "role": "ADMIN" },
    { "userId": "...", "externalId": "bob@example.com",   "email": "bob@example.com",   "role": "MEMBER" }
  ],
  "userCount": 2
}
```
The list query (`GET ?namespaceId=`) keeps the count-only shape — loading thousands of members per group in a namespace listing would be unacceptable at scale.

### PUT `/api/user-groups/{id}/members` (new)
Batch upsert of memberships. Each entry carries `{ userId, role }`:
- `role = "ADMIN"` — promote/grant ADMIN (revokes any existing MEMBER edge first)
- `role = "MEMBER"` — demote/grant MEMBER (revokes any existing ADMIN edge first)
- `role = null` — remove the user from the group entirely

Idempotent. Unknown `userId`s are silently skipped. Authorization: `WRITE` on the group (namespace ADMIN via transitive permission).

### PUT `/api/user-groups/{id}` (was POST)
Group update verb aligned with the rest of the API (`AgentConfigController`, `UserController`, etc.).

## Design notes

The Whoz sync paths (`addUsers`/`removeUsers` by `externalId`) are unchanged. Both mechanisms converge on the same `[:MEMBER]`/`[:ADMIN]` Neo4j edges — origin doesn't matter, there is always at most one edge per (user, group) pair.

The `upsertMembership` Cypher query uses `OPTIONAL MATCH … DELETE` + `FOREACH … MERGE` to atomically revoke the opposite edge and set the desired one in a single round-trip per entry.

## Tests

- 18 persistence contract tests (`EmbeddedNeo4jUserGroupPersistenceSpec`) covering `findByIdWithDetails` member list, `updateMemberships` promote/demote/remove/idempotency/unknown-user
- 14 service unit tests (`UserGroupServiceImplUnitSpec`) including the two new `updateMemberships` cases
- All green

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
@@ -13,6 +13,13 @@ private data class UserExternalIdGroupRow(
     val groupName: String,
 )
 
+private data class UserGroupMemberRow(
+    val userId: String,
+    val externalId: String,
+    val email: String,
+    val role: String,
+)
+
 open class Neo4jUserGroupRepository(
     private val neo4jRepository: UserGroupNodeNeo4jRepository,
     private val childLinkService: Neo4jChildLinkService,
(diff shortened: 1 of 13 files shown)


vincent-audibert-whoz:
Too old, dropping in favor of smaller more focused PRs.

PR #933, opened 2026-06-01
feat: agent-config-inspect component with definition view and cross-links

vincent-audibert-whoz:
## What

New read-only `AgentConfigInspectComponent` that calls `getDefinitionAgentConfig()` and displays the resolved `AgentDefinition` in a structured layout.

## Why

The `getDefinitionAgentConfig` endpoint resolves the full agent definition (model, provider, tools with their JSON schemas) from a config. This view makes that resolution visible and inspectable without having to hit the API manually.

### New component — `agent-config-inspect`

Route: `/:namespaceId/agent-configs/:agentConfigId/inspect`

Three sections:
- **General** — name, resolved model API name, provider, advanced execution badge, optional user context
- **Instructions** — collapsible `<pre>` block (collapsed by default), only shown when present
- **Tools** — accordion list; each tool expands to show its JSON `inputSchema` parsed and pretty-printed (fallback to raw string if invalid JSON)

### Navigation cross-links

- **`agent-config-item`** — new "Inspect definition" entry in the kebab menu (between Edit and Delete)
- **`agent-config-form`** — "View resolved definition →" link in the header, visible in edit mode only
- **`agent-config-inspect`** — "Edit config →" link in the header, always visible


vincent-audibert-whoz:
Requesting a backend review from @Reviewer for the AgentOS/backend parts of this PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -46,7 +48,27 @@ class AgentServiceImpl(
...
+    override fun resolveDefinition(
+        agentConfigId: UUID,
+        namespaceId: UUID,
+        userId: UUID?,
+    ): ResolvedAgentDefinition {
+        val agentConfig =
+            agentConfigService.findById(agentConfigId)
+                ?: throw IllegalArgumentException(
+                    "No AgentConfig found for id '$agentConfigId' in namespace $namespaceId.",
+                )
```
vincent-audibert-whoz, line 64: **Bug: `caseId` is semantically wrong here.**

```kotlin
val context =
    AgentExecutionContext(
        namespaceId = namespaceId,
        caseId = agentConfig.metadata.id,  // <- wrong
        userId = userId,
    )
```

`AgentExecutionContext.caseId` represents the live case (conversation) being executed. Passing `agentConfig.metadata.id` here substitutes a config entity ID for a case ID — two conceptually different things. When `resolveDefinition` is called from the inspection endpoint there is no case at all; the context is constructed only to thread `namespaceId` and `userId` through `buildInstructions` and `resolveToolsForRun`.

Using the agentConfig ID as a fake caseId will silently produce a wrong value if anything downstream reads `context.caseId`. The `caseEventsProvider` defaults to `emptyList()` so it does not blow up today, but the field is misrepresented and fragile. A `UUID.randomUUID()` placeholder, a nil UUID constant, or making `caseId` nullable on `AgentExecutionContext` would be more honest. At minimum document with a comment if the current approach is intentional.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
```diff
@@ -141,5 +143,52 @@ class AgentConfigController(
...
+    @GetMapping("/{id}/definition")
+    @PreAuthorize("hasPermission(#id, 'AgentConfig', 'READ')")
+    @HideOnAccessDenied
+    fun getDefinition(
+        @PathVariable id: UUID,
+        @RequestParam(required = false) userId: String?,
+    ): AgentDefinitionResource {
+        val agentConfig = agentConfigService.findById(id)
+            ?: throw ResourceNotFoundException("AgentConfig not found: $id")
+        val resolvedUserId = userId?.takeIf { it == "me" }?.let { userService.getCurrentUser().metadata.id }
```
vincent-audibert-whoz: **Leaky `userId` parameter contract — any value other than `"me"` is silently ignored.**

```kotlin
val resolvedUserId = userId?.takeIf { it == "me" }?.let { userService.getCurrentUser().metadata.id }
```

If a caller passes `userId=some-arbitrary-string`, `takeIf` returns null and the request silently falls back to namespace-only resolution without any feedback. The caller cannot distinguish "my userId was ignored" from "no userId was provided". This is especially confusing since the OpenAPI spec documents `userId` as an unconstrained `string`.

Two options: validate that the only accepted value is `"me"` and return `400` for anything else, or replace the string parameter with a boolean flag (e.g. `withUserContext: Boolean = false`) to make the intent unambiguous in the API surface.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -46,7 +48,27 @@ class AgentServiceImpl(
                 ?: throw IllegalArgumentException(
                     "No AgentConfig found for name '$namePart' in namespace ${context.namespaceId}.",
                 )
-        return createAgentFromConfig(agentConfig, context)
+        val definition = resolveAgentDefinition(agentConfig, context)
+        return instantiateAgent(definition, context)
+    }
+
+    override fun resolveDefinition(
+        agentConfigId: UUID,
```
vincent-audibert-whoz, line 56: **Code-style violation: guard-clause / early-exit pattern.**

```kotlin
val agentConfig =
    agentConfigService.findById(agentConfigId)
        ?: throw IllegalArgumentException(...)
// ...
return resolveAgentDefinition(agentConfig, context)
```

The `?: throw` is an early-exit guard in the middle of the method body. `agentos/docs/code-style.md` explicitly forbids this: *"Avoid early returns in the middle of a method — structure the logic so the single return is at the end"*. The same pattern exists in `findAgentByName` (pre-existing), but this PR introduces a new instance of it. Consider expressing the null check and happy path as a single `when` expression or restructuring so there is one terminal `return`.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agentConfig/AgentConfigControllerUnitSpec.kt
```diff
@@ -34,9 +35,10 @@ import java.util.UUID
 class AgentConfigControllerUnitSpec : StringSpec({
```
vincent-audibert-whoz, line 35: **Missing test coverage for `getDefinition`.**

`agentService` is now declared and injected into the controller but `AgentConfigControllerUnitSpec` contains zero tests for the new `getDefinition` endpoint. The existing spec covers all other controller methods with both happy-path and error cases. The new endpoint has its own logic worth exercising at the AgentOS production-grade bar:

- happy path: `agentConfigService.findById` returns a config → `agentService.resolveDefinition` is called with the correct `namespaceId` → resource fields are mapped correctly
- 404 path: `agentConfigService.findById` returns null → `ResourceNotFoundException` is thrown
- `userId="me"` path: `userService.getCurrentUser()` is called and its id is forwarded to `resolveDefinition`
- `userId=null` and `userId=<other-value>` paths (especially important given the silent-fallback issue noted above)

agentos/openapi/agentos-openapi.yaml
```diff
@@ -1736,6 +1736,30 @@ paths:
...
+      parameters:
+      - name: id
+        in: path
+        required: true
+        schema:
+          type: string
+          format: uuid
+      - name: userId
+        in: query
+        required: false
```
vincent-audibert-whoz, line 1753: **`userId` query parameter has no enum constraint and no description.**

```yaml
- name: userId
  in: query
  required: false
  schema:
    type: string
```

The only valid non-null value is `"me"` but the schema accepts any string. The generated TypeScript client (`agent-config-controller.service.ts`) therefore types the parameter as `string` with no hint about valid values. Either add `enum: [me]` to the schema, or add a `description` documenting the semantics. This tracks the controller-level issue noted above.

vincent-audibert-whoz (commented):
The AgentOS backend changes are well-structured overall: the two-phase split of agent construction into `resolveAgentDefinition` / `instantiateAgent` is a clean refactor, and the new endpoint follows the existing controller patterns correctly. Three issues need attention before merge: one semantic bug in `resolveDefinition`, one missing test coverage for the new endpoint, and one code-style violation per the project guidelines. The `userId` API contract on the new endpoint is also worth discussing — it currently accepts any string value silently, which is a leaky abstraction.

thomasmartin-whoz (approved):

PR #937, opened 2026-06-01
feat: add case-level context to Case entity, merged with sessionContext in agent prompts

vincent-audibert-whoz:
## What

Adds a `context: Map<String, Any?>?` field to `Case`, allowing integrating applications to pass stable key-value metadata at case creation time — independently of any user message.

## Why

The existing `sessionContext` on `AddMessageRequest` is the right place for per-message context (current page, entity in view, etc.), but it's only available from the first message onward. Integrators need a way to attach case-scoped metadata (tenant, user role, feature flags…) at creation time, before any message is sent.

## How

**Domain & persistence**
- `Case.kt`: new nullable `context` field
- `CaseNode.kt`: serialised as `contextJson: String?` (JSON, same pattern as `IntegrationConfigNode.parametersJson`). `toDomain`/`fromDomain` now take an `ObjectMapper` parameter.
- `Neo4jCaseRepository.kt`: injects `ObjectMapper`, passes it through to node conversion
- `Neo4jPersistenceConfiguration.kt`: passes `ObjectMapper` to the `Neo4jCaseRepository` bean

**HTTP layer**
- `CaseResource.kt`: exposes `context` field (readable and writable)
- `CaseController.kt`: maps `context` in `toResource`/`toDomain`; `toDomainForUpdate` merges with existing value (update is non-destructive — omitting `context` on PUT preserves the existing value)

**Agent prompt injection**
- `AgentExecutionContext.kt`: new `caseContext` field, populated in `CaseServiceImpl.runAgent` from the persisted `Case`
- `AgentServiceImpl.kt`: passes `context.caseContext` to both `AgentAdvanced` and `AgentSimple` constructors
- `AgentAdvanced.kt` / `AgentSimple.kt`: new `caseContext` constructor parameter, passed through to all `buildMessages`/`convertEventsToMessages` calls
- `AgentAdvancedContext.kt`: `buildMessages` and `convertEventsToMessages` accept optional `caseContext`
- `MessageConversions.kt`: `sessionContextPromptText` now takes `caseContext` and merges — `caseContext + sessionContext`, with `sessionContext` winning on key conflicts

**Tests**
- `MessageConversionsUnitSpec.kt`: new unit spec covering all merge scenarios (null/null → null, caseContext only, sessionContext only, merge with conflict, XML wrapping, injection escaping)

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
@@ -45,6 +45,8 @@ class AgentAdvanced(
     private val userExternalId: String? = null,
     private val caseEventsProvider: () -> List<CaseEvent> = { emptyList() },
     private val maxIterations: Int = 20,
+    /** Case-level context merged with per-message sessionContext in every prompt. */
+    private val caseContext: Map<String, Any?>? = null,
 ) : Agent {
     override fun run(
         events: List<CaseEvent>,
@@ -361,7 +363,7 @@ class AgentAdvanced(
         caseId: UUID,
         emitEvent: suspend (CaseEvent) -> Unit,
     ): GateOutcome {
-        val history = context.buildMessages(accumulatedEvents)
+        val history = context.buildMessages(accumulatedEvents, caseContext)
         val needsExplicit =
             when (tool.confirmationMode) {
                 ConfirmationMode.EVERY_TIME -> {
@@ -593,7 +595,7 @@ class AgentAdvanced(
                         // Slice the history shown to analyzeConfirmation to events at-or-after the pending.
                         // Prevents the LLM judge from picking up a "yes" from an unrelated earlier turn
                         // (defense-in-depth for tools with confirmationMode=EVERY_TIME).
-                        val historyFromPending = context.buildMessages(events.subList(pendingIndex, events.size))
+                        val historyFromPending = context.buildMessages(events.subList(pendingIndex, events.size), caseContext)
                         val decision =
                             context.confirmationManager.analyzeConfirmation(
                                 chatClient = context.chatClient,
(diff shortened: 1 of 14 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
```diff
@@ -45,6 +45,8 @@ class AgentAdvanced(
     private val userExternalId: String? = null,
     private val caseEventsProvider: () -> List<CaseEvent> = { emptyList() },
     private val maxIterations: Int = 20,
+    /** Case-level context merged with per-message sessionContext in every prompt. */
+    private val caseContext: Map<String, Any?>? = null,
```
BenoitAvertyWhoz, line 49: question: Serait-ce possible d'avoir une map non-nullable et de prendre une map vide en cas d'absence ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt
```diff
@@ -280,9 +285,10 @@ class AgentSimple(
         events.forEachIndexed { index, event ->
             when (event) {
                 is MessageEvent -> {
-                    // Inject session context as a UserMessage immediately before the last user message.
+                    // Inject merged context (caseContext + sessionContext) as a UserMessage
+                    // immediately before the last user message.
                     if (index == lastUserMessageIndex) {
-                        event.sessionContextPromptText()?.let { messages.add(UserMessage(it)) }
+                        event.sessionContextPromptText(caseContext)?.let { messages.add(UserMessage(it)) }
```
BenoitAvertyWhoz, line 291: Je pense qu'il faut les traiter différemment. à mon avis l'intérêt d'avoir deux contextes c'est de pouvoir les mettre à différents endroits dans la conversation : le case context au début de la conversation, et le session context ou message context au niveau des messages utilisateur.

ça serait potentiellement un bon endroit pour injecter le contexte whoz global @vincent-couturier-whoz @yannick-bride-whoz

vincent-audibert-whoz:
Dropping...

PR #939, opened 2026-06-02
feat(wz-32284): add dynamic confirmation mode hook for tools

selimbensenouciep-prog:
## Summary

Extends `StandardTool` with a **dynamic** `getConfirmationMode(argsJson, context)` hook and **routes `getConfirmationInstructions()`** into the `shouldConfirm` prompt template (tour 1) in addition to its existing use in `analyzeConfirmation` (tour 2). Paired with whoz-oss/whoz#5076 which uses these hooks to reproduce the legacy Copilot `UpdateProfileTool.shouldBypassProgrammatically` and to inject a "pourquoi pas not consent" guidance block for `UpdateProfile`.

## Why

Observed on `version-rc` (case `e4e08a89-23b5-4d47-a5d7-f962822bd5e7`, 2026-06-02):

1. **Bypass programmatique perdu** — legacy Copilot `UpdateProfileTool.kt:124-150` skips `confirmationManager.shouldConfirm` when `chatContext.previousActions` contains a `CreateProfile` action for the same `profileId` (profile-filling UX). In `agentos` mode, `AgentOsConfirmationManager` is a no-op and the SDK had no equivalent hook — so the user sees a confirmation on every `UpdateProfile` even right after `CreateProfile`. The static `val confirmationMode` (introduced in WZ-31596) cannot express args-dependent decisions.

2. **LLM-judge trop laxiste** — *"pourquoi pas"* answered to *"Veux-tu que j'ajoute cette expérience…"* was classified CONFIRMED by `analyzeConfirmation` with no `specificInstructions` (because `getConfirmationInstructions()` was empty by default). The same guidance, applied at decision-time (`shouldConfirm` INFER), would also help prevent inferring consent from ambiguous conversation context.

### 1. `StandardTool.getConfirmationMode(args, ctx)` — dynamic mode hook

```kotlin
suspend fun getConfirmationMode(
    argsJson: String? = null,
    context: ToolContext? = null,
): ConfirmationMode = confirmationMode   // default delegate to static val
```

Backward-compat preserved: all existing plugins (file, datetime, bash, tmux, filesystem-aiproviders) inherit the default that reads the static `val confirmationMode` — no change required on their side. The Copilot plugin (whoz#5076) overrides this hook to inspect `ToolContext.caseEvents` and return `ConfirmationMode.NONE` when the legacy bypass condition holds.

### 2. `AgentAdvanced.handleConfirmationGate` — read dynamic mode + forward instructions

[...]

PR #943, opened 2026-06-02
feat(wz-32284): dynamic confirmation mode resolution hook

selimbensenouciep-prog:
## Summary

Adds `suspend fun getConfirmationMode(argsJson, context): ConfirmationMode` on `StandardTool` so plugins can resolve the confirmation mode at runtime from the proposed args and the current case events, instead of being limited to the static `val confirmationMode`.

This is the **first of two** companion PRs splitting the original WZ-32284 work:

1. **This PR — dynamic mode hook** : the runtime extension required for plugins to do args-dependent gating (e.g. legacy bypass-programmatically behaviour for UpdateProfile).
2. **Follow-up — tool guidance injection** : routes `getConfirmationInstructions()` into the `shouldConfirm` prompt template, on top of the dynamic mode hook.

## Why

The AgentOS Copilot plugin needs to reproduce the Whoz Copilot legacy `UpdateProfileTool.shouldBypassProgrammatically` behaviour: bypass the confirmation prompt when a `CreateProfile` tool call earlier in the same conversation emitted a `PROFILE_CREATED` action on the same `profileId`. Without a dynamic hook, this requires either an HTTP round-trip back to Copilot for every tool call, or a per-call override of the static val, neither of which scales cleanly.

### `StandardTool.kt` — new hook

```kotlin
suspend fun getConfirmationMode(
    argsJson: String? = null,
    context: ToolContext? = null,
): ConfirmationMode = confirmationMode   // default delegates to the static val
```

- Backward-compat preserved: every existing plugin (file, datetime, bash, tmux, filesystem-aiproviders) inherits the default that reads the static `confirmationMode`. No change required on their side.
- KDoc explicitly flags the contract: **HOT PATH** — overrides MUST be cheap (no HTTP/DB/LLM) and side-effect-free (the orchestrator may retry within a single tool call). `suspend` is permitted but Kotlin-local matching on `context.caseEvents` is the intended pattern.

### `AgentAdvanced.kt` — read the dynamic mode + thread it through

The resolved mode is computed once per tool call before `handleConfirmationGate`:

```kotlin
val confirmationMode =
    if (tool != null && toolCtx != null) {
        tool.getConfirmationMode(parameters.args, toolCtx)
    } else {
        ConfirmationMode.NONE
    }
```

`handleConfirmationGate(mode: ConfirmationMode, …)` now takes the mode as an explicit parameter instead of re-reading `tool.confirmationMode` inside its `when` dispatch. The dispatch behaviour is unchanged.

## Test coverage

| Spec | New tests | Verifies |
|---|---|---|
| `StandardToolSpec` | 2 | Default delegate to static + dynamic override honored |
| `AgentAdvancedSpec` | 2 | E2E `NONE`-bypass-entirely (no `PendingConfirmation`, no `shouldConfirm` call) + seam test asserting that args/caseEvents reach the hook verbatim from the orchestrator |

Existing mocks of `mockTool.confirmationMode` updated with paired `coEvery { mockTool.getConfirmationMode(any(), any()) }` to keep them functional after the orchestrator now reads via the dynamic hook.

All `StandardToolSpec` (6/6) and `AgentAdvancedSpec` (32/32) pass locally.

## Backward compatibility

- Existing plugins keep working — `getConfirmationMode` default = `confirmationMode` (static val).
- `handleConfirmationGate` is private (only called by `AgentAdvanced` itself) — adding the `mode` param has no external impact.
- No DTO changes, no surface REST change.

## Paired PR

`feat(wz-32284): inject tool guidance in shouldConfirm prompt` will follow on top of this branch and add the `<tool_guidance>` injection. Splitting them keeps each diff focused on a single concern.

## Supersedes

Closes #939 (was bundling mode hook + guidance injection together).

## Test plan

- [x] `./gradlew :agentos-sdk:test --tests StandardToolSpec` → 6/6
- [x] `./gradlew :agentos-service:test --tests AgentAdvancedSpec` → 32/32
- [ ] E2E validated on version-rc (cf. transcript whoz#5076): create file via editFiles → remove → bypass confirmé via log `[getConfirmationMode] touchedInSession=true → NONE`

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
@@ -62,7 +62,9 @@ interface StandardTool<T> {
     // through its standard [execute] path — same as a tool that never opted in.
 
     /**
-     * Declares how this tool participates in the user-confirmation flow.
+     * Static fallback for the confirmation mode. Read directly only by tools that
+     * don't need dynamic resolution. The orchestrator reads via [getConfirmationMode]
+     * which defaults to this value.
      *
      * - [ConfirmationMode.NONE]: no confirmation required — tool executes directly (default).
      * - [ConfirmationMode.INFER]: confirmation required, but the orchestrator may skip
@@ -72,6 +74,35 @@ interface StandardTool<T> {
      *   is never trusted. Use this for irreversible side-effects (e.g. file deletion).
      */
     val confirmationMode: ConfirmationMode get() = ConfirmationMode.NONE
+
+    /**
+     * Dynamic resolution of the confirmation mode. The orchestrator calls this with the
+     * proposed args and current case events before deciding whether to gate the tool call.
+     * Override to compute the mode from business logic — e.g. inspect prior tool calls in
+     * [ToolContext.caseEvents] to bypass confirmation when an in-session create makes a
+     * follow-up update implicit.
+     *
+     * Default delegates to the static [confirmationMode] for backward compatibility:
+     * existing plugins that only override [confirmationMode] keep working without change.
+     *
+     * ⚠️ HOT PATH — called on **every** tool call by the orchestrator, including those that
+     * resolve to [ConfirmationMode.NONE]. Overrides MUST be:
+     *   - **cheap** : no HTTP, no DB call, no LLM. Pure local computation on `argsJson`
+     *     and `context.caseEvents` only.
+     *   - **side-effect-free** : the orchestrator may call this multiple times for the
+     *     same tool call (e.g. on retry). Mutating state, writing logs at INFO+, or
+     *     emitting events here will leak / duplicate / mislead.
+     *
+     * `suspend` is allowed (the orchestrator awaits) so simple async lookups stay possible,
+     * but in practice prefer plain Kotlin matching against `context.caseEvents`.
+     *
+     * @param argsJson Raw JSON args produced by the LLM for the impending tool call
+     * @param context Execution context (namespaceId, userId, caseEvents)
+     */
+    suspend fun getConfirmationMode(
+        argsJson: String? = null,
+        context: ToolContext? = null,
+    ): ConfirmationMode = confirmationMode
 
     /**
      * Instructions appended to the `analyzeConfirmation` prompt to guide the LLM judge
--- a/agentos/agentos-sdk/src/test/kotlin/io/whozoss/agentos/sdk/tool/StandardToolSpec.kt
+++ b/agentos/agentos-sdk/src/test/kotlin/io/whozoss/agentos/sdk/tool/StandardToolSpec.kt
@@ -67,5 +67,67 @@ class StandardToolSpec : StringSpec() {
                 dummyContext,
             ).output shouldBe "America/New_York"
         }
+
+        // getConfirmationMode(args, ctx) default = lit la val statique confirmationMode.
+        // Garantit la backward-compat : un plugin existant qui n'override que la val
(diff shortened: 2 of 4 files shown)


agentos/agentos-sdk/src/test/kotlin/io/whozoss/agentos/sdk/tool/StandardToolSpec.kt
```diff
@@ -67,5 +67,67 @@ class StandardToolSpec : StringSpec() {
                 dummyContext,
             ).output shouldBe "America/New_York"
         }
+
+        // getConfirmationMode(args, ctx) default = lit la val statique confirmationMode.
```
vincent-couturier-whoz, line 71: issue: french comment should be translated in english

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
```diff
@@ -62,7 +62,9 @@ interface StandardTool<T> {
     // through its standard [execute] path — same as a tool that never opted in.
 
     /**
-     * Declares how this tool participates in the user-confirmation flow.
+     * Static fallback for the confirmation mode. Read directly only by tools that
```
vincent-couturier-whoz, line 65: suggestion: Perhaps we should introduce a third possible value : DYNAMIC to allow fast processing in case of static value and have the overhead only when needed

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
```diff
@@ -73,6 +75,35 @@ interface StandardTool<T> {
      */
     val confirmationMode: ConfirmationMode get() = ConfirmationMode.NONE
 
+    /**
+     * Dynamic resolution of the confirmation mode. The orchestrator calls this with the
```
vincent-couturier-whoz, line 79: suggestion: as mentioned above it's better to use dynamic only when it's necessary.

agentos/agentos-sdk/src/test/kotlin/io/whozoss/agentos/sdk/tool/StandardToolSpec.kt
```diff
@@ -67,5 +67,67 @@ class StandardToolSpec : StringSpec() {
...
+                    override suspend fun execute(input: Unit?, context: ToolContext) =
+                        ToolExecutionResult.success("")
+                }
+            noneTool.getConfirmationMode() shouldBe ConfirmationMode.NONE
+            noneTool.getConfirmationMode("{}", dummyContext) shouldBe ConfirmationMode.NONE
+            everyTimeTool.getConfirmationMode() shouldBe ConfirmationMode.EVERY_TIME
+            everyTimeTool.getConfirmationMode("{}", dummyContext) shouldBe ConfirmationMode.EVERY_TIME
+        }
+
+        // Override dynamique : un plugin peut décider du mode à partir des args/events,
```
vincent-couturier-whoz, line 105: issue: french comment

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentAdvancedSpec.kt
```diff
@@ -1064,6 +1065,162 @@ class AgentAdvancedSpec :
             (response.output as MessageContent.Text).content shouldContain "boom"
         }
 
+        // Dynamique : un tool peut résoudre son mode au runtime via getConfirmationMode(args, ctx).
```
vincent-couturier-whoz, line 1068: Issue : french comment which should be translated in english

vincent-couturier-whoz:
already taken into account in another PR

PR #947, opened 2026-06-02
fix: #942 anchor scheduler nextRun to startTimestamp phase

### LINKED ISSUE #942: fix: scheduler nextRun drifts from configured time-of-day after missed executions
## Problem

When a scheduler misses its configured execution time (e.g., because Coday was shut down), the next run time drifts away from the intended time-of-day. 

For example, a daily scheduler configured to run at 8:30 AM (`startTimestamp: '2025-07-01T08:30:00Z'`, `interval: '1d'`) will execute at 11 PM if Coday is restarted at that time, and then schedule its next run for 11 PM the following day — permanently losing the 8:30 AM anchor.

## Root Cause

`calculateNextRun()` in `libs/utils/src/lib/interval-schedule.utils.ts` computes the next run as `fromDate + 1 × interval`, where `fromDate` is typically `new Date()` (current time). The `startTimestamp` is only used to gate the very first execution (`if fromDate < start, return start`), but is never used as a phase anchor for subsequent occurrences.

This affects two code paths:

1. **`executeScheduler()`** in `scheduler.service.ts` — calls `calculateNextRun(schedule, new Date(), occurrenceCount)` after execution, anchoring the next run to "now" instead of the schedule's configured time.

2. **`calculateNextRunSkippingMissed()`** in `scheduler.service.ts` — called at startup to skip past missed executions, also passes `now` as `fromDate`, causing the same drift.

## Expected Behavior

`startTimestamp` should serve as a **phase anchor**. All future occurrences should be computed as `startTimestamp + N × interval` (finding the smallest N such that the result is strictly after `fromDate`). This ensures a "daily at 8:30" scheduler always runs at 8:30, regardless of when Coday was last running.

## Proposed Fix

Rewrite `calculateNextRun()` to step forward from `startTimestamp` in intervals rather than from `fromDate`:

- For uniform intervals (minutes, hours, days): compute N arithmetically — `N = ceil((fromDate - startTimestamp) / intervalDuration)`. This is O(1).
- For months: iterate from `startTimestamp` adding months until past `fromDate` (months are non-uniform, so arithmetic doesn't work, but iteration is bounded and fast).
- The `daysOfWeek` filtering layer remains unchanged — it already correctly iterates forward from the candidate.

With this fix, `calculateNextRunSkippingMissed()` in `scheduler.service.ts` may simplify significantly since `calculateNextRun(schedule, now, occurrenceCount)` would directly return the correct phase-aligned slot.

## Files Involved

- `libs/utils/src/lib/interval-schedule.utils.ts` — core fix in `calculateNextRun()`
- `libs/utils/src/lib/interval-schedule.utils.test.ts` — update/add tests for phase-stability
- `libs/service/src/lib/scheduler.service.ts` — `executeScheduler()` and `calculateNextRunSkippingMissed()` may need minor adjustments

david-palita-whoz:
Fixes #942

## Problem

`calculateNextRun()` computed the next run as `fromDate + 1 × interval`, where `fromDate` was typically `new Date()`. This meant that if Coday was restarted at 11 PM after being down all day, a "daily at 8:30 AM" scheduler would schedule its next run for 11 PM the next day — permanently losing the 8:30 AM anchor.

## Fix

Rewrote `calculateNextRun()` to use `startTimestamp` as a **phase anchor**. The next occurrence is now computed as `startTimestamp + N × interval` (smallest N such that the result is strictly after `fromDate`).

- **Minutes, hours, days** (uniform duration): N computed arithmetically in O(1)
- **Months** (non-uniform): iterates forward from startTimestamp (bounded, fast)
- **daysOfWeek filtering**: unchanged — still iterates forward from the candidate

Also simplified `calculateNextRunSkippingMissed()` in `SchedulerService` from an iterative loop to a single `calculateNextRun()` call, since the phase-anchored function now directly returns the correct next slot.

## Files changed

- `libs/utils/src/lib/interval-schedule.utils.ts` — core fix
- `libs/utils/src/lib/interval-schedule.utils.test.ts` — added phase-stability tests
- `libs/service/src/lib/scheduler.service.ts` — simplified `calculateNextRunSkippingMissed()`

## Tests

- 85 tests pass (6 suites), lint clean on both utils and service

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -599,39 +599,7 @@ importers:
         specifier: catalog:dev
         version: 5.9.3
 
-  libs/agent/dist:
-    dependencies:
-      rxjs:
-        specifier: 'catalog:'
-        version: 7.8.2
-      tslib:
-        specifier: 'catalog:'
-        version: 2.8.1
-      yaml:
-        specifier: 'catalog:'
-        version: 2.8.1
-    devDependencies:
-      typescript:
-        specifier: catalog:dev
-        version: 5.9.3
-
-  libs/coday-events/dist:
-    dependencies:
-      tslib:
-        specifier: ^2.3.0
-        version: 2.8.1
-
   libs/coday-services:
-    dependencies:
-      tslib:
-        specifier: 'catalog:'
-        version: 2.8.1
-    devDependencies:
-      typescript:
-        specifier: catalog:dev
-        version: 5.9.3
-
-  libs/coday-services/dist:
     dependencies:
       tslib:
         specifier: 'catalog:'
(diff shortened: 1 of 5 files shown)


mathieu-dearmey (approved):

pnpm-lock.yaml
david-palita-whoz, line 1: this shouldn't have change, will remove it from PR

david-palita-whoz: @mathieu-dearmey I replaced this PR with https://github.com/whoz-oss/coday/pull/974

PR #950, opened 2026-06-02
fix(ci): serialize release runs

### LINKED ISSUE #949: Release pipeline: concurrent releases produce wrong versions and orphan commits
## Context

The release pipeline (`release.yml`) is triggered on every push to `master`. When multiple PRs are merged in quick succession, multiple release runs start concurrently — each doing a full 18-minute pipeline that includes version bumping, git commit, tag, and push.

## Problem: Concurrent releases produce incorrect version sequences

When two PRs land on master within seconds of each other, two `release.yml` runs start almost simultaneously. Each run does a shallow checkout of master at the point it was triggered, resolves the current version from the latest git tag it can see, computes the next version, and pushes a release commit.

Because the second run starts before the first has pushed its release commit and tag, it resolves the version from the same base tag as the first run. The result:

- Run A (starts first): resolves `0.168.0` → bumps to `0.169.0`, pushes commit `a25f1dc` with tag `release/0.169.0`
- Run B (starts ~2 minutes later, but from a checkout that predated Run A's push): also resolves `0.168.0` → bumps to `0.168.1`, pushes a new commit on top of `a25f1dc`

Run B's push **overwrites** Run A's release commit on master. The `0.169.0` commit becomes orphaned — not reachable from `HEAD`. The `release/0.169.0` tag still exists but points to a commit no longer on `master`.

This was observed on 2026-06-02:
- Release #796 produced `0.169.0` at commit `a25f1dc`
- Release #799 (started ~5 minutes later from a stale checkout) produced `0.168.1`, pushing on top and orphaning `0.169.0`
- Subsequent releases produced `0.168.2`, which became the final HEAD on master

mathieu-dearmey:
## What

Add `concurrency` group to `release.yml` to ensure only one release runs at a time — subsequent triggers queue and wait instead of running in parallel.

## Why

Two PRs merging close together triggered simultaneous release runs. Each resolved the same base tag, produced conflicting versions, and the second push overwrote the first — leaving `0.169.0` orphaned on master.

Closes #949

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,10 @@ on:
       - master
   workflow_dispatch:
 
+concurrency:
+  group: release
+  cancel-in-progress: false
+
 env:
   NPM_CONFIG_PROVENANCE: true
   NX_MAX_WORKERS: 4


vincent-audibert-whoz (approved):
Let’s try that to see if it does what it pretends 👍

PR #978, opened 2026-06-05
fix: #32491 detect redirect loop in RedirectTool

vincent-audibert-whoz:
## Problem

Two agents (PCT and DB) were ping-ponging indefinitely via the redirect tool — each delegating to the other without making progress, resulting in 50+ tool calls before hitting the max-iterations limit.

The existing repetition detection in `AgentAdvanced` only looks at `ToolResponseEvent`s within a single agent run. Since each redirect starts a fresh agent run, the loop was invisible to that mechanism.

## Solution

`RedirectTool.execute` now inspects `ToolContext.caseEvents` before throwing `AgentInterrupt.Redirect`. The `detectRedirectLoop` method scans the last `REDIRECT_LOOP_WINDOW = 10` `AgentSelectedEvent`s since the last **user** `MessageEvent` (agent `MessageEvent`s are correctly excluded). If the target agent appears `REDIRECT_LOOP_THRESHOLD = 2` or more times in that window, the tool returns a clear error instead of redirecting:

> "Redirect loop detected: 'X' has already been invoked during this turn. Stop redirecting and answer the user directly."

The LLM receives this as a tool response and can answer the user directly.

## Test coverage

- Loop detected when target reaches threshold in window
- No false positive below threshold
- Selections outside the window are ignored
- Agent `MessageEvent` between selections does not reset the window
- Selections from a previous turn (before last user message) do not count"

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
@@ -1,6 +1,9 @@
 package io.whozoss.agentos.redirect
 
 import io.whozoss.agentos.agent.AgentInterrupt
+import io.whozoss.agentos.sdk.actor.ActorRole
+import io.whozoss.agentos.sdk.caseEvent.AgentSelectedEvent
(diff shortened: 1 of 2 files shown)


vincent-audibert-whoz:
PR to put on hold, the redirectTool is not the best place to handle the repetition.

PR #983, opened 2026-06-07
fix: #WZ-32024 prevent `@CreatedBy` from being overridden on updates

fdelsert-whoz:
### Problem
On `AgentConfigNode`, `@CreatedBy` was being overridden on every update instead of only on creation, causing the creator field to be incorrectly changed each time the entity was saved.

### Root cause
The `AgentConfigNode` ID is an application-generated UUID. Spring Data's default `isNew()` logic assumes that an entity is new when its ID is `null` or database-generated. With an application-assigned UUID, the entity always appears as "already existing" to `isNew()`, so the auditing mechanism cannot distinguish creation from update. As a result, `@CreatedBy` is treated the same way on every `save()` operation and gets overwritten on each update.

### Fix
Because `isNew()` cannot be used reliably with an application-generated UUID, a version-based condition is used instead to detect initial creation. The version field (managed by Spring Data Neo4j for optimistic locking) is null on first save and non-null on subsequent updates, allowing @CreatedBy to be set only once. This field is kept internal and excluded from API serialization via @JsonIgnore.

The test *"save on update preserves original createdBy when auditor changes"* was written first to reproduce and prove the bug, no existing test covered this case. The test was failing initially, confirming the incorrect behaviour. After implementing the fix, the test passes, ensuring `@CreatedBy` is only set on creation and remains unchanged on subsequent updates even when the auditor changes.

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/entity/EntityMetadata.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/entity/EntityMetadata.kt
@@ -11,9 +11,14 @@ import java.util.UUID
  * - Identity (id)
  * - Audit trail (created, createdBy, modified, modifiedBy)
  * - Soft delete (removed)
+ * - Optimistic locking version (version)
  *
  * This class will be annotated with Spring Data annotations when we add database persistence.
  * Spring Data will handle automatic timestamp updates on save.
+ *
+ * [version] is used by Spring Data Neo4j for optimistic locking and for the `isNew()` heuristic:
+ * a null version means the entity has never been persisted, so Spring Data treats it as new
+ * (triggering `@CreatedBy`). A non-null version means the entity already exists (update path).
  *
  * @JsonIgnoreProperties(ignoreUnknown = true) allows forward-compatible deserialization
  * when new fields are added to the metadata in future versions.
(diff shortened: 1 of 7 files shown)


vincent-audibert-whoz (dismissed):

frederic-delsert-whoz:
Je repasse en DRAFT, c'est KO pour les noeuds existants :

```
org.springframework.dao.OptimisticLockingFailureException: An entity with the required version does not exist.
	at org.springframework.data.neo4j.core.Neo4jTemplate.saveImpl(Neo4jTemplate.java:464)
	at org.springframework.data.neo4j.core.Neo4jTemplate.lambda$save$14(Neo4jTemplate.java:386)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:140)
	at org.springframework.data.neo4j.core.Neo4jTemplate.save(Neo4jTemplate.java:386)
	at org.springframework.data.neo4j.repository.support.SimpleNeo4jRepository.save(SimpleNeo4jRepository.java:120)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:360)
```

vincent-audibert-whoz (approved):

PR #991, opened 2026-06-09
feat: #WZ-32424 enrich redirect tool description with agent integrations and exclude calling agent

vincent-audibert-whoz:
## What

Enriches the `RedirectTool` description seen by the LLM with two pieces of information per eligible agent:
- the list of its integrations
- for each integration that has a tool whitelist, the list of allowed tools (omitted otherwise)

Also ensures the calling agent is never listed as a redirect target (self-exclusion).

## Why

The LLM currently only sees agent name + description when choosing a redirect target. Adding integrations gives it the key differentiator between agents with similar descriptions — it can now route to the right agent based on what integrations it has access to (e.g. JIRA, FILES, etc.) and which specific tools are whitelisted.

The self-exclusion prevents the LLM from redirecting back to the agent that owns the redirect tool, which would create a no-op loop.

## Changes

- **`ToolContext` (SDK)** — new nullable `agentName: String? = null` field; rétrocompatible with all existing plugins
- **`ToolResolverService`** — passes `agentName` down into the `ToolContext` built at `provideTools` call time
- **`AgentServiceImpl`** — passes `config.name` to both `resolveTools*` calls
- **`RedirectToolPlugin`** — filters out the calling agent from eligible targets; maps `AgentConfig.integrations` onto `EligibleAgent.integrations`
- **`RedirectTool`** — new `Integration` inner class; `description` builder renders integrations and optional tool whitelists
- **Tests** — new cases covering integration rendering, self-exclusion, and propagation from `AgentConfig`

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolContext.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolContext.kt
@@ -19,10 +19,15 @@ import java.util.UUID
  * @param caseEvents A live snapshot of the current case's event list at the moment
  *   of tool invocation. Evaluated lazily — each call reflects events added during
  *   the current agent run, including prior tool responses in the same turn.
+ * @param agentName The name of the agent that owns this tool invocation, or null when
+ *   the context is not associated with a specific agent (e.g. tool-set resolution at
+ *   namespace level). Plugins may use this to exclude the running agent from lists they
+ *   build (e.g. the redirect tool excludes the calling agent from eligible targets).
  */
 data class ToolContext(
     val namespaceId: UUID,
     val userId: UUID?,
     val userExternalId: String?,
     val caseEvents: List<CaseEvent>,
+    val agentName: String? = null,
 )
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentExecutionContext.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentExecutionContext.kt
@@ -1,6 +1,7 @@
 package io.whozoss.agentos.agent
 
 import io.whozoss.agentos.sdk.caseEvent.CaseEvent
+import io.whozoss.agentos.sdk.tool.ToolContext
 import java.util.UUID
 
 /**
@@ -23,4 +24,16 @@ data class AgentExecutionContext(
     val caseId: UUID? = null,
     val userId: UUID? = null,
     val caseEventsProvider: () -> List<CaseEvent> = { emptyList() },
-)
+) {
+    fun toToolContext(
+        userExternalId: String?,
+        agentName: String?,
+    ): ToolContext =
+        ToolContext(
+            namespaceId = namespaceId,
+            userId = userId,
+            userExternalId = userExternalId,
+            caseEvents = caseEventsProvider(),
+            agentName = agentName,
+        )
+}
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
@@ -37,13 +37,32 @@ class RedirectTool(
     data class Input(val agentName: String)
 
     /**
+     * One integration available to an eligible agent, optionally restricted to a
+     * specific set of tool names.
+     *
+     * @param name Integration name (matches [IntegrationConfig.name] or
+     *   [ToolPlugin.integrationType] for config-less plugins).
+     * @param allowedTools When non-null, only these tool names (or suffixes for
+     *   multi-instance tools) are available. When null, all tools of the integration
+     *   are available — the list is omitted from the prompt in that case.
+     */
+    data class Integration(val name: String, val allowedTools: List<String>?)
+
+    /**
      * A resolved agent that the LLM may redirect to.
      *
      * @param name Exact agent name as stored in [AgentConfig.name].
      * @param description Human-readable description injected into the tool schema
      *   so the LLM can reason about which agent to choose without a separate list call.
+     * @param integrations Integrations available to this agent, derived from
+     *   [AgentConfig.integrations]. Empty when the agent has no restriction (all
+     *   namespace tools). Each entry may optionally carry a tool whitelist.
      */
-    data class EligibleAgent(val name: String, val description: String?)
+    data class EligibleAgent(
+        val name: String,
+        val description: String?,
+        val integrations: List<Integration> = emptyList(),
+    )
 
     override val name: String = configName?.let { "${it}__redirect" } ?: "redirect"
 
@@ -54,6 +73,14 @@ class RedirectTool(
             agent.description?.takeIf { it.isNotBlank() }
                 ?.also { desc -> appendLine("  - ${agent.name}: $desc") }
                 ?: run { appendLine("  - ${agent.name}") }
+            if (agent.integrations.isNotEmpty()) {
+                appendLine("    Integrations:")
+                agent.integrations.forEach { integration ->
+                    integration.allowedTools
+                        ?.also { tools -> appendLine("      - ${integration.name}: ${tools.joinToString(", ")}") }
+                        ?: run { appendLine("      - ${integration.name}") }
+                }
+            }
         }
     }.trimEnd()
 
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
@@ -70,15 +70,18 @@ class RedirectTool(
         appendLine("Route the current request to another agent. Use this when the request is better handled by a specialised agent.")
         appendLine("Available agents:")
         eligibleAgents.forEach { agent ->
-            agent.description?.takeIf { it.isNotBlank() }
-                ?.also { desc -> appendLine("  - ${agent.name}: $desc") }
-                ?: run { appendLine("  - ${agent.name}") }
+            val desc = agent.description?.takeIf { it.isNotBlank() }
+            when (desc) {
+                null -> appendLine("  - ${agent.name}")
+                else -> appendLine("  - ${agent.name}: $desc")
+            }
             if (agent.integrations.isNotEmpty()) {
                 appendLine("    Integrations:")
                 agent.integrations.forEach { integration ->
-                    integration.allowedTools
-                        ?.also { tools -> appendLine("      - ${integration.name}: ${tools.joinToString(", ")}") }
-                        ?: run { appendLine("      - ${integration.name}") }
+                    when (val tools = integration.allowedTools) {
+                        null -> appendLine("      - ${integration.name}")
+                        else -> appendLine("      - ${integration.name}: ${tools.joinToString(", ")}")
+                    }
                 }
             }
         }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
@@ -74,8 +74,20 @@ class RedirectToolPlugin(
             ?: listOf("*")
 
         val userId = context.userId
+        val callingAgentName = context.agentName
         val eligibleAgents = agentResolver(namespaceId, userId, patterns)
-            .map { RedirectTool.EligibleAgent(name = it.name, description = it.description) }
+            .filter { agentConfig -> agentConfig.name != callingAgentName }
+            .map { agentConfig ->
+                RedirectTool.EligibleAgent(
+                    name = agentConfig.name,
+                    description = agentConfig.description,
+                    integrations = agentConfig.integrations
+                        ?.map { (integrationName, allowedTools) ->
+                            RedirectTool.Integration(name = integrationName, allowedTools = allowedTools)
+                        }
+                        ?: emptyList(),
+                )
+            }
 
         if (eligibleAgents.isEmpty()) {
             logger.warn { "[RedirectToolPlugin] No eligible agents found for namespace $namespaceId with patterns $patterns" }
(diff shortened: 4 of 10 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -74,8 +74,20 @@ class RedirectToolPlugin(
             ?: listOf("*")
 
         val userId = context.userId
+        val callingAgentName = context.agentName
```
vincent-audibert-whoz, line 77: Self-exclusion is a string equality check on `agentConfig.name`, which is a mutable, admin-editable display name. Two problems:
1. If two agents share the same name, the wrong one gets excluded.
2. If the calling agent is renamed between the time the config was loaded and when `provideTools` runs, the exclusion silently disappears.

The correct anchor for identity here is the config's opaque ID (UUID), not its display name. `ToolContext.agentName` would need to become `agentConfigId: UUID?` (or carry both) to make this exclusion reliable. This is a correctness issue: the self-exclusion guarantee is weaker than it appears.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -74,8 +74,20 @@ class RedirectToolPlugin(
...
         val eligibleAgents = agentResolver(namespaceId, userId, patterns)
-            .map { RedirectTool.EligibleAgent(name = it.name, description = it.description) }
+            .filter { agentConfig -> agentConfig.name != callingAgentName }
+            .map { agentConfig ->
+                RedirectTool.EligibleAgent(
+                    name = agentConfig.name,
```
vincent-audibert-whoz, line 82: The integrations mapped here come from `agentConfig.integrations` — i.e., what the **calling agent's config** declares about the target, not what the target agent itself resolved at runtime. If the target agent has integrations configured at namespace level that are not in this map, they will be absent from the LLM description. The LLM may then route incorrectly because it sees an incomplete picture of a target agent's capabilities.

This is a semantic gap: the description is built from a partial projection of the target's config, not from the target's actual resolved tool set.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
```diff
@@ -36,14 +36,33 @@ class RedirectTool(
 ) : StandardTool<RedirectTool.Input> {
     data class Input(val agentName: String)
 
+    /**
+     * One integration available to an eligible agent, optionally restricted to a
+     * specific set of tool names.
+     *
+     * @param name Integration name (matches [IntegrationConfig.name] or
+     *   [ToolPlugin.integrationType] for config-less plugins).
+     * @param allowedTools When non-null, only these tool names (or suffixes for
```
vincent-audibert-whoz, line 45: `Integration` is a pure data class with no behaviour tied to `RedirectTool`. Nesting it inside the tool creates a coupling that doesn't exist in the domain — `Integration` is a concept of agent configuration, not of the redirect tool's execution logic. The same applies to `EligibleAgent`. Both would be cleaner as top-level classes (in the `redirect` package or a shared domain file), which also simplifies references in tests and in `RedirectToolPlugin`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
```diff
@@ -54,6 +73,14 @@ class RedirectTool(
             agent.description?.takeIf { it.isNotBlank() }
```
vincent-audibert-whoz: The `also { … } ?: run { … }` pattern is used twice here (lines 73–74 and 78–80). The code style guide prefers `when` over chained if/else for two-branch conditions. Both of these are effectively `if (x != null) … else …` and would be clearer expressed as:
```kotlin
when (val desc = agent.description?.takeIf { it.isNotBlank() }) {
    null -> appendLine("  - ${agent.name}")
    else -> appendLine("  - ${agent.name}: $desc")
}
```
This is a style issue per the project's own code-style doc, not a blocker.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -19,6 +19,7 @@ class ToolResolverService(
     fun resolveToolsForNamespace(
         namespaceId: UUID,
         agentIntegrations: Map<String, List<String>?>? = null,
+        agentName: String? = null,
```
vincent-audibert-whoz: The new `agentName` parameter on both `resolveToolsForNamespace` and `resolveToolsForRun` has no KDoc entry. The existing parameters are all documented. Consistency requires a `@param agentName` entry explaining what it is and when it is null — especially since the nullability carries semantic meaning (namespace-level resolution vs. agent-run resolution).

vincent-audibert-whoz (commented):
The PR delivers its two stated goals cleanly. No security or structural issues. The findings below are all design-level and worth discussing before merge.

**Two semantic gaps worth addressing.** First, the integrations surfaced in the redirect description come from `AgentConfig.integrations` (the calling agent's view of the target's config), not from what the target agent actually resolved at runtime. If the target agent has integrations that were not declared in the calling agent's config, they will be silently absent from the description — the LLM's view is incomplete and can be misleading. Second, and more critically, `agentName` in `ToolContext` is sourced from `config.name` which is the display name, a mutable string. The self-exclusion filter (`agentConfig.name != callingAgentName`) is a string equality check on this mutable, admin-editable value. If two agents share a name, one will be incorrectly excluded; if the calling agent's name is renamed between config load and redirect resolution, the exclusion silently disappears. An opaque identity (config ID or UUID) would be the correct anchor for self-exclusion.

**Minor design issues.** The `Integration` inner class inside `RedirectTool` is a pure data carrier with no behaviour tied to the tool — it belongs as a top-level class or alongside `EligibleAgent` in a shared domain file, not nested inside the tool. The `also`/`run` pattern used twice in the description builder (once for agent descriptions, once for integration tool lists) is idiomatic but unusual enough that a `buildString` with a plain `if`/`else` or `let`/`?: run` would be more readable to someone unfamiliar with the codebase. The code style guide also prefers `when` over chained if/else for two-branch conditions — both `also`/`run` pairs qualify. Finally, `ToolResolverService` now accepts `agentName` but the KDoc on neither function was updated to document the new parameter.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
vincent-audibert-whoz: False issue, the name is a critical part of the identity of agents.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
vincent-audibert-whoz: False issue, these integrations are those of the eligible agents.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectTool.kt
vincent-audibert-whoz: Pure internal to the RedirectTool, dismissed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
```diff
@@ -74,8 +74,20 @@ class RedirectToolPlugin(
             ?: listOf("*")
 
         val userId = context.userId
+        val callingAgentName = context.agentName
         val eligibleAgents = agentResolver(namespaceId, userId, patterns)
-            .map { RedirectTool.EligibleAgent(name = it.name, description = it.description) }
+            .filter { agentConfig -> agentConfig.name != callingAgentName }
```
selimbensenouciep-prog, line 79: Self-exclusion par `name` plutôt que `metadata.id: UUID` — invariant fragile. `AgentConfig.name` n'a pas de contrainte d'unicité (namespaceId, name) sur `AgentConfigNode` côté Neo4j, et `PUT /agent-configs/{id}` permet le rename libre via `AgentConfigController.toDomainForUpdate`. Scénario : rename d'AgentA → AgentX puis création d'un nouvel AgentA — le filtre `agentConfig.name != callingAgentName` peut sur-exclure ou sous-exclure. `metadata.id: UUID` est stable. Le LLM ne connaît que les names (l'`Input.agentName: String` reste utile côté tool input), mais le filtre interne pourrait s'appuyer sur l'UUID en comparant `agentConfig.metadata.id != context.agentId`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -139,6 +164,7 @@ class ToolResolverService(
                                     userId = userId,
                                     userExternalId = null,
                                     caseEvents = emptyList(),
+                                    agentName = agentName,
```
selimbensenouciep-prog: Duplication du builder `ToolContext` — ce site et celui de `resolveToolsForNamespace` (autour de la ligne 64 post-PR) ne diffèrent que sur `userId` (null vs param). La PR ajoute `agentName` aux 2 sites, l'occasion d'extraire une extension function ou un helper privé. Cf. pattern "centralize helpers" déjà utilisé sur AgentSimple/AgentAdvanced.

selimbensenouciep-prog (dismissed):

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolContext.kt
```diff
@@ -19,10 +19,15 @@ import java.util.UUID
...
+ *   the context is not associated with a specific agent (e.g. tool-set resolution at
+ *   namespace level). Plugins may use this to exclude the running agent from lists they
+ *   build (e.g. the redirect tool excludes the calling agent from eligible targets).
  */
 data class ToolContext(
     val namespaceId: UUID,
     val userId: UUID?,
     val userExternalId: String?,
     val caseEvents: List<CaseEvent>,
+    val agentName: String? = null,
```
selimbensenouciep-prog, line 32: Question sur le choix d'étendre `ToolContext` côté SDK pour ce besoin — `RedirectToolPlugin` est aujourd'hui le seul consommateur de `agentName` et il vit dans `agentos-service` (interne, pas un plugin externe PF4J). Aucun autre plugin (File, Bash, Case, …) ne consomme ce field. As-tu envisagé un canal interne pour passer le calling agent à `RedirectToolPlugin.provideTools` — par exemple un param dédié sur la signature, ou une injection au moment où `RedirectConfiguration` instancie le plugin ? L'agentName est déjà disponible côté `AgentServiceImpl.findAgentByName` au call site de `resolveTools*`. Question d'altitude : est-ce qu'on ne gagne pas à garder le SDK minimal et router ce besoin interne par un autre canal, quitte à élargir le SDK plus tard si un vrai cas externe apparaît ?

vincent-audibert-whoz: D’accord sur l’idée de restreindre la surface du sdk, mais:
- le passage d’information par un canal parallèle implique une mécanique alternative de gestion des plugins et pour l’instant la réutilisation bête de l’API de plugin permet d’aller vite et sans différence interne/externe.
- cette information d’agentName qui exécute le tool est intéressante en soi pour les plugins, ne serait-ce que pour leur traça/log interne.
=> du coup rajouté dans le `ToolContext` pour ce coup.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
vincent-audibert-whoz: Factorisé, le `ToolContext` est utile dès en amont.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/RedirectToolPlugin.kt
vincent-audibert-whoz: Le `name` est un élément critique de l’identité de l’agent. Ce n’est pas un invariant mais il est utilisé pour adresser les agents. Le tool de redirection redirige aussi par `name` donc le masquage est nécessaire ici.
On a peut-être des trous dans l’unicité des name par niveau, mais ce sera opportun à adresser hors de cette PR ou pendant l’implem du niveau platform.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentExecutionContext.kt
```diff
@@ -23,4 +24,16 @@ data class AgentExecutionContext(
     val caseId: UUID? = null,
```
selimbensenouciep-prog, line 24: The `caseId` field is currently not consumed anywhere — `toToolContext` forwards only `caseEventsProvider` (not the id itself), `ToolContext` has no `caseId` field, and no service or plugin reads `context.caseId`. The KDoc on `caseEventsProvider` even hints at conditional behaviour ("Ignored when `caseId` is null") that isn't actually implemented. Either drop the field or wire it where the doc promises — keeping it as a documented-but-dead field is semantically confusing. Note: plugins that need the case id can already extract it from `ToolContext.caseEvents.firstOrNull()?.caseId`.

selimbensenouciep-prog (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentExecutionContext.kt
vincent-audibert-whoz: I’ll leave it for now because even if unused currently, it will be needed for case-aware tools (like the delegate tool to port the Coday functionality).

PR #1004, opened 2026-06-14
Integrate neo4j-migrations, downgrade embedded neo4j, clean up schema initializer

### LINKED ISSUE #1026: AgentOS remove nullable node properties when possible
Neo4j entities have a number of properties that are historically nullable (like `removed`), but it complexifies greatly the queries for no real reason.

While not too late, do per entity concerned:
- run a query to set these null attributes to their default value (eg `removed = false`)
- simplify neo4j queries accordingly

fdelsert-whoz:
From : https://github.com/whoz-oss/coday/pull/992

Introduce `neo4j-migrations-spring-boot-starter` as a Flyway equivalent for Neo4j, replacing the manual ApplicationRunner approach.  
Versioned .cypher files in classpath:neo4j/migrations/ are now applied automatically at startup and tracked via: `__Neo4jMigration` nodes.

Changes:

- Add `neo4j-migrations-spring-boot-starter` dependency (version centralized in _libs.versions.toml_)
- Integrate existing migrations V001 (unique constraints), V002 (version backfill), V003 (enabled backfill)
- Align embedded Neo4j version with production (`neo4j:5.26-community` confirmed in _docker-compose.yml_ ?)
- Remove `Neo4jSchemaInitializerSpec`

No functional change. All migrations are idempotent and compatible with existing environments.


agentos/agentos-service/build.gradle.kts
```diff
@@ -249,60 +218,14 @@ tasks.withType<Test> {
...
-// classpath is assembled.
+// Neo4j ships org.neo4j:neo4j-slf4j-provider which registers SLF4JLogBridge as an SLF4J
+// service provider. When it wins the ServiceLoader race over Logback, Spring Boot's
+// LogbackLoggingSystem fails with "LoggerFactory is not a Logback LoggerContext".
+// This global exclusion guarantees it never appears on any classpath configuration.
 configurations.all {
     exclude(group = "org.neo4j", module = "neo4j-slf4j-provider")
 }
 
-// Neo4j 2026.x (embedded engine) requires Netty 4.2.x for BoltServer.
```
vincent-couturier-whoz, line 264: suggestion: we should check with @thomasmartin-whoz if it's safe to remove those dependencies which seems necessary for neo4j-embedded 2026.x

agentos/gradle/libs.versions.toml
```diff
@@ -109,14 +108,17 @@ jackson-module-kotlin = { module = "com.fasterxml.jackson.module:jackson-module-
 # Neo4j
 spring-boot-starter-data-neo4j = { module = "org.springframework.boot:spring-boot-starter-data-neo4j" }
 neo4j-java-driver = { module = "org.neo4j.driver:neo4j-java-driver" }
-# Neo4j embedded engine (Community Edition) — for zero-Docker local deployment
-# Version 2026.x uses calendar versioning; keep aligned with the Bolt protocol
```
vincent-couturier-whoz, line 114: question : why this change as we need neo4j 2026.x community compliancy to be able to use vector index to be able to do semantic search

fdelsert-whoz: Yes, I had some doubts about the version being used.
If that’s the case, I wasn’t able to get neo4j-migrations working with our dependencies.

fdelsert-whoz: The docker-compose file is still using neo4j:5.26-community.

PR #1006, opened 2026-06-15
feat: #31851 add platform-level scope to IntegrationConfig

vincent-audibert-whoz:
## Summary

Introduces a **platform-level** configuration scope for `IntegrationConfig` — the new lowest-precedence tier in the reconciliation chain.

Platform configs are stored as `(namespaceId=null, userId=null, name)` rows and serve as environment-wide defaults visible across all namespaces without manual per-namespace setup.

### Core: 4-tier reconciliation (`ConfigMergeService`)

`resolve()` now folds four layers from lowest to highest precedence:

```
platform (null, null)  →  user-global (null, user)  →  namespace-shared (ns, null)  →  user×namespace (ns, user)
```

The platform layer is always consulted first; any more-specific layer overrides it key by key via the existing `MergeStrategy` — no change to merge semantics.

### `IntegrationConfig` service

- `requireScope` guard removed — the service now accepts all four scope modes; authorization is enforced in the controller where the security context is available
- `assertConsistentIntegrationTypeAcrossLayers` extended to also check the platform layer `(null, null, name)`
- `findPlatform()` added for listing platform-scoped configs

### `IntegrationConfig` controller

| Operation | Required condition |
|---|---|
| `POST` with `(null, null)` | `user.isAdmin == true` |
| `PUT` / `DELETE` on a platform entity | `user.isAdmin == true` |
| `GET` without params | any authenticated user |

`requireAdminForPlatform(namespaceId, userId)` is a shared guard used on create, update, and delete paths.

### Neo4j

`findActivePlatform()` query added — plain label scan filtered on `namespaceId IS NULL AND userId IS NULL`. No schema changes needed: `tripleKey` for `(null, null, name)` produces `_:_:name`, which is already valid and unique under the existing constraint.

## Out of scope

`AiProvider` follows the same pattern but is left for a follow-up iteration to keep this PR focused and reviewable.


vincent-audibert-whoz (commented):
## Backend Review — `feat: #31851 add platform-level scope to IntegrationConfig`

The architecture of the 4-tier reconciliation chain is sound and the PR description is unusually precise. The core data model, `ConfigMergeService`, Neo4j persistence, and the authorization split (service accepts all four scopes, controller enforces admin guard) are all correct. The test coverage is broad and the Kotest/MockK patterns are consistent with the codebase.

That said there are several issues to address before merging, one of which is a security defect.

**Security — GET platform list is open to all authenticated users, documented as "Super Admin only"**: The PR description, the OpenAPI table, and all test names say platform listing requires Super Admin. The implementation says the opposite: `list()` returns `findPlatform()` for any authenticated caller. Non-admin users can enumerate all platform-scoped integration names, types, and — critically — their `parameters` payload which may contain API URLs or other configuration details. This contradicts the stated design and is the only critical issue.

**Correctness — `findEffective` merges incorrectly**: `findEffective` is used for the descriptive path (system prompt, instructions, tool enumeration). The KDoc explicitly says it is *"not a parameter-level merge — it returns the raw config from the winning layer for each name"*, but the implementation calls `mergeStrategy.merge()` and folds with `sorted.first()` as the initial accumulator. This means the result's `id`, `namespaceId`, and `userId` come from the **lowest**-priority layer (e.g. the platform layer), not the winning one. The correct implementation is simply `configs.maxByOrNull { it.getPriority() }`.

**Correctness — `resolveToolsForRun` silently drops all tools when `agentIntegrations` is null**: `integrationNames = agentIntegrations?.keys?.toList() ?: emptyList()` makes a null `agentIntegrations` produce an empty filter, returning zero tools. The KDoc still says "When null, all namespace integrations are included." Any `AgentConfig` without an `integrations` field will now run with no tools. This is a silent breaking change that needs an explicit decision and consistent KDoc.

**Style — `requireAdminForPlatform` calls `getCurrentUser()` redundantly on every `update`/`delete`**: The guard unconditionally calls `userService.getCurrentUser()` even when `namespaceId != null || userId != null`, causing a user lookup on every non-platform mutation. Short-circuit with an early return when the scope is not platform.

**Style — `getPriority()` on the domain entity leaks reconciliation mechanics**: This function is consumed only by `IntegrationConfigServiceImpl`. A local `Comparator` in the service would keep `IntegrationConfig` a pure data holder. The Java-bean naming convention (`getPriority`) is also non-idiomatic Kotlin; a `val priority: Int` computed property would be preferred.

**Missing tests**: (1) The `IntegrationConfigControllerMvcIntegrationSpec` platform-GET test asserts `200 OK` for a non-admin user and is named "for any authenticated user" — it will need to flip to `403` once the security fix is applied, and a positive admin case must be added. (2) `IntegrationConfigServiceImplSpec` has no coverage for the new platform-layer check inside `assertConsistentIntegrationTypeAcrossLayers` (create NS-shared with same name as platform config but different `integrationType` → expect 409, and vice-versa).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -207,14 +217,19 @@ class IntegrationConfigController(
...
+        if (namespaceId == null && userId == null) {
+            return integrationConfigService.findPlatform().map { toResource(it) }
+        }
+
+        val resolvedNs = parseNamespaceParam(namespaceId)
         val all = integrationConfigService.findFiltered(
             namespaceId = resolvedNs,
             namespaceIsNone = namespaceId?.equals(NONE_SENTINEL, ignoreCase = true) == true,
-            callerId = me,
+            callerId = currentUser.id,
```
vincent-audibert-whoz: **[CRITICAL — Security] Platform listing is unrestricted**

```kotlin
// Platform scope: no namespaceId and no userId — list is open to all authenticated users
if (namespaceId == null && userId == null) {
    return integrationConfigService.findPlatform().map { toResource(it) }
}
```

The comment says "open to all authenticated users" but the PR description, the OpenAPI table, and every test name for this path says "Super Admin only". Any authenticated user can enumerate all platform-scoped configs including their `parameters` payload (API URLs, tenant IDs, etc.).

`currentUser` is already in scope just above this block. Add an admin guard before the early return, e.g. `requireAdminForPlatform(null, null)`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
```diff
@@ -76,43 +78,72 @@ class IntegrationConfigServiceImpl(
...
+                            override.getPriority()
+                        ) {
+                            logger.warn {
+                                "[IntegrationConfigService] Inconsistency detected as same level between configs ${base.id} and ${override.id}"
+                            }
+                        }
+                        mergeStrategy.merge(base, override)
+                    }
+                }
+            }
```
vincent-audibert-whoz, line 110: **[MAJOR — Correctness] `findEffective` applies a parameter merge when it should return the winning layer's raw entity**

```kotlin
val sorted = configs.sortedBy { it.getPriority() }
sorted.drop(1).fold(
    initial = sorted.first(),   // lowest-priority layer
) { base, override ->
    mergeStrategy.merge(base, override)
}
```

The KDoc states: *"This is intentionally not a parameter-level merge — it returns the raw config from the winning layer for each name."*

The implementation contradicts this in two ways:
1. It calls `mergeStrategy.merge()`, performing the parameter-level merging the KDoc forbids.
2. The fold's `initial` is `sorted.first()` (the **lowest**-priority entity). After folding, the result's `id`, `namespaceId`, and `userId` come from the platform layer, not the winning layer — breaking downstream code that uses those identity fields.

For the descriptive path (system prompt, tool enumeration) the correct implementation is `configs.maxByOrNull { it.getPriority() }`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -73,94 +28,77 @@ class ToolResolverService(
...
-
-        if (agentIntegrations != null) {
-            val orphans = agentIntegrations.keys - distinctNames
-            if (orphans.isNotEmpty()) {
-                logger.warn {
-                    "[ToolResolver] Agent declares integration(s) ${orphans.toSortedSet()} but no matching " +
-                        "IntegrationConfig found in namespace $namespaceId for user $userId — these tools will be empty."
+        val integrationNames = agentIntegrations?.keys?.toList() ?: emptyList()
+        val integrationConfigs = allIntegrationConfigs.filter { it.name in integrationNames }
+        val allTools =
```
vincent-audibert-whoz, line 36: **[MAJOR — Correctness] `agentIntegrations = null` now silently returns an empty tool set**

```kotlin
val integrationNames = agentIntegrations?.keys?.toList() ?: emptyList()
val integrationConfigs = allIntegrationConfigs.filter { it.name in integrationNames }
```

When `agentIntegrations` is `null`, `integrationNames` is `emptyList()`, so `integrationConfigs` is always empty and no tools are returned. The KDoc still says *"When null, all namespace integrations are included"*, but the new test asserts the opposite.

This is a silent breaking change: any `AgentConfig` without an `integrations` field will now run with zero tools. Decide the intended semantics and make the KDoc and implementation consistent. If "null means all" is preserved, the filter becomes: `val integrationConfigs = if (agentIntegrations == null) allIntegrationConfigs else allIntegrationConfigs.filter { it.name in agentIntegrations.keys }`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfig.kt
```diff
@@ -41,4 +41,25 @@ data class IntegrationConfig(
...
+     * | User × namespace   | non-null    | non-null | 3        |
+     *
+     * Namespace-shared (priority 2) intentionally overrides user-global (priority 1):
+     * namespace admins can enforce a config that supersedes user preferences.
+     * User × namespace (priority 3) lets the user restore a personal override
+     * scoped to that specific namespace.
+     */
+    fun getPriority(): Int {
+        val firstPriority = if (namespaceId == null) 0 else 2
+        val secondPriority = if (userId == null) 0 else 1
```
vincent-audibert-whoz: **[Nit — Style] `getPriority()` leaks reconciliation mechanics into the domain entity**

```kotlin
fun getPriority(): Int {
    val firstPriority = if (namespaceId == null) 0 else 2
    val secondPriority = if (userId == null) 0 else 1
    return firstPriority + secondPriority
}
```

`getPriority()` is consumed only by `IntegrationConfigServiceImpl.findEffective`. Having precedence/reconciliation logic on the domain data class means `IntegrationConfig` carries service-layer concerns. A `Comparator<IntegrationConfig>` (or a local `val priority` lambda) defined in `IntegrationConfigServiceImpl` would keep the domain model a pure data holder.

Also, the Java-bean getter naming convention is non-idiomatic Kotlin; a `val priority: Int get() = ...` computed property would be preferred if it stays on the entity.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigControllerMvcIntegrationSpec.kt
```diff
@@ -311,7 +312,13 @@ class IntegrationConfigControllerMvcIntegrationSpec : StringSpec() {
                 .andExpect(status().isNotFound)
         }
```
vincent-audibert-whoz, line 314: **[Minor — Test] Platform GET test asserts the wrong security expectation**

```kotlin
"LIST without params returns platform configs for any authenticated user" {
    mockMvc.perform(get("/api/integration-configs"))
        .andExpect(status().isOk)
```

This test is named "for any authenticated user" and asserts `200 OK` for `alice` (non-admin). Once the security fix is applied this must flip to `403`. A complementary test asserting `200 OK` for an admin user should also be added, otherwise the fix has no positive-path regression coverage.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImplSpec.kt
```diff
@@ -151,19 +151,11 @@ class IntegrationConfigServiceImplSpec : StringSpec() {
         // Triple-mode (story 6.1, AC1, AC2, FR27bis)
         // -------------------------------------------------------------------------
 
-        "create throws 400 when both namespaceId and userId are null (invariant)" {
+        "create succeeds with namespaceId=null and userId=null (platform scope)" {
             val service = newService()
```
vincent-audibert-whoz, line 155: **[Minor — Missing test] No coverage for the new platform-layer check in `assertConsistentIntegrationTypeAcrossLayers`**

The method now checks whether a platform-level config with the same `name` but a different `integrationType` already exists. This guard has no dedicated test case. Missing scenarios:
- Create a namespace-shared config with the same `name` as an existing platform config but a different `integrationType` → expect 409.
- Create a platform config with the same `name` as an existing namespace-shared config but a different `integrationType` → expect 409.

Without these, a future refactor could silently remove the guard.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
vincent-audibert-whoz: Doc error, rectified.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImplSpec.kt
vincent-audibert-whoz: added

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
vincent-audibert-whoz: Doc updated

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
vincent-audibert-whoz: doc updated

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfig.kt
vincent-audibert-whoz: Agreed, moved to service layer

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
```diff
@@ -76,43 +78,70 @@ class IntegrationConfigServiceImpl(
...
+            .findAllForNamespaceIdAndUserId(
+                namespaceId = namespaceId,
+                userId = userId,
+            ).groupBy { it.name }
+            .mapNotNull { (_, configs) ->
+                if (configs.size == 1) {
+                    configs.first()
+                } else {
+                    val sorted = configs.sortedWith(LAYER_COMPARATOR)
+                    sorted.drop(1).fold(
```
fdelsert-whoz, line 99: Remplaçable par un `reduce` ?

```kotlin
.mapNotNull { (_, configs) ->
    configs.sortedWith(LAYER_COMPARATOR).reduce { base, override ->
        if (LAYER_COMPARATOR.compare(base, override) == 0) {
            logger.warn {
                "[IntegrationConfigService] Inconsistency detected as same level between configs ${base.id} and ${override.id}"
            }
        }
        mergeStrategy.merge(base, override)
    }
}
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/reconciliation/ConfigMergeService.kt
```diff
@@ -81,20 +86,25 @@ class ConfigMergeService<T : Entity>(
...
      *
-     * @throws ConfigNotFoundException when none of the three layers contains a row.
+     * @throws ConfigNotFoundException when none of the four layers contains a row.
      */
     fun resolve(
         namespaceId: UUID,
         userId: UUID,
         name: String,
     ): T {
+        val platform = lookup.findByTriple(null, null, name)
```
fdelsert-whoz: Proposition : 

```kotlin
        listOf(
            null to null,           // platform
            namespaceId to null,    // namespace shared
            null to userId,         // user global
            namespaceId to userId,  // user + namespace
        )
            .mapNotNull { (ns, user) -> lookup.findByTriple(ns, user, name) }
            .reduceOrNull { base, layer -> mergeStrategy.merge(base, layer) }
            ?: throw ConfigNotFoundException(namespaceId, userId, name)
```

foldLayer devient inutile.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -66,101 +21,85 @@ class ToolResolverService(
...
```
fdelsert-whoz, line 36: Proposal : 
```kotlin
        val allTools = integrationConfigs.flatMap { config ->
            val plugin = toolRegistryService.findPlugin(config.integrationType)
            if (plugin == null) {
                logger.warn { "[ToolResolver] No plugin found for type ${config.integrationType}" }
                return@flatMap emptyList()
            }
            extractTools(
                allowedNames = agentIntegrations?.get(config.name),
                config = config,
                plugin = plugin,
                context = context,
            )
        }
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -66,101 +21,85 @@ class ToolResolverService(
...
-                        config = config.parameters,
-                        configName = config.name,
-                        context = context,
-                    ).filter { tool -> isToolAllowed(tool.name, config.name, allowedToolNames) }
-            tools.forEach { tool ->
-                if (resolved.containsKey(tool.name)) {
-                    logger.warn {
-                        "[ToolResolver] Tool name conflict: '${tool.name}' from integrationType='${config.integrationType}' overrides an existing entry"
-                    }
+    ): List<StandardTool<*>>? {
```
fdelsert-whoz, line 68: `List<StandardTool<*>>?` why not `List<StandardTool<*>>` ?

fdelsert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
```diff
@@ -228,5 +276,33 @@ class IntegrationConfigServiceImpl(
...
+         * | User-global      | null        | non-null | 1    |
+         * | Namespace-shared | non-null    | null     | 2    |
+         * | User × namespace | non-null    | non-null | 3    |
+         *
+         * Namespace-shared (rank 2) intentionally overrides user-global (rank 1):
+         * namespace admins can enforce a config that supersedes user preferences.
+         * User × namespace (rank 3) lets the user restore a personal override
+         * scoped to that specific namespace.
+         */
+        val LAYER_COMPARATOR: Comparator<IntegrationConfig> =
```
fdelsert-whoz, line 295: Another naive question: wouldn't it be more appropriate to fetch all the integrationConfig in a single request and do the ranking in the query (if possible via the Cypher query)? I don't know if it's technically possible, but I just want to know if this was considered or rejected?
This is out of scope for the current issue and corresponds more to a refactoring.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
vincent-audibert-whoz: Early returns are prohibited in Whoz-related codebases.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigServiceImpl.kt
vincent-audibert-whoz: Could be, the neo4j model and queries still need some drying, left for later indeed 👍

fdelsert-whoz (approved):

PR #1029, opened 2026-06-17
fix: wz-32561 add force stop intention to avoid execution hallucination

adrien-ogor-whoz:
**When the repetition guard triggers a force-stop, set `lastIntention` to a synthetic `IntentionGeneratedEvent` (toolName=`Answer`) describing the loop problem before breaking out of the iteration loop. This ensures `generateFinalResponse` receives an accurate analysis context and produces a grounded explanation to the user, rather than hallucinating a successful outcome.**

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
@@ -1,6 +1,7 @@
 package io.whozoss.agentos.agent
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import io.whozoss.agentos.agent.AgentIntentionGenerator.Companion.ANSWER_TOOL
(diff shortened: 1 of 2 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
yannick-bride-whoz, line 55: La fonction est tellement longue faudra la split pour la rendre plus clair un jour :(

yannick-bride-whoz (approved):

thibault-drillien-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt
adrien-ogor-whoz: :+1:

PR #1030, opened 2026-06-17
refactor: wz-31851 de-nullify removed on user, namespaces and agentConfigs

vincent-audibert-whoz:
Covers partially #1026


fdelsert-whoz (dismissed):

fdelsert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNodeNeo4jRepository.kt
```diff
@@ -40,18 +40,16 @@ interface AgentConfigNodeNeo4jRepository : Neo4jRepository<AgentConfigNode, Stri
...
+            MATCH (ns:Namespace {id: $namespaceId, removed: false})
+            MATCH (a:AgentConfig {removed: false})
+            WHERE (a.namespaceId = $namespaceId OR a.namespaceId IS NULL)
               AND a.enabled
               AND NOT COALESCE(a.removed, false)
               AND ($agentName IS NULL OR toLower(a.name) = toLower($agentName))
               AND (
                 u.isAdmin = true
-                OR EXISTS { MATCH (u)-[:MEMBER]->(g:UserGroup)-[:BELONGS_TO]->(ns) MATCH (a)-[:DEPLOYED_TO]->(g) WHERE NOT COALESCE(g.removed, false) }
+                OR EXISTS { MATCH (u)-[:MEMBER]->(g:UserGroup)-[:BELONGS_TO]->(ns) MATCH (a)-[:DEPLOYED_TO]->(g) WHERE g.removed IS NULL OR g.removed = false }
```
AlanMenantWhoz: Je ne sais pas si c'est un oubli mais ça fait la même choe qu'avant (ou alors on veut enlever COALESCE)

vincent-audibert-whoz: Missed that one, rolledbacked

fdelsert-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNodeNeo4jRepository.kt
```diff
@@ -17,7 +17,7 @@ interface AgentConfigNodeNeo4jRepository : Neo4jRepository<AgentConfigNode, Stri
         $$"""
             MATCH (a:AgentConfig)
             WHERE a.namespaceId = $namespaceId
-              AND NOT COALESCE(a.removed, false)
+              AND NOT a.removed
```
vincent-couturier-whoz, line 20: question: I don't recognize what we exchanged. Why do we need removed property here if we manage AgentConfig and RemovedAgentConfig separately ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
```diff
@@ -5,59 +5,43 @@ import mu.KLogging
...
+ * [io.whozoss.agentos.user.UserSchemaInitializer],
+ * [io.whozoss.agentos.agentConfig.AgentConfigSchemaInitializer],
+ * [io.whozoss.agentos.userGroup.UserGroupSchemaInitializer]).
+ */
 @Component
+@Order(1)
 @ConditionalOnExpression(
-    "'\${agentos.persistence.mode:embedded-neo4j}' == 'neo4j' " +
-        "or '\${agentos.persistence.mode:embedded-neo4j}' == 'embedded-neo4j'",
+    "'\${agentos.persistence.mode:in-memory}' == 'neo4j' " +
```
vincent-couturier-whoz, line 28: question: why the change from embedded-neo4j by default to in-memory ? I think both are wrong in any case 😄 but I would like to know your motivations.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserSchemaInitializer.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun run(args: ApplicationArguments) {
+        backfillRemoved()
+        ensureExternalIdUniqueConstraint()
+        ensureExternalIdIndex()
+    }
+
+    private fun backfillRemoved() {
+        val backfilled =
+            neo4jClient
+                .query("MATCH (u:User) WHERE u.removed IS NULL SET u.removed = false RETURN count(u) AS count")
```
vincent-couturier-whoz, line 37: suggestion: if there's no index on removed property this change will scan all nodes each time we start the service.... it could be a little long. I would wait for the change proposed by @fdelsert-whoz to use neo4j migration to avoid a run at each start and perhaps we could launch this way the migration before service startup to limit the impact of this migration on the service startup.

vincent-couturier-whoz, line 37: suggestion: if we cannot use easily neo4j migration now we could have a schema node with a version associated and a check of the version before running the migration queries. It's less powerfull than neo4j migration which more fine grained and have more tracability and ability to execute migrations before startup but it allows us to reduce the impact of the migrations scripts at each startup.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigNodeNeo4jRepository.kt
vincent-audibert-whoz: We did not manage to do that in that PR. The PR is about:
- `removed` not null to simplify queries
- the ADR about going beyond

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
vincent-audibert-whoz: That is an agent error, will have to track down why...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/user/UserSchemaInitializer.kt
vincent-audibert-whoz: In that case we might as well drop this PR entirely, given the ADR will send us in another direction.

vincent-audibert-whoz:
PR not usefull enough, closed. ADR  in #1052 .

PR #1031, opened 2026-06-17
feat: wz-32484 unify confirmationMode into single getConfirmationMode() entry point

fdelsert-whoz:
Remove the `val confirmationMode` property from `StandardTool`.
`getConfirmationMode()` is now the sole override point for the confirmation flow.
Eliminates the dual static/dynamic ambiguity without changing behavior.

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
@@ -53,18 +53,17 @@ interface StandardTool<T> {
 
     // ─── User-confirmation opt-in (WZ-31596) ───────────────────────────────────────────────
     //
-    // A tool opts in to the confirmation flow by setting [confirmationMode] to a value
-    // other than [ConfirmationMode.NONE]. The orchestrator then routes via the
+    // A tool opts in to the confirmation flow by overriding [getConfirmationMode] to return
+    // a value other than [ConfirmationMode.NONE]. The orchestrator then routes via the
     // [PendingConfirmationEvent] persistence cycle and calls [executeWithJson]
     // (post-confirmation) or [onRejected] (post-refusal).
     //
-    // When [confirmationMode] is [ConfirmationMode.NONE] (the default), the tool runs
+    // When [getConfirmationMode] returns [ConfirmationMode.NONE] (the default), the tool runs
     // through its standard [execute] path — same as a tool that never opted in.
 
     /**
-     * Static fallback for the confirmation mode. Read directly only by tools that
-     * don't need dynamic resolution. The orchestrator reads via [getConfirmationMode]
-     * which defaults to this value.
+     * Resolution of the confirmation mode. The orchestrator calls this with the proposed args
+     * and current case events before deciding whether to gate the tool call.
      *
      * - [ConfirmationMode.NONE]: no confirmation required — tool executes directly (default).
      * - [ConfirmationMode.INFER]: confirmation required, but the orchestrator may skip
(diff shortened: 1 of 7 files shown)


agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
```diff
@@ -102,7 +99,7 @@ interface StandardTool<T> {
     suspend fun getConfirmationMode(
```
vincent-audibert-whoz, line 99: The default implementation now returns `ConfirmationMode.NONE` directly, which is correct. However, removing `val confirmationMode` is a **breaking SDK API change**: any external plugin compiled against the previous SDK that overrides `confirmationMode` as a `val` will fail to compile or behave incorrectly at load time against this new version. The `agentos-sdk` version in `gradle/libs.versions.toml` must be bumped (semver: at minimum a minor bump, arguably a major if external plugins are already published against `1.0.0`) to signal the incompatibility to downstream consumers.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentAdvancedSpec.kt
```diff
@@ -61,7 +61,7 @@ import kotlin.io.path.writeText
 internal class TestRemoveTool(
     private val rootDir: java.nio.file.Path,
     override val name: String = "FILES__remove",
-    override val confirmationMode: ConfirmationMode = ConfirmationMode.EVERY_TIME,
+    private val confirmationMode: ConfirmationMode = ConfirmationMode.EVERY_TIME,
```
vincent-audibert-whoz: `TestRemoveTool` now stores `confirmationMode` as a `private val` constructor parameter and delegates `getConfirmationMode` to it. This is a reasonable test-helper pattern, but the private field is now named identically to the removed interface property, which could mislead future readers into thinking the interface property still exists. Consider renaming it (e.g. `forcedMode`) to make the indirection explicit and avoid confusion with the old pattern.

vincent-audibert-whoz (changes requested):
Agent:

This PR achieves its stated goal cleanly — the dual static/dynamic ambiguity is gone and all in-repo call sites are migrated. Two issues need addressing before merge.

**Branch name typo**: `featutre/fdelsert/...` contains a typo (`featutre` instead of `feature`). This violates the branch naming convention from DEV_WORKFLOW.md and is worth fixing for consistency.

**Breaking SDK change without a version bump**: Removing `val confirmationMode` from `StandardTool` is a breaking change to the public SDK interface. External plugins that override the property (as the old pattern encouraged, and as documented in the SDK README) will fail to compile against the new SDK without any signal from a version change. The `agentos-sdk` version in `gradle/libs.versions.toml` must be bumped to reflect this incompatibility.

Human: yeah, no real issues there.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/StandardTool.kt
vincent-audibert-whoz: ignored, we manage the version through the coday release

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentAdvancedSpec.kt
vincent-audibert-whoz: ...optional

vincent-audibert-whoz (dismissed):

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agent/AgentAdvancedSpec.kt
fdelsert-whoz: done at commit: feda9a9a1cf4059acfb07f71328ee3de60393abe

vincent-audibert-whoz (approved):

PR #1036, opened 2026-06-18
feat: WZ-31851 prepare permission platform level

vincent-audibert-whoz:
Prepare platform level for permission system by making `entityId` nullable as it will correspond to the absence of namespaceId, signaling a platform-level entity.

In that case:
- READ for all
- WRITE for super-admins only (`user.isAdmin == true`)

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
@@ -12,9 +12,8 @@ import mu.KLogging
  * PermissionNodeNeo4jRepository with its @Query annotations.
  */
 class Neo4jPermissionRepository(
-    private val permissionNodeRepository: PermissionNodeNeo4jRepository
+    private val permissionNodeRepository: PermissionNodeNeo4jRepository,
 ) : PermissionRepository {
-
     companion object : KLogging() {
         /**
          * Entity types where a namespace MEMBER does NOT gain transitive READ
@@ -33,22 +32,23 @@ class Neo4jPermissionRepository(
         userId: String,
         entityType: EntityType,
         entityId: String,
-        relation: PermissionRelation
-    ): Boolean {
-        return try {
+        relation: PermissionRelation,
+    ): Boolean =
+        try {
             when (relation) {
                 PermissionRelation.ADMIN -> {
                     permissionNodeRepository.hasAdminPermission(
                         userId = userId,
                         entityId = entityId,
-                        entityLabel = entityType.label
+                        entityLabel = entityType.label,
                     )
                 }
+
                 PermissionRelation.MEMBER -> {
                     permissionNodeRepository.hasMemberOrAdminPermission(
                         userId = userId,
                         entityId = entityId,
-                        entityLabel = entityType.label
+                        entityLabel = entityType.label,
                     )
                 }
             }
@@ -56,13 +56,12 @@ class Neo4jPermissionRepository(
             logger.error(e) { "Error checking direct permission for user=$userId, entity=$entityType:$entityId, relation=$relation" }
             false // Fail-closed: any error returns false
         }
-    }
 
     override fun hasTransitivePermission(
         userId: String,
         entityType: EntityType,
         entityId: String,
-        relation: PermissionRelation
+        relation: PermissionRelation,
     ): Boolean {
         return try {
             // Only check transitive permissions for namespace child entities
@@ -75,9 +74,10 @@ class Neo4jPermissionRepository(
                     permissionNodeRepository.hasAdminAccessViaNamespace(
                         userId = userId,
                         entityId = entityId,
-                        entityLabel = entityType.label
+                        entityLabel = entityType.label,
                     )
                 }
+
                 PermissionRelation.MEMBER -> {
                     if (entityType in OWNER_PRIVATE_ENTITY_TYPES) {
                         // Owner-private entities (e.g. Case, FR15): a namespace MEMBER
@@ -87,7 +87,7 @@ class Neo4jPermissionRepository(
                         permissionNodeRepository.hasAdminAccessViaNamespace(
                             userId = userId,
                             entityId = entityId,
-                            entityLabel = entityType.label
+                            entityLabel = entityType.label,
                         )
                     } else {
                         // Shared entities (AgentConfig, IntegrationConfig, AiProvider,
@@ -96,7 +96,7 @@ class Neo4jPermissionRepository(
                         permissionNodeRepository.hasReadAccessViaNamespace(
                             userId = userId,
                             entityId = entityId,
-                            entityLabel = entityType.label
+                            entityLabel = entityType.label,
                         )
                     }
                 }
@@ -111,20 +111,25 @@ class Neo4jPermissionRepository(
         userId: String,
         entityType: EntityType,
         entityId: String,
-        relation: PermissionRelation
+        relation: PermissionRelation,
     ) {
         try {
             when (relation) {
-                PermissionRelation.ADMIN -> permissionNodeRepository.createAdminPermission(
-                    userId = userId,
-                    entityId = entityId,
-                    entityLabel = entityType.label
-                )
-                PermissionRelation.MEMBER -> permissionNodeRepository.createMemberPermission(
-                    userId = userId,
-                    entityId = entityId,
-                    entityLabel = entityType.label
-                )
+                PermissionRelation.ADMIN -> {
+                    permissionNodeRepository.createAdminPermission(
+                        userId = userId,
+                        entityId = entityId,
+                        entityLabel = entityType.label,
+                    )
+                }
+
+                PermissionRelation.MEMBER -> {
+                    permissionNodeRepository.createMemberPermission(
+                        userId = userId,
+                        entityId = entityId,
+                        entityLabel = entityType.label,
+                    )
+                }
             }
             logger.info { "Granted $relation permission to user=$userId on $entityType:$entityId" }
         } catch (e: Exception) {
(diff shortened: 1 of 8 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
```diff
@@ -25,41 +25,50 @@ import java.util.UUID
...
+                logger.debug { "Platform-scope check: user=$userId, action=$action -> $granted" }
+                return granted
+            }
+
+            // 3. Cache lookup
             val cacheKey = PermissionCache.generateKey(userId, entityType, entityId, action)
             permissionCache.get(cacheKey)?.let { cached ->
                 logger.debug { "Cache hit for permission check: $cacheKey -> $cached" }
                 return cached
             }
```
vincent-audibert-whoz, line 66: Misplaced comment: `// MEMBER or ADMIN can READ` now annotates the `WRITE, DELETE` branch instead of the `READ` branch. This is a reformatting artefact — the comment should be moved above `Action.READ` or removed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
```diff
@@ -18,199 +18,230 @@ import org.springframework.data.repository.query.Param
...
 
     // Permission management queries — separate methods per relationship type
     // because Neo4j does not support parameterized relationship types.
 
-    @Query("""
-        MATCH (u:User {id: ${'$'}userId})
-        MATCH (e {id: ${'$'}entityId})
-        WHERE ${'$'}entityLabel IN labels(e)
+    @Query(
+        $$"""
```
vincent-audibert-whoz, line 96: The `$checkPlatform` parameter leaks a service-level policy decision ("this entity type supports null namespaceId") into the repository. The repository now encodes the platform-scope concept, which violates separation between query execution and authorization rules. Consider keeping both Cypher methods policy-agnostic and handling platform-id inclusion in the service layer.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
```diff
@@ -18,199 +18,230 @@ import org.springframework.data.repository.query.Param
...
+        MATCH (u:User {id: $userId})
+        MATCH (e {id: $entityId})
+        WHERE $entityLabel IN labels(e)
         MERGE (u)-[r:MEMBER]->(e)
         RETURN r
-    """)
+    """,
+    )
```
vincent-audibert-whoz, line 118: `filterIdsWhereUserIsAdmin` (used for WRITE/DELETE) has no platform-scope branch, creating an asymmetry with `filterIdsWhereUserHasAccess`. Platform entities are currently protected by the super-admin short-circuit in `filterVisibleIds`, but the two query methods are inconsistent. If the service-level bypass is ever bypassed or refactored, platform entities will silently be invisible for WRITE/DELETE even for admins. The asymmetry should be documented or resolved.

vincent-audibert-whoz (commented):
The permission model for platform-scoped entities is correctly reasoned and the critical path through `hasPermission` is sound: super-admin bypass fires first, then the `entityId == null` short-circuit cleanly separates the platform case from the namespace-bound path. The Cypher restructuring in `filterIdsWhereUserHasAccess` (single `MATCH` + compound `WHERE` instead of two `UNION` branches) is a meaningful improvement.

Two issues need to be addressed before merge.

---

**[Security gap] `filterVisibleIds` WRITE/DELETE is asymmetric with READ at the Cypher layer.**

`filterIdsWhereUserIsAdmin` (used for WRITE/DELETE via `filterVisibleIds`) was not extended with a platform-scope branch, while `filterIdsWhereUserHasAccess` (READ) was. Platform-scoped entities are visible for READ via the Cypher `$checkPlatform` branch, but for WRITE they rely entirely on the super-admin short-circuit added in `PermissionServiceImpl.filterVisibleIds`. This means:
- Today: non-admin users correctly get nothing for WRITE on platform entities (the query simply doesn't match), and super-admins are short-circuited before the query runs — so there is no active privilege escalation.
- Risk: the two code paths are now inconsistent. If `filterVisibleIds` is called from a new path that bypasses the super-admin check, or if the service-level bypass is refactored away, platform entities will silently disappear from WRITE results even for admins. The asymmetry should be resolved or explicitly documented with a comment explaining the invariant that must be preserved.

**[Architecture] `checkPlatform: Boolean` pushes a service-level policy into the repository.**

The `checkPlatform` parameter in `filterIdsWhereUserHasAccess` embeds the concept of "this entity type may have a null namespaceId" directly in the Cypher query. The repository layer now needs to know the platform-scope rule, which breaks the separation between query execution (repository) and authorization policy (service). The existing `isNamespaceChildEntity` predicate already living in `Neo4jPermissionRepository` is a pre-existing smell — this PR deepens it. Consider whether the service should handle platform-id inclusion before delegating (e.g. by first collecting platform-scoped ids separately with a dedicated query), keeping both Cypher queries policy-agnostic. At minimum, the coupling should be acknowledged with a comment explaining why the repository carries this responsibility.

---

**[Style] Misplaced comment in `evaluatePermission`'s `when` block.**

After reformatting, the comment `// MEMBER or ADMIN can READ` now sits above the `Action.WRITE, Action.DELETE` branch instead of the `Action.READ` branch (`PermissionServiceImpl.kt`). It should be moved or removed.

**[Tests] Missing DELETE coverage for platform-scope denial.**

`PermissionServiceImplSpec` tests platform-scope READ (granted) and WRITE (denied) but not DELETE. The `Action` enum has three values; DELETE is semantically distinct and should have its own assertion that it is denied for non-admins and does not hit the repository."

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
```diff
@@ -259,10 +280,15 @@ class Neo4jPermissionRepository(
...
                     } else {
+                        // Namespace-child entities may have platform-scoped instances
+                        // (namespaceId IS NULL). Use the variant query that includes a
+                        // third UNION branch for those — safe because all types in
+                        // isNamespaceChildEntity have namespaceId as a node property.
                         permissionNodeRepository.filterIdsWhereUserHasAccess(
                             userId = userId,
                             entityLabel = entityType.label,
                             ids = ids,
+                            checkPlatform = isNamespaceChildEntity(entityType),
```
vincent-audibert-whoz: ℹ️ change is here in this file

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
vincent-audibert-whoz: No, the source of this is in the service source code, the repository does not encode the platform aspect.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
vincent-audibert-whoz: Indeed, platform-level entities are only editable by super-admins and the check is done upstream, short-circuiting the repo in that case.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
```diff
@@ -219,33 +250,47 @@ interface PermissionNodeNeo4jRepository : Neo4jRepository<UserNode, String> {
...
+          )
         RETURN e.id
-    """)
+    """,
+    )
     fun filterIdsWhereUserIsAdmin(
         @Param("userId") userId: String,
         @Param("entityLabel") entityLabel: String,
         @Param("ids") ids: Collection<String>,
+        @Param("checkPlatform") checkPlatform: Boolean,
```
vincent-audibert-whoz, line 294: ℹ️ method & queries changed, the rest of the file is formatting.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
```diff
@@ -162,50 +172,59 @@ class Neo4jPermissionRepository(
...
+                    )
+                }
             }
         } catch (e: Exception) {
             logger.error(e) { "Error listing users with permission on $entityType:$entityId, relation=$relation" }
             emptyList() // Fail-closed: return empty list on error
         }
-    }
 
+    // TODO: used only for namespace, to un-generalize ?
```
vincent-audibert-whoz, line 205: ℹ️ #1037

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
```diff
@@ -25,41 +25,50 @@ import java.util.UUID
...
 
-            // 2. Cache lookup
+            // 2. Platform-scope check (entityId = null means the entity belongs to no namespace).
+            //    READ is open to any authenticated user; WRITE/DELETE are super-admin only
+            //    (already handled by the bypass above, so we deny here for non-admins).
+            if (entityId == null) {
+                val granted = action == Action.READ
+                logger.debug { "Platform-scope check: user=$userId, action=$action -> $granted" }
+                return granted
+            }
```
vincent-audibert-whoz, line 59: ℹ️ only change in file, rest is formatting.

fdelsert-whoz (approved):

PR #1039, opened 2026-06-19
feat: #WZ-32858 evict idle CaseRuntime when no SSE subscribers

vincent-audibert-whoz:
## What

Idle `CaseRuntime` instances now evict themselves from `activeRuntimes` once all SSE subscribers have disconnected, preventing unbounded memory growth in long-running deployments.

## How

- heartbeat added to SSE to detect disconnected clients
- grace delay started when no subscriber AND case is IDLE

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
@@ -4,13 +4,17 @@ import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Content
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
+import io.whozoss.agentos.caseFlow.CaseConfigProperties
 import io.whozoss.agentos.caseFlow.CaseService
 import io.whozoss.agentos.sdk.caseEvent.CaseEvent
 import io.whozoss.agentos.security.declarative.HideOnAccessDenied
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import mu.KLogging
 import org.springframework.security.access.prepost.PreAuthorize
@@ -34,7 +38,9 @@ import java.util.UUID
 class CaseEventSseController(
     private val caseService: CaseService,
     private val caseEventService: CaseEventService,
+    private val caseConfig: CaseConfigProperties = CaseConfigProperties(),
 ) {
+    private val heartbeatIntervalMs get() = caseConfig.sseHeartbeatIntervalMs
     /**
      * Stream events for a case via SSE.
      *
@@ -107,27 +113,57 @@ class CaseEventSseController(
                         }
                     }
                     emitter.complete()
+                } catch (_: CancellationException) {
+                    // Normal path: collectorJob was cancelled because the client disconnected
+                    // (onError/onCompletion fired and called scope.cancel()). Not an error.
+                    logger.debug { "SSE collector cancelled for case $caseId (client disconnected)" }
                 } catch (error: Exception) {
                     logger.error("Error in event stream for case $caseId", error)
                     emitter.completeWithError(error)
                 }
             }
 
+        // Heartbeat: periodically send an SSE comment frame so that a client
+        // disconnect is detected even when the case is IDLE and emits no events.
+        // Without this, the collector coroutine above stays suspended on collect()
+        // indefinitely, keeping the SharedFlow subscriber alive and blocking eviction.
+        val heartbeatJob =
+            scope.launch {
+                while (isActive) {
+                    delay(heartbeatIntervalMs)
+                    try {
+                        // SSE comment — ignored by EventSource but forces a socket write.
+                        emitter.send(
+                            SseEmitter
+                                .event()
+                                .comment("keep-alive"),
+                        )
+                    } catch (e: Exception) {
+                        // The socket is gone. collectorJob will be cancelled via onError/onCompletion.
+                        logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
+                        break
+                    }
+                }
+            }
+
         emitter.onCompletion {
             logger.debug { "SSE emitter completed for case $caseId" }
             collectorJob.cancel()
+            heartbeatJob.cancel()
             scope.cancel()
         }
 
         emitter.onTimeout {
             logger.debug { "SSE emitter timed out for case $caseId" }
             collectorJob.cancel()
+            heartbeatJob.cancel()
             scope.cancel()
         }
 
         emitter.onError { throwable ->
-            logger.warn { "SSE emitter error for case $caseId: ${throwable.message}" }
+            logger.debug { "SSE emitter error for case $caseId: ${throwable.message}" }
             collectorJob.cancel()
+            heartbeatJob.cancel()
             scope.cancel()
         }
 
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
@@ -38,7 +38,7 @@ import java.util.UUID
 class CaseEventSseController(
     private val caseService: CaseService,
     private val caseEventService: CaseEventService,
-    private val caseConfig: CaseConfigProperties = CaseConfigProperties(),
+    private val caseConfig: CaseConfigProperties,
 ) {
     private val heartbeatIntervalMs get() = caseConfig.sseHeartbeatIntervalMs
     /**
@@ -139,8 +139,11 @@ class CaseEventSseController(
                                 .comment("keep-alive"),
                         )
                     } catch (e: Exception) {
-                        // The socket is gone. collectorJob will be cancelled via onError/onCompletion.
+                        // The socket is gone. Cancel the scope explicitly here rather than
+                        // relying solely on Tomcat's onError callback: if that callback never
+                        // fires, collectorJob would stay subscribed indefinitely.
                         logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
+                        scope.cancel()
                         break
                     }
                 }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
@@ -11,6 +11,7 @@ import io.whozoss.agentos.security.declarative.HideOnAccessDenied
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.delay
@@ -41,6 +42,7 @@ class CaseEventSseController(
     private val caseConfig: CaseConfigProperties,
 ) {
     private val heartbeatIntervalMs get() = caseConfig.sseHeartbeatIntervalMs
+
     /**
      * Stream events for a case via SSE.
      *
@@ -113,10 +115,11 @@ class CaseEventSseController(
                         }
                     }
                     emitter.complete()
-                } catch (_: CancellationException) {
+                } catch (e: CancellationException) {
                     // Normal path: collectorJob was cancelled because the client disconnected
                     // (onError/onCompletion fired and called scope.cancel()). Not an error.
                     logger.debug { "SSE collector cancelled for case $caseId (client disconnected)" }
+                    throw e  // re-throw so cancellation propagates correctly through the coroutine hierarchy
                 } catch (error: Exception) {
                     logger.error("Error in event stream for case $caseId", error)
                     emitter.completeWithError(error)
@@ -127,52 +130,54 @@ class CaseEventSseController(
         // disconnect is detected even when the case is IDLE and emits no events.
         // Without this, the collector coroutine above stays suspended on collect()
         // indefinitely, keeping the SharedFlow subscriber alive and blocking eviction.
-        val heartbeatJob =
-            scope.launch {
-                while (isActive) {
-                    delay(heartbeatIntervalMs)
-                    try {
-                        // SSE comment — ignored by EventSource but forces a socket write.
-                        emitter.send(
-                            SseEmitter
-                                .event()
-                                .comment("keep-alive"),
-                        )
-                    } catch (e: Exception) {
-                        // The socket is gone. Cancel the scope explicitly here rather than
-                        // relying solely on Tomcat's onError callback: if that callback never
-                        // fires, collectorJob would stay subscribed indefinitely.
-                        logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
-                        scope.cancel()
-                        break
-                    }
-                }
-            }
+        startHeartbeatJob(scope, emitter, caseId)
 
         emitter.onCompletion {
             logger.debug { "SSE emitter completed for case $caseId" }
-            collectorJob.cancel()
-            heartbeatJob.cancel()
-            scope.cancel()
+            scope.cancel()  // cancels all child jobs (collectorJob, heartbeatJob)
         }
 
         emitter.onTimeout {
             logger.debug { "SSE emitter timed out for case $caseId" }
-            collectorJob.cancel()
-            heartbeatJob.cancel()
             scope.cancel()
         }
 
         emitter.onError { throwable ->
             logger.debug { "SSE emitter error for case $caseId: ${throwable.message}" }
-            collectorJob.cancel()
-            heartbeatJob.cancel()
             scope.cancel()
         }
 
         logger.info { "SSE connection established for case: $caseId" }
         return emitter
     }
+
+    private fun startHeartbeatJob(
+        scope: CoroutineScope,
+        emitter: SseEmitter,
+        caseId: UUID,
+    ): Job =
+        scope.launch {
+            while (isActive) {
+                delay(heartbeatIntervalMs)
+                try {
+                    // SSE comment — ignored by EventSource but forces a socket write.
+                    emitter.send(
+                        SseEmitter
+                            .event()
+                            .comment("keep-alive"),
+                    )
+                } catch (e: Exception) {
+                    // The socket is gone. Cancel the scope explicitly here rather than
+                    // relying solely on Tomcat's onError callback: if that callback never
+                    // fires, collectorJob would stay subscribed indefinitely.
+                    // scope.cancel() propagates to all child jobs so no explicit break is
+                    // needed — this coroutine will receive CancellationException at the
+                    // next delay() and exit the while loop naturally.
+                    logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
+                    scope.cancel()
+                }
+            }
+        }
 
     companion object : KLogging()
 }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
@@ -91,40 +91,39 @@ class CaseEventSseController(
                     .data(event),
             )
 
-        val collectorJob =
-            scope.launch {
-                try {
-                    if (includePreviousEvents == true) {
-                        // Replay persisted history first so clients connecting mid-run
-                        // or reconnecting after a disconnect receive the full sequence.
-                        caseEventService.findByParent(caseId).forEach { sendEvent(it) }
-                    }
-
-                    // If the case is still active, subscribe to the live flow.
-                    // findActiveRuntime never rehydrates — safe for observation only.
-                    // If the case is completed, the history replay above is sufficient
-                    // and the emitter completes at the end of the try block.
-                    val activeCase = caseService.findActiveRuntime(caseId)
-                    activeCase?.events?.collect { event ->
-                        try {
-                            sendEvent(event)
-                            logger.trace { "Event ${event.type} sent to SSE for case $caseId" }
-                        } catch (e: Exception) {
-                            logger.debug { "Failed to send event to SSE for case $caseId: ${e.message}" }
-                            throw e
-                        }
-                    }
-                    emitter.complete()
-                } catch (e: CancellationException) {
-                    // Normal path: collectorJob was cancelled because the client disconnected
-                    // (onError/onCompletion fired and called scope.cancel()). Not an error.
-                    logger.debug { "SSE collector cancelled for case $caseId (client disconnected)" }
-                    throw e  // re-throw so cancellation propagates correctly through the coroutine hierarchy
-                } catch (error: Exception) {
-                    logger.error("Error in event stream for case $caseId", error)
-                    emitter.completeWithError(error)
+        scope.launch {
+            try {
+                if (includePreviousEvents == true) {
+                    // Replay persisted history first so clients connecting mid-run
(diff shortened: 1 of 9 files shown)


vincent-audibert-whoz (commented):
The overall approach is sound: using `combine(subscriptionCount, statusFlow)` to gate eviction on the conjunction of IDLE + zero subscribers is the right reactive pattern, and the double-check guard before evicting is a good safeguard against races. Test coverage is thorough and well-structured. Two issues need addressing before merge, plus two smaller points.

---

**1. `statusFlow` initial value is wrong — correctness issue** (`CaseRuntime.kt`)

`_statusFlow` is hardcoded to `CaseStatus.PENDING`. For a rehydrated runtime representing a case already persisted as IDLE (e.g. a client reconnecting to view history without sending a new message), `statusFlow` will remain `PENDING` indefinitely — `run()` is never called, so the IDLE transition never happens in-memory. The watcher will never fire even though the case is idle with no subscribers, defeating the eviction mechanism for rehydrated runtimes.

The `CaseStatus` should be passed as a constructor parameter (defaulting to `PENDING` for fresh runtimes) and used to initialise `_statusFlow`. `buildRuntime` already has `case.status` available.

---

**2. One eviction-watcher coroutine leaked per `buildRuntime` call** (`CaseServiceImpl.kt`, `.also { startEvictionWatcher(case.id, it) }`)

`startEvictionWatcher` is called inside `buildRuntime`, which is invoked by both `create` and `rehydrate`. Every `getCaseRuntime` call for a case not in `activeRuntimes` triggers `rehydrate → buildRuntime → startEvictionWatcher`, spawning a new long-lived coroutine in the service scope. These coroutines are never cancelled when the runtime is subsequently evicted from `activeRuntimes` — they keep `collect`-ing until `scope.cancel()` at shutdown. In a long-running deployment where clients reconnect to many historical cases, this accumulates unboundedly, which is exactly the class of memory leak this PR is meant to prevent.

The watcher job should be cancelled when the runtime exits `activeRuntimes`. One clean approach: have `startEvictionWatcher` return the `Job`, store it alongside the runtime (e.g. a small `data class RuntimeEntry(val runtime, val watcherJob)` or a parallel `ConcurrentHashMap`), and cancel it in both the idle-eviction path and in `handleStatusChange` when a terminal status is reached.

---

**3. `requestKill` does not update `statusFlow` — KDoc is misleading** (`CaseRuntime.kt`, `requestKill()`)

The KDoc for `statusFlow` states it is "Updated synchronously inside `run` before and after each transition, **and by `requestKill` for the terminal path**". But `requestKill` only sets the atomic flags; `_statusFlow.value = CaseStatus.KILLED` is assigned inside `run()` after the loop exits. Meanwhile, `killCase()` in the service calls `handleStatusChange(KILLED)` synchronously, which evicts the runtime from `activeRuntimes` before `run()` has finished. Any concurrent read of `runtime.statusFlow.value` during this window will see the pre-kill value. Either correct the KDoc to reflect the actual behaviour, or have `requestKill` update `_statusFlow` immediately.

---

**4. Test timing fragility** (`CaseServiceImplSpec.kt` — "NOT evicted when client disconnects while agent is still running")

The assertion `delay(200) → findActiveRuntime shouldBe runtime` relies on a 300 ms margin (agent delays 500 ms, check at 200 ms). This can flake under CI load. The runtime's survival can be asserted *immediately* after `shortLivedJob.join()` — at that point, status is still RUNNING and the grace period has not started, so no timing margin is needed at all. The `delay(200)` assertion adds fragility without adding coverage.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -131,6 +148,68 @@ class CaseServiceImpl(
...
+                    count == 0 && status == CaseStatus.IDLE
+                }.filter { it }
+                    .collect {
+                        // Grace period: absorb brief reconnects and fast follow-up messages.
+                        delay(idleEvictionGraceMs)
+
+                        // Re-check: a subscriber may have reconnected or a new message may
+                        // have arrived (status back to RUNNING) during the grace period.
+                        if (activeRuntimes.containsKey(caseId) &&
+                            runtime.subscriptionCount.value == 0 &&
```
vincent-audibert-whoz, line 193: **Blocking: watcher self-cancellation is semantically wrong.**

```kotlin
activeRuntimes.remove(caseId)
cancelEvictionWatcher(caseId)   // calls job.cancel() on the currently-running coroutine
```

`cancelEvictionWatcher` does `watcherJobs.remove(caseId)?.cancel()`. At this point `watcherJobs[caseId]` is the job that is currently executing this `collect` block. Calling `cancel()` on yourself from inside your own body is a no-op for cooperative cancellation — the coroutine will simply return naturally. The call is misleading and could mask bugs if the logic is refactored.

Correct pattern: just `activeRuntimes.remove(caseId)` and `watcherJobs.remove(caseId)` (without `.cancel()`), then let the coroutine fall off the end. Or register `job.invokeOnCompletion { watcherJobs.remove(caseId) }` at creation time so cleanup is automatic.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -42,7 +46,10 @@ class CaseServiceImpl(
     private val caseEventService: CaseEventService,
```
vincent-audibert-whoz, line 46: **Blocking: default constructor value bypasses Spring fail-fast.**

```kotlin
private val caseConfig: CaseConfigProperties = CaseConfigProperties(),
```

If `@EnableConfigurationProperties(CaseConfigProperties::class)` is ever removed from `AgentOSApplication`, Spring will silently use the hardcoded default instead of failing at startup. For a `@Service` that controls eviction timing, this is a silent misconfiguration risk. Remove the default value — tests already pass explicit instances.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -34,7 +38,9 @@ import java.util.UUID
 class CaseEventSseController(
     private val caseService: CaseService,
     private val caseEventService: CaseEventService,
```
vincent-audibert-whoz, line 41: **Blocking (same issue): default constructor value bypasses Spring fail-fast.**

```kotlin
private val caseConfig: CaseConfigProperties = CaseConfigProperties(),
```

Same problem as `CaseServiceImpl`. A misconfigured application context silently uses a 30-second heartbeat interval rather than failing at startup. Remove the default; tests already pass `CaseConfigProperties(sseHeartbeatIntervalMs = Long.MAX_VALUE)` explicitly.

vincent-audibert-whoz (commented):
The eviction mechanism is well-designed overall: the `combine(subscriptionCount, statusFlow)` approach correctly avoids the race where a client disconnects while the agent is still running, and the grace period with a re-check guard handles the fast-follow-up scenario cleanly. The test suite is thorough and the documentation in `CaseConfigProperties` is exemplary. There are however two issues that need addressing before merge, plus a few non-blocking observations.

**Blocking: watcher self-cancellation creates a semantically wrong pattern.** In `startEvictionWatcher`, when eviction fires, the watcher calls `cancelEvictionWatcher(caseId)` on itself — removing and cancelling its own `Job` from within the coroutine body. `cancelEvictionWatcher` does `watcherJobs.remove(caseId)?.cancel()`, which calls `cancel()` on the job that is currently executing this code. In practice this is a no-op (the coroutine will complete naturally after the `collect` block), but it is semantically wrong and misleading: it suggests the coroutine is being forcibly stopped when it is in fact just finishing. The correct pattern is to simply remove the map entry (`watcherJobs.remove(caseId)`) without calling `cancel()`, and let the coroutine return. Alternatively, register `Job.invokeOnCompletion { watcherJobs.remove(caseId) }` at creation time and remove the manual cleanup entirely.

**Blocking: default constructor values for `CaseConfigProperties` bypass Spring's fail-fast behaviour.** Both `CaseServiceImpl` and `CaseEventSseController` declare `private val caseConfig: CaseConfigProperties = CaseConfigProperties()`. Spring injects the `@ConfigurationProperties` bean in production, so the default is never used — but if the bean is missing (e.g. `@EnableConfigurationProperties` accidentally removed from `AgentOSApplication`), the application starts silently with hardcoded defaults rather than failing at startup. For production Spring `@Service`/`@RestController` constructors, mandatory parameters with no defaults are the correct pattern. Tests already pass explicit instances (`CaseConfigProperties(idleEvictionGraceMs = 50L)`), so removing the defaults has no test impact.

**Non-blocking: `_statusFlow` is updated before `updateStatus` (the persistence callback) at every transition site in `CaseRuntime.run()`.** This is operationally fine — the grace period absorbs the latency — but it creates a brief window where `statusFlow.value == IDLE` while the database still shows `RUNNING`. A short inline comment at each transition site explaining the intentional ordering would prevent a future reader from "fixing" it to the wrong order.

**Non-blocking: the "idle runtime NOT evicted when new message arrives" test uses `delay(100)` inside a 500ms grace window.** This 5x margin is usually sufficient but is an implicit timing assumption. On a heavily loaded CI machine, asserting immediately after `secondIdle.join()` and `awaitNotRunning(runtime)` would be safer: at that point the second grace period has just started and no eviction could have fired yet.

**Non-blocking: the heartbeat `break` path relies implicitly on Tomcat calling `onError` after a failed write.** If that callback does not fire (mock emitter in tests, different Servlet container), `collectorJob` stays subscribed indefinitely, defeating the eviction mechanism. Explicitly cancelling `scope` (or `collectorJob`) after `break` would make the disconnect path self-contained and not depend on the container's callback chain.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
```diff
@@ -228,6 +244,7 @@ class CaseRuntime(
         logger.info { "[CaseRuntime $id] run() started" }
```
vincent-audibert-whoz, line 252: **Non-blocking: `_statusFlow` is updated before `updateStatus` — document the intent.**

```kotlin
_statusFlow.value = CaseStatus.RUNNING
updateStatus(id, CaseStatus.RUNNING)  // persistence callback, called after
```

The reactive flow is updated before persistence at every transition site. This is fine operationally (the eviction watcher's grace period absorbs the latency), but it creates an observable window where `statusFlow.value == IDLE` while the database still shows `RUNNING`. A short inline comment explaining the intentional ordering would prevent a future reader from inverting it.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
```diff
@@ -1123,6 +1126,172 @@ class CaseServiceImplSpec :
...
+            secondIdle.join()
+
+            // Check BEFORE the second grace period elapses (delay << graceMs=500).
+            // The runtime must still be alive: the first eviction was cancelled by the
+            // second message, and the second eviction hasn't fired yet.
+            delay(100)
+
+            service.findActiveRuntime(case.id) shouldBe runtime
+            service.getById(case.id).status shouldBe CaseStatus.IDLE
+        }
```
vincent-audibert-whoz, line 1271: **Non-blocking: timing-dependent assertion without explicit justification.**

```kotlin
delay(100)
service.findActiveRuntime(case.id) shouldBe runtime
```

Asserts the runtime is alive 100ms into a 500ms grace period. On a heavily loaded CI machine (GC pauses, scheduler starvation) this 5x margin may not hold. Asserting immediately after `secondIdle.join()` + `awaitNotRunning(runtime)` would be safer: at that point the second grace period has just started and no eviction could have fired yet.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,57 @@ class CaseEventSseController(
...
         emitter.onCompletion {
             logger.debug { "SSE emitter completed for case $caseId" }
             collectorJob.cancel()
+            heartbeatJob.cancel()
             scope.cancel()
         }
 
         emitter.onTimeout {
             logger.debug { "SSE emitter timed out for case $caseId" }
             collectorJob.cancel()
```
vincent-audibert-whoz: **Non-blocking: implicit assumption about Tomcat's `onError` callback after a failed write.**

```kotlin
} catch (e: Exception) {
    logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
    break
}
```

After `break`, `collectorJob` remains subscribed to the `SharedFlow`. The comment says Tomcat fires `onError`/`onCompletion`, which then cancels `collectorJob`. This is correct for Tomcat's `AsyncContext` lifecycle, but is an implicit container assumption. If it does not hold (mock emitter in tests, different container), `collectorJob` stays alive indefinitely — defeating the eviction mechanism. Explicitly cancelling `scope` after `break` would make the disconnect path self-contained.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
```diff
@@ -244,28 +268,34 @@ class CaseRuntime(
...
+                    updateStatus(CaseStatus.IDLE)
                 }
             }
         } catch (e: Exception) {
             logger.error(e) { "[CaseRuntime $id] Error during execution" }
-            updateStatus(id, CaseStatus.ERROR)
+            updateStatus(CaseStatus.ERROR)
         } finally {
             runInFlight.set(false)
         }
```
vincent-audibert-whoz, line 289: **Residual observation (non-blocking): `updateStatus` helper is non-`suspend`.**

```kotlin
private fun updateStatus(caseStatus: CaseStatus) {
    _statusFlow.value = caseStatus
    updateStatusCallback(id, caseStatus)
}
```

`updateStatusCallback` is currently `(UUID, CaseStatus) -> Unit` (non-suspend), so calling it synchronously from a `suspend` context is fine. If `updateStatusCallback` is ever changed to a `suspend` lambda, this helper will need to be marked `suspend` as well. No action needed now — just a forward-compatibility note.

vincent-audibert-whoz (commented):
Follow-up review after the round of fixes. All four blocking and non-blocking issues from the initial review are now correctly addressed. No new bugs were introduced by the refactoring. One residual observation worth noting.

**All issues resolved:**

- **Watcher self-cancellation** (`CaseServiceImpl`): fixed correctly. On eviction the code now calls `watcherJobs.remove(caseId)` without `.cancel()`, with an inline comment explaining exactly why. The coroutine falls off the end naturally. ✅

- **Default constructor values** (`CaseServiceImpl` and `CaseEventSseController`): both defaults have been removed. Both constructors are now mandatory-parameter, fail-fast. ✅

- **`_statusFlow` ordering** (`CaseRuntime`): the four separate update sites have been consolidated into a single private `updateStatus(CaseStatus)` helper that always updates `_statusFlow` first, then calls `updateStatusCallback`. The helper carries an inline comment explaining the intentional ordering. The KDoc on the constructor parameter was also updated to `updateStatusCallback`. ✅

- **Timing-dependent test assertion** (`CaseServiceImplSpec`): the `delay(100)` has been replaced with `awaitNotRunning(runtime)`, which polls `isRunning()` without any wall-clock assumption. ✅

- **Heartbeat `break` path** (`CaseEventSseController`): `scope.cancel()` is now called explicitly before `break`, with a comment explaining that this is a defensive measure against containers that may not fire `onError`. ✅

**Residual observation (non-blocking):**

The `updateStatus` private helper in `CaseRuntime` is a non-`suspend` `fun` that calls `updateStatusCallback`, which itself is typed as `(UUID, CaseStatus) -> Unit` (non-suspend). The existing callers inside `run()` — which is `suspend` — call it synchronously. This is fine for the current implementation where `updateStatusCallback` delegates to `handleStatusChange`, a regular blocking function. Worth keeping in mind if `updateStatusCallback` is ever made suspending: the helper would need to be marked `suspend` at that point, but there is nothing to fix now.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,60 @@ class CaseEventSseController(
...
                     emitter.complete()
+                } catch (_: CancellationException) {
+                    // Normal path: collectorJob was cancelled because the client disconnected
+                    // (onError/onCompletion fired and called scope.cancel()). Not an error.
+                    logger.debug { "SSE collector cancelled for case $caseId (client disconnected)" }
                 } catch (error: Exception) {
                     logger.error("Error in event stream for case $caseId", error)
                     emitter.completeWithError(error)
                 }
             }
```
thomasmartin-whoz, line 158: dedicated method

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,60 @@ class CaseEventSseController(
...
+                    } catch (e: Exception) {
+                        // The socket is gone. Cancel the scope explicitly here rather than
+                        // relying solely on Tomcat's onError callback: if that callback never
+                        // fires, collectorJob would stay subscribed indefinitely.
+                        logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
+                        scope.cancel()
+                        break
+                    }
+                }
+            }
```
thomasmartin-whoz: dedicated method

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,60 @@ class CaseEventSseController(
...
+                                .event()
+                                .comment("keep-alive"),
+                        )
+                    } catch (e: Exception) {
+                        // The socket is gone. Cancel the scope explicitly here rather than
+                        // relying solely on Tomcat's onError callback: if that callback never
+                        // fires, collectorJob would stay subscribed indefinitely.
+                        logger.debug { "Heartbeat write failed for case $caseId — client likely disconnected" }
+                        scope.cancel()
+                        break
```
thomasmartin-whoz: pourquoi ce break ? le `scope.cancel()` devrait cancel cette coroutine non ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,60 @@ class CaseEventSseController(
...
+                    }
+                }
+            }
+
         emitter.onCompletion {
             logger.debug { "SSE emitter completed for case $caseId" }
             collectorJob.cancel()
+            heartbeatJob.cancel()
```
thomasmartin-whoz: `scope.cancel()` doit cancel ses jobs donc             collectorJob.cancel() et             heartbeatJob.cancel() sont inutiles je pense

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
```diff
@@ -325,7 +360,12 @@ class CaseRuntime(
                         interruptRequested.set(true)
                         return
                     }
-                    runAgent(event.agentName, eventList.getAll(), { eventList.getAll() }, userId) { !interruptRequested.get() }
+                    runAgent(
```
thomasmartin-whoz, line 363: named params

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -131,6 +148,66 @@ class CaseServiceImpl(
...
+     * On each transition to `(IDLE, 0)`, the watcher:
+     * 1. Waits [idleEvictionGraceMs] to absorb brief reconnects (e.g. page refresh,
+     *    or a user typing a quick follow-up).
+     * 2. Re-checks the combined condition — if a subscriber reconnected or a new
+     *    message arrived (status back to RUNNING) during the grace period, eviction
+     *    is skipped.
+     * 3. If both conditions still hold, evicts the runtime.
+     *
+     * The coroutine runs for the lifetime of the service scope. It is automatically
+     * cancelled when [scope] is cancelled (service shutdown).
```
thomasmartin-whoz, line 174: this confirms the scope.cancel() propagation to jobs I talked about earlier

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -107,27 +113,60 @@ class CaseEventSseController(
                         }
                     }
                     emitter.complete()
+                } catch (_: CancellationException) {
```
thomasmartin-whoz: Catching CancellationException can break the cancellation propagation. If you must catch it, rethrow it to let the cancellation propagate correctly through the coroutine hierarchy.

https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-propagation

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -131,6 +148,66 @@ class CaseServiceImpl(
...
+    ) {
+        logger.trace { "Case $caseId eviction watcher started" }
+        val job =
+            scope.launch {
+                combine(runtime.subscriptionCount, runtime.statusFlow) { count, status ->
+                    count == 0 && status == CaseStatus.IDLE
+                }.filter { it }
+                    .collect {
+                        // Grace period: absorb brief reconnects and fast follow-up messages.
+                        delay(idleEvictionGraceMs)
```
fdelsert-whoz: à remplacer par un `debounce` avant le `filter` ?

```
.distinctUntilChanged()
.debounce(idleEvictionGraceMs)
.filter { it }
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
vincent-audibert-whoz: I don’t get it ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
vincent-audibert-whoz: removed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
vincent-audibert-whoz: removed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
vincent-audibert-whoz: Cannot name params because it is a lambda.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
vincent-audibert-whoz: re-throwed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
thomasmartin-whoz: all this collectorJob stuff should be in a dedicated method

thomasmartin-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
thomasmartin-whoz: all the collectorJob handling should be in a dedicated method, just like heartbeatJob

vincent-audibert-whoz: method extracted

thomasmartin-whoz (approved):

PR #1051, opened 2026-06-25
feat: agent delegation via sub-cases

vincent-audibert-whoz:
## What

Introduces agent-to-agent delegation in AgentOS: an agent configured with `subAgents` can delegate tasks to specialized sub-agents by spawning child cases and waiting for their result.

### New: `delegation/` package
- **`DelegationTool`** — `StandardTool` that creates a sub-case, waits for it to reach `IDLE` (or a terminal status), extracts the last agent `MessageEvent`, and returns it as the tool result. Supports configurable timeout with automatic sub-case kill on expiry.
- **`SubCaseLauncher`** — minimal interface extracted from `CaseService` to break the circular dependency `CaseService → AgentService → CaseService`.

### `AgentConfig`
- New `subAgents: List<String>` field — allowlist of agent names this agent may delegate to.
- Persisted on `AgentConfigNode`, exposed in `AgentConfigResource`.

### `AgentServiceImpl`
- When `subAgents` is non-empty, a `DelegationTool` instance is injected into the agent's tool list, scoped to the current case and namespace.

### `Case` / `CaseService`
- New `parentCaseId: UUID?` field — links a sub-case to its parent.
- `CaseServiceImpl` implements `SubCaseLauncher`: `startSubCase` creates a child case, injects the task as the first user message, and returns its live `CaseRuntime`. `killSubCase` delegates to the existing kill flow.

### Tests
- **`DelegationToolUnitSpec`** — 10 unit tests covering allowlist enforcement, nominal path (IDLE), multi-part messages, terminal statuses (ERROR, KILLED), and timeout with kill verification.
- **`AgentServiceImplUnitSpec`** — extended to cover `DelegationTool` injection when `subAgents` is configured.


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +602,61 @@ class CaseServiceImpl(
         runtime.requestInterrupt()
     }
 
+    override fun killSubCase(caseId: UUID) = killCase(caseId)
+
     override fun killCase(caseId: UUID) {
```
vincent-audibert-whoz: **Blocking — potential StackOverflowError on deep/cyclic delegation chains.**

```kotlin
caseRepository.findActiveByParentCaseId(caseId).forEach { subCase ->
    killCase(subCase.id)  // recursive call
}
```

`killCase` calls itself synchronously for each sub-case, and each of those calls does the same. With N levels of delegation the JVM call stack grows N frames deep. If a cycle exists in the `parentCaseId` graph (possible via direct DB manipulation or a bug), this is an infinite recursion. Even without cycles, a few hundred levels is enough to overflow the default JVM stack.

Recommend switching to an iterative BFS/DFS with an explicit queue and a `visitedIds: MutableSet<UUID>` guard to handle both depth and cycles safely.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +602,61 @@ class CaseServiceImpl(
...
+        namespaceId: UUID,
+        agentName: String,
+        task: String,
+        userId: UUID,
+    ): CaseRuntime {
+        val user =
+            userService.findById(userId)
+                ?: throw ResourceNotFoundException("User not found: $userId")
+        val displayName =
+            listOfNotNull(user.firstname, user.lastname)
```
vincent-audibert-whoz: **Minor — redundant `getCaseRuntime` call.**

```kotlin
val subCase = create(Case(...))
val runtime = getCaseRuntime(subCase.id)  // <-- redundant
```

`create()` already stores the runtime in `activeRuntimes`. `getCaseRuntime` then calls `computeIfAbsent`, which finds the key already present and returns the existing instance — no harm done, but it reads as if the runtime might not exist yet and obscures intent. Prefer a direct internal accessor or restructure so the runtime is returned from `create` directly.

vincent-audibert-whoz (commented):
The delegation feature is architecturally well thought-out: the `SubCaseLauncher` seam breaks the circular dependency cleanly, the glob-based allowlist resolution is correctly scoped per user/namespace at build time, the kill-propagation is depth-first and handles arbitrary chains, and the test suite covers all the meaningful paths. The implementation is production-quality in most respects.

There are issues that should be addressed before merge, ranked by severity.

**Blocking — correctness:**
1. **Recursive kill can stack-overflow on deep delegation chains.** `killCase` calls itself recursively depth-first without any cycle detection or depth limit. A misconfigured or malicious delegation chain (A→B→A, or simply A→B→C→…→Z with 100+ levels) will produce a `StackOverflowError` on the calling thread. The depth of the recursion is bounded by the database query result, not by the JVM stack.
2. **`startSubCase` uses fully-qualified `io.whozoss.agentos.sdk.actor.Actor` and `io.whozoss.agentos.sdk.actor.ActorRole.USER` inline** when both types are importable from the same package already used in the file. This is an unnecessary FQN that contradicts the Kotlin convention and the rest of the file's style.

**Design concern (not a hard blocker):**
3. **`startSubCase` is declared on `CaseService`** (the public interface), but it is logically only needed by `SubCaseLauncher` consumers. Exposing it on the broad `CaseService` interface means any injector of `CaseService` can call it, bypassing the allowlist enforced in `DelegationTool`. Since `CaseServiceImpl` already implements `SubCaseLauncher`, `DelegationTool` only needs `SubCaseLauncher` — it never needs `CaseService`. Consider removing `startSubCase` from `CaseService` and keeping it exclusively on `SubCaseLauncher`.

**Minor observations (no action required, noted for awareness):**
- `extractLastAgentMessage` (the DB read via `loadCaseEvents`) is called *outside* the `withTimeout` block, so a slow event store could block the parent agent's coroutine indefinitely after the sub-case has completed.
- The `inputSchema` is hand-rolled JSON with string interpolation. Agent names containing `"`, `\`, or other JSON-special characters will produce malformed JSON. Risk is low today since names are validated upstream, but worth noting.
- The redundant `getCaseRuntime(subCase.id)` call inside `startSubCase` — `create()` already stores the runtime in `activeRuntimes`, so `computeIfAbsent` is a no-op. No harm, but it reads as if the runtime might not exist yet.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +602,61 @@ class CaseServiceImpl(
...
+            logger.info { "Killing sub-case ${subCase.id} (parent: $caseId)" }
+            killCase(subCase.id)
+        }
         // Signal the runtime loop to exit cleanly if it is currently running,
         // then let handleStatusChange evict it via the isTerminal() path.
         activeRuntimes[caseId]?.requestKill()
         handleStatusChange(caseId, CaseStatus.KILLED)
     }
 
+    override fun startSubCase(
```
vincent-audibert-whoz, line 682: **Style — unnecessary fully-qualified class names.**

```kotlin
val actor =
    io.whozoss.agentos.sdk.actor.Actor(
        ...
        role = io.whozoss.agentos.sdk.actor.ActorRole.USER,
    )
```

`Actor` and `ActorRole` are importable and used with short names elsewhere in the codebase. Replace with imported short names — this is the only style deviation in an otherwise clean diff.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseService.kt
```diff
@@ -140,4 +140,35 @@ interface CaseService : EntityService<Case, UUID> {
...
+     * @param userId       The user on whose behalf the sub-case runs.
+     */
+    fun startSubCase(
+        parentCaseId: UUID,
+        namespaceId: UUID,
+        agentName: String,
+        task: String,
+        userId: UUID,
+    ): CaseRuntime
 }
```
vincent-audibert-whoz, line 143: **Design concern — `startSubCase` on the public `CaseService` interface.**

This operation is delegation-internal; its safety depends on the `allowedAgents` allowlist enforced in `DelegationTool`. Placing it on `CaseService` exposes it to every bean that injects `CaseService`, making it easy for future callers to bypass the allowlist. Since `DelegationTool` only ever receives a `SubCaseLauncher`, removing `startSubCase` from `CaseService` would tighten the contract without losing any capability.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,188 @@
...
+    private val loadCaseEvents: (UUID) -> List<CaseEvent>,
+    private val timeoutMs: Long = 5 * 60 * 1_000L,
+) : StandardTool<DelegationTool.Args> {
+
+    data class Args(
+        val agentName: String,
+        val task: String,
+    )
+
+    override val name = "DELEGATE__delegate"
```
vincent-audibert-whoz, line 83: **Latent bug — hand-rolled JSON schema with string interpolation.**

```kotlin
val names = allowedAgents.joinToString(", ") { "\"$it\"" }
append("""{ ... "enum": [$names] ... }""")
```

Agent names containing `"`, `\`, or other JSON-special characters will silently produce malformed JSON. Risk is low today (names are validated upstream), but using Jackson to build the schema object would eliminate the class of bugs entirely.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,188 @@
...
+                    "Agent '${input.agentName}' is not in the delegation allowlist. " +
+                        "Allowed agents: ${allowedAgents.joinToString(", ")}.",
+                success = false,
+            )
+        }
+        val userId =
+            context.userId
+                ?: return ToolExecutionResult(
+                    output = "Delegation requires a user context (userId is null).",
+                    success = false,
```
vincent-audibert-whoz: **Observation — `extractLastAgentMessage` runs outside the timeout.**

```kotlin
val finalStatus = withTimeout(timeoutMs) {
    subRuntime.statusFlow.filter { ... }.first()
}
// DB read happens here, outside the timeout:
val result = extractLastAgentMessage(subCaseId)
```

The timeout only guards the status observation. If the event store is slow, the tool can block the parent agent's coroutine indefinitely after the sub-case has already completed. Not a problem today, but worth knowing if `loadCaseEvents` ever becomes async or expensive.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +603,61 @@ class CaseServiceImpl(
         runtime.requestInterrupt()
     }
 
+    override fun killSubCase(caseId: UUID) = killCase(caseId)
+
```
vincent-audibert-whoz, line 656: **Still blocking — recursive `killCase` StackOverflow risk unchanged.**

This code is identical to the previous commit:
```kotlin
caseRepository.findActiveByParentCaseId(caseId).forEach { subCase ->
    killCase(subCase.id)  // synchronous recursion, no depth limit, no cycle guard
}
```

Commit `828bbfa1` addressed the other review items but did not touch this method. The risk is the same as described in the first review: a delegation chain deep enough to exhaust the JVM stack (or a cycle in `parentCaseId` due to data corruption) will throw `StackOverflowError`. An iterative approach with an explicit queue and a `visitedIds: MutableSet<UUID>` guard is the correct fix.

vincent-audibert-whoz (commented):
Follow-up review of commit `828bbfa1` — addressing the previous round's findings.

**All previous issues resolved ✓**

- FQN `io.whozoss.agentos.sdk.actor.Actor` / `ActorRole.USER` replaced with proper imports — confirmed clean.
- `startSubCase` removed from `CaseService` interface — `CaseService.kt` no longer declares it; it lives exclusively on `SubCaseLauncher`. Contract boundary is correct.
- Redundant `getCaseRuntime(subCase.id)` replaced with `activeRuntimes[subCase.id]!!` — direct map access, intent is now explicit.
- `DelegationTool` success output now includes `subCaseId` alongside `result` in the JSON — confirmed in the source.

**One new issue introduced — correctness:**

The recursive `killCase` StackOverflow concern flagged in the first review is **still present**. Commit `828bbfa1` did not touch `killCase`. The recursive depth-first propagation through `caseRepository.findActiveByParentCaseId(caseId).forEach { killCase(it.id) }` remains unchanged, with no cycle guard and no depth limit. This was the only blocking issue from the previous review and it was not addressed.

No new issues were introduced by this commit. The PR is otherwise ready to merge once the kill-propagation recursion is made safe.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: Case hierarchy will be relatively shallow and not very wide, point ignored.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: Case hierarchy will remain very far from reaching StackOverflowError, point ignored.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,188 @@
...
+                withTimeout(timeoutMs) {
+                    subRuntime.statusFlow
+                        .filter { it == CaseStatus.IDLE || it.isTerminal() }
+                        .first()
+                }
+
+            when {
+                finalStatus.isTerminal() && finalStatus != CaseStatus.IDLE -> {
+                    logger.warn { "[DelegationTool] Sub-case $subCaseId ended with terminal status $finalStatus" }
+                    ToolExecutionResult(
```
vincent-audibert-whoz: **Design note — ambiguous timeout error when `EVENT_LOAD_TIMEOUT_MS` fires.**

The outer `catch (e: TimeoutCancellationException)` block handles both the 5-min delegation timeout and the 10-s event-load timeout identically: it calls `killSubCase` and reports "timed out after ${timeoutMs / 1000}s". If the event store is the slow party, the sub-case is already done — killing it is a no-op at best, misleading at worst — and the reported duration is wrong.

A comment here explaining that the catch covers both timeout origins would prevent future confusion. Low-priority.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/delegation/DelegationToolUnitSpec.kt
```diff
@@ -0,0 +1,223 @@
...
+        every { launcher.startSubCase(any(), any(), any(), any(), any()) } returns runtime
+
+        val result = tool.execute(DelegationTool.Args(agentName = "sub-agent", task = "do X"), toolContext)
+
+        result.success shouldBe false
+        result.output shouldBe "Delegation to 'sub-agent' timed out after 0s. The sub-agent did not complete in time."
+        result.metadata.containsKey("subCaseId") shouldBe true
+        verify(exactly = 1) { launcher.killSubCase(subCaseId) }
+    }
+})
```
vincent-audibert-whoz: **Missing test for `EVENT_LOAD_TIMEOUT_MS` path.**

The new `withTimeout(EVENT_LOAD_TIMEOUT_MS)` guard has no corresponding test. A test supplying a `loadCaseEvents` lambda that suspends indefinitely (e.g. `loadCaseEvents = { kotlinx.coroutines.delay(Long.MAX_VALUE); emptyList() }`) would verify the guard fires and returns a failure rather than hanging. Not a blocker, but the protection is currently untested.

vincent-audibert-whoz (commented):
Follow-up review of commit `2e7d00c4` — the two minor observations from the previous round on `DelegationTool.kt`.

**Both observations addressed correctly ✓**

**1. Jackson-built `inputSchema` (JSON-injection fix)**
The hand-rolled string interpolation is gone. The schema is now built entirely via `objectMapper.createObjectNode()` with the `allowedAgents` list added through `putArray("enum").also { arr -> allowedAgents.forEach(arr::add) }`. Jackson serialises every agent name safely regardless of content. No issues.

One minor observation: `inputSchema` is computed at construction time as a property initialiser, which means `objectMapper` (a companion object field) is accessed during instance initialisation. This is safe in Kotlin — companion object fields are initialised before class body — but worth knowing if the tool ever becomes lazy-initialised. No action needed.

**2. `EVENT_LOAD_TIMEOUT_MS` wrapping `extractLastAgentMessage`**
The `withTimeout(EVENT_LOAD_TIMEOUT_MS)` call correctly wraps the `loadCaseEvents` invocation inside the `else` branch. The 10-second constant is declared as a `private const val` in the companion object, named clearly. The implementation is correct.

One design note (not a blocker): a `TimeoutCancellationException` thrown by the inner `EVENT_LOAD_TIMEOUT_MS` timeout is caught by the outer `catch (e: TimeoutCancellationException)` block, which then calls `killSubCase` and reports "timed out after 5 min". If the event store is slow, this produces a misleading error message (wrong duration) and kills a sub-case that already completed successfully. This is an acceptable edge-case trade-off, but a comment near the catch block explaining the dual-timeout semantics would help future readers.

**Test coverage gap**
The `DelegationToolUnitSpec` was not updated for the `EVENT_LOAD_TIMEOUT_MS` path. There is no test that injects a blocking `loadCaseEvents` lambda to verify the inner timeout fires and produces a failure result rather than a hang. Not a blocker given the edge-case nature, but the protection added in this commit is untested.

**Remaining blocker unchanged**
The recursive `killCase` StackOverflow risk in `CaseServiceImpl` is still present and unaddressed. The PR should not be merged until that is resolved.

alizarion (commented):
(Superseded: these comments were reposted under the correct account.)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -419,5 +427,57 @@ class AgentServiceImpl(
...
+        context: AgentExecutionContext,
+        subCaseLauncher: SubCaseLauncher?,
+    ): List<DelegationTool> {
+        val patterns = config.subAgents?.filter { it.isNotBlank() }
+        if (subCaseLauncher == null || patterns.isNullOrEmpty() || context.caseId == null) return emptyList()
+
+        val accessibleNames =
+            if (context.userId != null) {
+                agentConfigService
+                    .findAvailableByNamespaceIdAndUserId(context.namespaceId, context.userId)
```
selimbensenouciep-prog: **Critical: no self / cycle / depth guard on delegation.** The candidate list is built here but the calling agent (config.name) is never excluded, so an agent whose subAgents includes `*` (or its own name) can delegate to itself. Sub-cases run through the normal runAgent path, so a sub-agent is also granted a DelegationTool, and startSubCase (CaseServiceImpl) adds no depth counter or ancestor / cycle check. A `*`-delegating agent, or a mutual A to B to A pair, then spawns sub-cases without bound (each blocking up to timeoutMs), exhausting the IO dispatcher and memory. Suggest excluding config.name here and enforcing a max-depth plus ancestor-set check in startSubCase.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,189 @@
...
+                userId = userId,
+            )
+        val subCaseId = subRuntime.id
+        logger.info { "[DelegationTool] Sub-case $subCaseId started, waiting for completion (timeout=${timeoutMs}ms)" }
+
+        return try {
+            val finalStatus =
+                withTimeout(timeoutMs) {
+                    subRuntime.statusFlow
+                        .filter { it == CaseStatus.IDLE || it.isTerminal() }
```
selimbensenouciep-prog, line 257: **Critical: a sub-agent that asks a question is reported as success.** The wait is filter { it == IDLE || it.isTerminal() }, but IDLE is not terminal (only KILLED and ERROR are), and IDLE is exactly the state a sub-agent enters after emitting a QuestionEvent (turn complete, waiting for the next user message). So a question-blocked sub-case falls into the success branch: the parent receives the question text as if it were the answer, and the sub-case is never killed, lingering as an idle orphan. This is the opposite of the documented 'fails and the sub-case is killed' behavior. Suggest treating IDLE-with-a-pending-question as a failure (and killing it), or modeling an explicit completion signal.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +603,61 @@ class CaseServiceImpl(
...
+            Actor(
+                id = userId.toString(),
+                displayName = displayName,
+                role = ActorRole.USER,
+            )
+        // Use @mention syntax so the normal selectAgent resolution picks up the
+        // requested agent without any special-casing in the runtime.
+        val mentionedTask = "@$agentName $task"
+        val subCase =
+            create(
```
selimbensenouciep-prog, line 700: **Critical: the sub-case is orphaned from the permission graph.** This calls the service-level create(...), which persists the node and the namespace edge but grants the creator no permission. User-facing cases get an ADMIN grant via CaseController.create; that step is skipped here. Listing (findAccessibleByUserInNamespace) and getById / events / SSE all require a direct ADMIN or MEMBER edge, so the delegating user can never list, open, or stream the sub-case they spawned, even though parentCaseId is now exposed in the API and UI. Suggest granting the creator (or the parent case owner) a permission on the sub-case.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
```diff
@@ -72,4 +72,16 @@ data class AgentConfig(
...
+     *
+     * When null or empty, no delegation tool is provided to the agent.
+     * When non-empty, a [io.whozoss.agentos.delegation.DelegationTool] is instantiated
+     * and added to the agent's tool set at build time, with the allowlist resolved by
+     * matching these patterns against agents accessible to the current user in the namespace.
+     *
+     * `*` matches any sequence of characters (anchored, case-insensitive).
+     * Examples: `["*"]` allows all agents, `["*Fixer"]` allows `BugFixer`, `StoryFixer`, etc.
+     */
+    val subAgents: List<String>? = null,
```
selimbensenouciep-prog, line 89: **Major (design): this duplicates the existing REDIRECT capability one layer lower.** subAgents is a new domain field threaded through Node, Resource, Controller, OpenAPI, the TS client, and a bespoke UI editor. The existing REDIRECT ToolPlugin already does agent-to-agent routing (a glob target list, per agent, admin-controlled) with zero new AgentConfig fields, by reusing the integrations mechanism plus an IntegrationConfig. Likewise buildDelegationTools hand-rolls glob and accessibility resolution outside resolveToolsForRun, and subCaseLauncher is threaded through findAgentByName where REDIRECT breaks the same service cycle with a @Bean resolver. Modeling delegation as a DELEGATE ToolPlugin would inherit tool scoping, metrics, allowlist, and config validation instead of forking a second path. Worth discussing before this hardens.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,189 @@
...
+                    logger.warn { "[DelegationTool] Sub-case $subCaseId ended with terminal status $finalStatus" }
+                    ToolExecutionResult(
+                        output = "Sub-case ended with status $finalStatus without producing a result.",
+                        success = false,
+                        metadata = mapOf("subCaseId" to subCaseId.toString()),
+                    )
+                }
+
+                else -> {
+                    val result = withTimeout(eventLoadTimeoutMs) { extractLastAgentMessage(subCaseId) }
```
selimbensenouciep-prog: **Major: the event-load timeout is caught by the run-timeout handler.** withTimeout(eventLoadTimeoutMs) { extractLastAgentMessage(...) } runs inside the same try whose only catch (TimeoutCancellationException) assumes the operation timeout fired. If event loading exceeds eventLoadTimeoutMs after the sub-case already reached IDLE, the catch reports timeoutMs / 1000 s (300s, the wrong duration) and calls killSubCase on a sub-case that actually finished successfully. Suggest distinguishing the two timeouts (a separate try, or a sentinel) so the message and the kill only apply to the real run timeout.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +603,61 @@ class CaseServiceImpl(
...
     }
 
+    override fun killSubCase(caseId: UUID) = killCase(caseId)
+
     override fun killCase(caseId: UUID) {
         logger.info { "Killing case: $caseId" }
+        // Propagate kill to all direct sub-cases first (depth-first recursion).
+        // Each recursive call will in turn kill its own sub-cases, so arbitrarily
+        // deep delegation chains are handled without any extra tracking.
+        caseRepository.findActiveByParentCaseId(caseId).forEach { subCase ->
```
selimbensenouciep-prog: **Major: kill propagation overwrites terminal statuses, and there is no cleanup on success.** findActiveByParentCaseId filters on removed only (not status) and handleStatusChange has no transition guard, so killing a parent re-kills a sub-case that already reached ERROR, overwriting ERROR with KILLED and losing the diagnostic. The recursion also has no visited set, so a cyclic parentCaseId would loop forever. Separately, only the timeout path in DelegationTool calls killSubCase: the normal IDLE completion path leaves the sub-case runtime registered until idle eviction. Suggest skipping terminal sub-cases when propagating, and an iterative walk with a visited set.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +603,61 @@ class CaseServiceImpl(
...
+    ): CaseRuntime {
+        val user =
+            userService.findById(userId)
+                ?: throw ResourceNotFoundException("User not found: $userId")
+        val displayName =
+            listOfNotNull(user.firstname, user.lastname)
+                .joinToString(" ")
+                .ifBlank { userId.toString() }
+        val actor =
+            Actor(
```
selimbensenouciep-prog: **Major: the sub-agent runs with the user's full identity, and the authorization is split.** The sub-case is created with Actor(role = USER, id = userId) and there is no re-check that the user still has access to the namespace or the target agent at delegation time; the only gate is the allowlist applied at parent tool-call time. Note also that namespaceId is baked into the DelegationTool at build time while userId comes from the context at execution time, so they can disagree if the tool instance outlives a context change. Suggest anchoring identity and the access check to the parent case record.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,189 @@
...
+        .let { objectMapper.writeValueAsString(it) }
+
+    override suspend fun execute(
+        input: Args?,
+        context: ToolContext,
+    ): ToolExecutionResult {
+        if (input == null) {
+            return ToolExecutionResult(
+                output = "Delegation requires agentName and task arguments.",
+                success = false,
```
selimbensenouciep-prog: **Minor (reuse): build failures via the SDK factories.** These results are constructed by hand as ToolExecutionResult(output = ..., success = false), so delegation failures (allowlist reject, null user, timeout, terminal status) carry no errorType or errorMessage, unlike every other tool. The sibling RedirectTool uses ToolExecutionResult.error(output, errorType = ...) and .success(...). Reusing those keeps a machine-readable error category for downstream metrics and handling. Same area: startSubCase uses activeRuntimes[subCase.id]!! where the rest of the file uses ?: throw ResourceNotFoundException(...).

selimbensenouciep-prog (commented):
Review of the delegation feature, focused on correctness, security, and fit with the existing infra. Inline comments are tagged Critical / Major / Minor. The three Critical items (self/cycle/depth guard, IDLE treated as success, sub-case permission grant) are the ones worth resolving before merge.

A few smaller items not inlined:
- DelegationTool KDoc says it blocks 'via runBlocking', but execute is a suspend fun using withTimeout. Docs should match the code.
- The timeout message uses integer division (timeoutMs / 1000), so sub-second timeouts render as 'after 0s' (a test even asserts this).
- The allowedAgents.isEmpty() description branch and the empty-enum schema in DelegationTool are unreachable, since buildDelegationTools returns emptyList when there are no patterns.
- The User to Actor construction is duplicated (also in CaseController). A User.toUserActor() helper would centralize it.
- The agentName-enum inputSchema duplicates RedirectTool's schema builder.
- parentCaseId is filtered on in findActiveByParentCaseId but does not appear to get a Neo4j index in this PR.
- extractLastAgentMessage reloads the full sub-case history from Neo4j even though the in-memory runtime is already held.
- interruptCase is not propagated to sub-cases (only killCase is), so interrupting the parent leaves the sub-case running.

vincent-audibert-whoz:
## Additional commit: delegation robustness fixes

Six issues addressed on top of the initial delegation implementation:

**1. Kill propagation respects terminal sub-cases** — `findActiveByParentCaseId` now excludes sub-cases already in `KILLED`/`ERROR` from the Cypher query. Killing a parent no longer overwrites the diagnostic status of sub-cases that already completed with an error.

**2. `QuestionEvent` returned as structured result** — When a sub-case reaches `IDLE` with a pending `QuestionEvent` (no agent message emitted after it), `DelegationTool` returns `success=true` with a `pendingQuestion` field and the `subCaseId`. The parent agent can relay the question to the user or resolve it autonomously, then resume via `subCaseId`.

**3. `subCaseId` optional arg for resuming existing sub-cases** — `DelegationTool.Args` gains an optional `subCaseId: UUID?`. When provided, the tool calls the new `SubCaseLauncher.resumeSubCase` instead of `startSubCase`, injecting the new task as a user message into the existing IDLE sub-case.

**4. ADMIN grant on sub-case creation** — `startSubCase` now calls `permissionService.grantPermission(ADMIN)` for the delegating user on the new sub-case. Without this the user could not list or stream the sub-case. Best-effort (logs WARN on failure).

**5. Delegation depth limit (max 5)** — `linkParentToChild` creates a `[:PARENT_OF]` Neo4j edge after each `startSubCase`. `countAncestorDepth` traverses `[:PARENT_OF*]` to count hops. `startSubCase` rejects creation when depth ≥ 5.

**6. Timeout isolation** — The run-timeout and event-load `withTimeout` are now in separate `try/catch` blocks, preventing a slow event store from accidentally triggering the kill path.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentService.kt
```diff
@@ -21,11 +22,19 @@ interface AgentService {
...
+     * [io.whozoss.agentos.delegation.DelegationTool] is added to the agent's tool set.
+     * The allowed agents list is computed by resolving the subAgents patterns against
+     * the agents actually accessible to the current user in the namespace.
+     *
      * Throws if no model matches [namePart].
      */
     suspend fun findAgentByName(
         namePart: String,
         context: AgentExecutionContext,
+        subCaseLauncher: SubCaseLauncher? = null,
```
vincent-couturier-whoz: question: why a subCaseLauncher needed to findAgentByName ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -419,5 +427,57 @@ class AgentServiceImpl(
...
+        val allowedAgents = accessibleNames.filter { name -> regexes.any { it.matches(name) } }
+
+        if (allowedAgents.isEmpty()) {
+            logger.warn { "DelegationTool: no accessible agents matched patterns $patterns for agent '${config.name}'" }
+            return emptyList()
+        }
+
+        logger.info { "Adding DelegationTool for agent '${config.name}' with allowedAgents=$allowedAgents" }
+        return listOf(
+            DelegationTool(
```
vincent-couturier-whoz: suggestion: provide only data information and have the business logic managed by a service with dependency injection to allow low coupling between contract and implementation.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
vincent-couturier-whoz, line 89: question : why the redirect is managed by tool configuration whereas we manage delegation by specific agent config ? perhaps we could manage delegation through an evolution of the existing REDIRECT which is a kind of delegation.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -82,4 +83,57 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+     * a parent must not overwrite the diagnostic status of sub-cases that already completed.
+     *
+     * The BELONGS_TO edge is always present (written by [Neo4jCaseRepository.save]),
+     * so MATCH (not OPTIONAL MATCH) is safe here.
+     */
+    @Query(
+        $$"""MATCH (c:Case)-[r:BELONGS_TO]->(ns:Namespace)
+            WHERE c.parentCaseId = $parentCaseId
+              AND (c.removed IS NULL OR c.removed = false)
+              AND c.status <> 'KILLED' AND c.status <> 'ERROR'
```
vincent-couturier-whoz: suggestion: we could have findByParentCaseIdAndStatus and let the service layer add the needed status

vincent-couturier-whoz: suggestion: `AND NOT c.status IN ['KILLED', 'ERROR]`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -82,4 +83,57 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+     *
+     * Returns 0 when [caseId] has no parent, 1 when it has one parent, etc.
+     * Used by [io.whozoss.agentos.caseFlow.CaseServiceImpl.startSubCase] to enforce a
+     * maximum delegation depth before creating a new sub-case.
+     *
+     * Traverses the [:PARENT_OF] graph edges written by [linkParentToChild].
+     */
+    @Query(
+        $$"""MATCH (c:Case {id: $caseId})
+            OPTIONAL MATCH path = (c)<-[:PARENT_OF*]-(ancestor:Case)
```
vincent-couturier-whoz: suggestion: add a limit to avoid big cycles or any other risk on size of the relationship.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -82,4 +83,57 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+            RETURN CASE WHEN path IS NULL THEN 0 ELSE length(path) END AS depth
```
vincent-couturier-whoz: suggestion: `RETURN length(coalesce(path,[])) as depth`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -82,4 +83,57 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+            ORDER BY depth DESC LIMIT 1
```
vincent-couturier-whoz, line 129: suggestion: SHORTEST PATH should be preferred

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -82,4 +83,57 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+    /**
+     * Find all active, non-terminal sub-cases whose parentCaseId matches, with their namespace edge.
+     *
+     * Excludes sub-cases that are already in a terminal status (KILLED or ERROR) — killing
+     * a parent must not overwrite the diagnostic status of sub-cases that already completed.
+     *
+     * The BELONGS_TO edge is always present (written by [Neo4jCaseRepository.save]),
+     * so MATCH (not OPTIONAL MATCH) is safe here.
+     */
+    @Query(
```
vincent-couturier-whoz, line 105: suggestion: perhaps we should add `@Transactional(readonly=true)` to ensure better use of readonly query

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +607,130 @@ class CaseServiceImpl(
...
+                id = userId.toString(),
+                displayName = displayName,
+                role = ActorRole.USER,
+            )
+        val runtime = getCaseRuntime(subCaseId)
+        runtime.addUserMessage(actor, listOf(MessageContent.Text("@$agentName $task")))
+        scope.launch { runtime.run() }
```
vincent-couturier-whoz, line 676: suggestion: perhaps we should use coroutine structured concurrency to ensure proper coroutine cancellation. Perhaps @thomasmartin-whoz or Nicolas Remy could help on this topic.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -600,14 +607,130 @@ class CaseServiceImpl(
...
+                    namespaceId = namespaceId,
+                    title = "Sub-case: $task".take(120),
+                    parentCaseId = parentCaseId,
+                ),
+            )
+
+        // Grant the delegating user ADMIN on the sub-case so they can list,
+        // open, and stream it — same grant that CaseController.create applies
+        // for user-created cases.
+        runCatching {
```
vincent-couturier-whoz: suggestion: externalize the creation and of subcase in dedicated method with the needed transactional annotation to ensure data coherency. Not sure that the runCatching here allow a rolllback in case of failure on permission grant and/or creation of parent link.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,352 @@
...
+        else ->
+            "Delegate one or more tasks to specialized sub-agents and wait for all results. " +
+                "All delegations run in parallel — pass multiple entries to fan out work concurrently. " +
+                "Sub-agents run autonomously; if one needs to ask a question, the question is returned so you can relay it or resolve it, then resume via subCaseId. " +
+                "Available agents: ${allowedAgents.joinToString(", ")}."
+    }
+    override val version = "1.0.0"
+    override val paramType: Class<Args> = Args::class.java
+
+    override val inputSchema: String = objectMapper
```
vincent-couturier-whoz: suggestion: split this tool into three tool. One to start delegation on a subcase, one to get information from a subcase and another one to add a message in a sub case. It avoid optional caseId argument which could be complicated to manage.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        // Run all delegations in parallel, bounded by the global wall-clock timeout.
+        val results: List<DelegationResult> =
+            try {
+                withTimeout(timeoutMs) {
+                    coroutineScope {
+                        input.delegations.map { delegation ->
+                            async { runSingleDelegation(delegation, userId) }
+                        }.map { it.await() }
```
vincent-couturier-whoz: question: what's the behavior of a sequential coroutine await ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/SubCaseLauncher.kt
```diff
@@ -0,0 +1,55 @@
...
+ * dependency between [io.whozoss.agentos.agent.AgentServiceImpl] and
+ * [io.whozoss.agentos.caseFlow.CaseService]:
+ *
+ *   CaseService → AgentService → CaseService  (cycle)
+ *   CaseService → AgentService → SubCaseLauncher  (no cycle)
+ *
+ * [io.whozoss.agentos.caseFlow.CaseServiceImpl] implements this interface alongside
+ * [io.whozoss.agentos.caseFlow.CaseService] — no logic is duplicated.
+ */
+interface SubCaseLauncher {
```
vincent-couturier-whoz: suggestion: it seems to be closer to a SubCaseManager or SubCaseOrchestrator

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
```diff
@@ -1270,6 +1282,43 @@ class CaseServiceImplSpec :
...
+            val subCase2 = service.create(Case(namespaceId = namespaceId, parentCaseId = parentCase.id))
+
+            service.killCase(parentCase.id)
+
+            service.getById(parentCase.id).status shouldBe CaseStatus.KILLED
+            service.getById(subCase1.id).status shouldBe CaseStatus.KILLED
+            service.getById(subCase2.id).status shouldBe CaseStatus.KILLED
+        }
+
+        "killing a parent case kills nested sub-sub-cases recursively" {
```
vincent-couturier-whoz, line 1446: question: how you manage coroutine in this test ? 🤔

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/delegation/DelegationToolUnitSpec.kt
vincent-couturier-whoz, line 1: question: how we ensure coherency of coroutine state ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
vincent-audibert-whoz: Guard added with a maximum depth level of 5 (arbitrary, in practice enough for very advanced agent compositions).
If an agent is designed that way, it should be able to delegate to itself.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
vincent-audibert-whoz: The redirect is a tool toward the outside of the agent, away from its responsability. In practice, we will have the configuration of a few main agents that one or several standard or unlinked agents can redirect to, passing the responsability of the conversation to more specialized agents. Redirect configuration has some probability to be shared by several agents.

Delegation is the opposite: the parent agent keeps control of the conversation/task and uses its sub-agents for him. We are there on tree configurations specific to each team of specialized agents under one or more higher-level coordinators. Delegation configuration will probably be specific to each orchestrator agent.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: fixed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: This is covered by the runtime of the sub-case that will instantiate for the user, according to the agents he can access.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
vincent-audibert-whoz: fixed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentService.kt
vincent-audibert-whoz: To break the CaseService -> AgentService -> CaseService dependency cycle.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
vincent-audibert-whoz: TooManyAbstractWordsInSentenceException 😅 
Delegation is tied to the inner workings of the framework through this subCaseLauncher interface, hence the specific status and build/injection among the other agent tools.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
vincent-audibert-whoz: Redirection and Delegation are completely different on the way the responsability of the calling agent is managed. See also [here](https://github.com/whoz-oss/coday/pull/1051#discussion_r3506790077) as why Redirect configs will often be shared and not delegate ones.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
vincent-audibert-whoz: Indeed but for now we have only a single use case, premature optimization ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
vincent-audibert-whoz: Apparently shortest path is only between known nodes and that is not the case here.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
vincent-audibert-whoz: From the Coday experience, this is a bad tool design to give the agent the responsability to coordinate async tasks through time, it would need a wait or wake up tool besides. I tested the async delegation but the sole concept that worked was:
- have the tool respond by saying the sub-task/agent is started on thread xyz and will post a result when finished.
- let the parent agent do other stuff with the user
- on sub-thread finished, post back the last sub-agent message as result (same as sync delegation)
Hence the priority on sync delegation (with sensible timeouts).

When the agent has to check intentionally the result on a sub-thread still running... agentic hell is unleashed very fast.

However, adding a tool to read the end of another case can be useful and would be redundant with the CASE internal plugin, to keep for later.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/SubCaseLauncher.kt
vincent-audibert-whoz: Yep, interface drifted a bit, renamed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: changed for O(1) query + simple loop

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: This runs already in the scope of the DelegationTool using it (and from the scope of the parent caseRuntime), is it not enough ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
vincent-audibert-whoz: ...but this is not sequential ?

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImplSpec.kt
vincent-audibert-whoz: killCase is synchronous, no coroutines in it. Distinct than the withTimeout and stuff from DelegationTool.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: Permission is apart, reworked to cover creation and linkage under transactional

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
vincent-audibert-whoz: withTimeout relocated appropriately, timeouts distinct.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/delegation/DelegationToolUnitSpec.kt
vincent-audibert-whoz: test added

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -482,5 +490,60 @@ class AgentServiceImpl(
...
+     * 4. Build a [DelegationTool] only when the resolved allowlist is non-empty and a
+     *    [context.caseId] is available (no delegation outside a live case).
+     */
+    private fun buildDelegationTools(
+        config: AgentConfig,
+        context: AgentExecutionContext,
+        subCaseManager: SubCaseManager?,
+    ): List<DelegationTool> {
+        val patterns = config.subAgents?.filter { it.isNotBlank() }
+        if (subCaseManager == null || patterns.isNullOrEmpty() || context.caseId == null) return emptyList()
```
thomasmartin-whoz: si la méthode ne fait rien lorsque `subCaseManager == null`, pourquoi mettre subCaseManager nullable ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
```diff
@@ -482,5 +490,60 @@ class AgentServiceImpl(
...
+     *    glob patterns via [globToRegex] (anchored, case-insensitive, `*` = any sequence).
+     *    Examples: `"*Fixer"` matches `BugFixer` and `StoryFixer`; `"*"` matches all agents.
+     * 4. Build a [DelegationTool] only when the resolved allowlist is non-empty and a
+     *    [context.caseId] is available (no delegation outside a live case).
+     */
+    private fun buildDelegationTools(
+        config: AgentConfig,
+        context: AgentExecutionContext,
+        subCaseManager: SubCaseManager?,
+    ): List<DelegationTool> {
```
thomasmartin-whoz: pourquoi on renvoie une liste ? j'ai l'impression que la taille max est de 1

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -74,12 +79,93 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
...
+    fun findActiveDescendants(caseId: String): List<CaseNode>
+
+    /**
+     * Creates the [:PARENT_OF] relationship from [parentCaseId] to [childCaseId].
+     *
+     * Called by [io.whozoss.agentos.caseFlow.Neo4jCaseRepository] after creating a sub-case
+     * so that [countAncestorDepth] can traverse the chain.
+     */
+    @Query(
+        $$$"""MATCH (parent:Case {id: $parentCaseId}), (child:Case {id: $childCaseId})
```
thomasmartin-whoz: pourquoi $$$ ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+        subCaseId: UUID,
+        agentName: String,
+        task: String,
+        userId: UUID,
+        allowedAgents: List<String>,
+    ): CaseRuntime {
+        val subCase =
+            findById(subCaseId)
+                ?: throw ResourceNotFoundException("Sub-case not found: $subCaseId")
+        require(subCase.status == io.whozoss.agentos.sdk.caseFlow.CaseStatus.IDLE) {
```
thomasmartin-whoz: j'aurais mis `check` plutôt que `require` puisque ce n'est pas un mauvais argument mais un mauvais état

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+        require(agentName in allowedAgents) {
+            "Agent '$agentName' is not in the delegation allowlist for sub-case $subCaseId."
+        }
+        val user =
+            userService.findById(userId)
+                ?: throw ResourceNotFoundException("User not found: $userId")
+        val displayName =
+            listOfNotNull(user.firstname, user.lastname)
+                .joinToString(" ")
+                .ifBlank { userId.toString() }
```
thomasmartin-whoz: ça devrait être une méthode sur `user` ou  `userService`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+        val actor =
+            Actor(
+                id = userId.toString(),
+                displayName = displayName,
+                role = ActorRole.USER,
+            )
```
thomasmartin-whoz: ceci aussi probablement

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
thomasmartin-whoz: là le scope est partagé, ça me semble ok comme design.
Le mieux serait de lancer les coroutines de délégation à l'intérieur de la coroutine du case principal, mais ça va demander de la refacto pour un gain marginal.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+    @Transactional
+    override fun startSubCase(
+        parentCaseId: UUID,
+        namespaceId: UUID,
+        agentName: String,
+        task: String,
+        userId: UUID,
+    ): CaseRuntime {
+        val ancestorDepth = caseRepository.countAncestorDepth(parentCaseId)
+        if (ancestorDepth >= MAX_DELEGATION_DEPTH) {
```
thomasmartin-whoz:
```suggestion
        check (ancestorDepth < MAX_DELEGATION_DEPTH) {"Delegation depth limit..."}
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+            throw IllegalStateException(
+                "Delegation depth limit ($MAX_DELEGATION_DEPTH) reached for case $parentCaseId. " +
+                    "Cannot create a sub-case at depth ${ancestorDepth + 1}.",
+            )
+        }
+
+        val user =
+            userService.findById(userId)
+                ?: throw ResourceNotFoundException("User not found: $userId")
```
thomasmartin-whoz: on pourrait comme dans Whoz avoir un userService.getById() qui renvoie un non nullable et throw si null.
ça éviterait ces early return et cette répétition.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+        val displayName =
+            listOfNotNull(user.firstname, user.lastname)
+                .joinToString(" ")
+                .ifBlank { userId.toString() }
+        val actor =
+            Actor(
+                id = userId.toString(),
+                displayName = displayName,
+                role = ActorRole.USER,
+            )
```
thomasmartin-whoz: displayName/actor encore copier/coller, cf plus haut

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+        // Use @mention syntax so the normal selectAgent resolution picks up the
+        // requested agent without any special-casing in the runtime.
+        val mentionedTask = "@$agentName $task"
+        val subCase =
+            create(
+                Case(
+                    namespaceId = namespaceId,
+                    title = "Sub-case: $task".take(120),
```
thomasmartin-whoz: magic number
à voir si on veut importer la méthode `tak` dans agentOS

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -610,14 +619,131 @@ class CaseServiceImpl(
...
+                EntityType.CASE,
+                subCase.id.toString(),
+                PermissionRelation.ADMIN,
+            )
+        }.onFailure { e ->
+            logger.warn(e) {
+                "Auto-ADMIN grant failed for sub-case ${subCase.id} (user $userId) — sub-case persisted. " +
+                    "Recovery: a super-admin or namespace ADMIN must grant ADMIN on the case manually."
+            }
+        }
```
thomasmartin-whoz, line 733: j'ai un doute sur ce catch, l'utilisateur va avoir un problème sans GRANT non ?
Et personne ne lira ces logs, user ou admin.

thomasmartin-whoz: pourquoi pas plutôt ajouter un message dans le caseParent qui dit que la délégation a échoué en cas d'exception dans la création du sub-case ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -663,6 +789,16 @@ class CaseServiceImpl(
...
     companion object : KLogging() {
+        /**
+         * Maximum depth of delegation chains allowed.
+         *
+         * A top-level case has depth 0. Its direct sub-case has depth 1, etc.
+         * If a parent case is already at depth [MAX_DELEGATION_DEPTH] − 1, creating
+         * a child would reach the limit and is rejected. This prevents runaway
+         * delegation chains without forbidding legitimate multi-level orchestration.
+         */
+        private const val MAX_DELEGATION_DEPTH = 5
```
thomasmartin-whoz, line 795: bonne idée.
est-ce que cela mériterait une clé de conf ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,366 @@
...
+        when {
+            allowedAgents.isEmpty() -> {
+                "No sub-agents are currently available for delegation. Do not attempt to delegate — handle the request yourself or inform the user that no sub-agent can address it."
+            }
+
+            else -> {
+                "Delegate one or more tasks to specialized sub-agents and wait for all results. " +
+                    "All delegations run in parallel — pass multiple entries to fan out work concurrently. " +
+                    "Sub-agents run autonomously; if one needs to ask a question, the question is returned so you can relay it or resolve it, then resume via subCaseId. " +
+                    "Available agents: ${allowedAgents.joinToString(", ")}."
```
thomasmartin-whoz: you should use """

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,366 @@
...
+        logger.info {
+            "[DelegationTool] Launching ${input.delegations.size} delegation(s) in parallel " +
+                "from parent case $parentCaseId (timeout=${timeoutMs}ms)"
+        }
+
+        // Run all delegations in parallel. Each delegation has its own timeout budget
+        // so a single slow sub-case does not cancel the others — successful results are
+        // preserved and only the timed-out ones are reported as failures.
+        val results: List<DelegationResult> =
+            coroutineScope {
```
thomasmartin-whoz, line 180: quel est le dispatcher ?

vincent-audibert-whoz: C’est `Dispatchers.IO` qui vient du `CaseServiceImpl.kt`

thomasmartin-whoz: ce n'est pas précisé pourtant, comment cela peut être inféré ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: constantifié.
Le `tak` que l’on a dans Whoz concerne les listes, pas pour une string (sauf si liste de char)?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
vincent-audibert-whoz: more is betteeeeeeeer 🙈

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentServiceImpl.kt
vincent-audibert-whoz: check remonté en amont

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: ok

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: adopté:
- kill case
- remontée exception (catch by DelegationTool = message pour l’output du tool)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: Je ne pense pas, on la montera progressivement si ça limite, les usages les plus avancés seront sur Coday je pense, je suis de temps en temps à une depth de 3...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: genre celui qui existe déjà 🥲 
corrigé.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
vincent-couturier-whoz: `.map { it.await() }` is by design sequential because it iterates on the array items.

thomasmartin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,371 @@
...
+                output = "Delegation requires at least one entry in the delegations list.",
+                errorType = "INVALID_INPUT",
+            )
+        }
+
+        val invalidAgents = input.delegations.map { it.agentName }.filter { it !in allowedAgents }
+        if (invalidAgents.isNotEmpty()) {
+            return ToolExecutionResult.error(
+                output =
+                    "Agent(s) ${invalidAgents.joinToString(", ") { "'$it'" }} are not in the delegation allowlist. " +
```
vincent-couturier-whoz, line 158: suggestion: show only the first tenth invalidAgents and indicate the number of invalid Agents to avoid too big output

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -0,0 +1,371 @@
...
+     * the sub-case and returns a failure result, leaving sibling delegations unaffected.
+     */
+    private suspend fun runSingleDelegation(
+        delegation: Delegation,
+        userId: UUID,
+    ): DelegationResult {
+        val agentName = delegation.agentName
+        val subRuntime =
+            runCatching {
+                when (val resumeId = delegation.subCaseId) {
```
vincent-couturier-whoz, line 215: suggestion: resumedCaseId or just subCaseId

vincent-couturier-whoz (approved):

PR #1062, opened 2026-06-26
feat: wz-33057 ScheduledPrompt CRUD & Execution Trigger

fdelsert-whoz:
### What was built

A complete scheduling feature allowing agents to automatically trigger conversations at predefined times, without manual intervention.

### Data model

Introduced ScheduledPrompt as a first-class entity replacing the former CaseDefinition. The entity captures:
- **Identity & scope:** name, description, agentConfigId, namespaceId/userId (platform / user-global / namespace-shared / user×namespace overlay model — 4 priority levels)
- **Recurrence:** unit (WEEK/MONTH only, no DAY, no interval multiplier) + days (day-of-week filter, WEEK only) + timeUtc (LocalTime, UTC)
- **Planning:** startDate, endType (NEVER/ON_DATE/OCCURRENCES), endDate, maxOccurrenceCount
- **Execution tracking:** nextRunAt: Instant (computed on create/update/toggle) + lastRunAt: Instant? + ScheduledPromptRun audit records (slotKey UNIQUE constraint for distributed dedup, statuses: CLAIMED/SKIPPED/RUNNING/DONE/FAILED)
- **Prompt lifecycle:** the opening message (promptContent) is managed transparently by the backend via a linked generic Prompt entity (promptTemplateId, named `scheduled--{nameSlug}`), invisible to API consumers
Naming design decisions for ScheduledPrompt

The `ScheduledPrompt.name` field is a free-form label ("Daily Standup", "Rapport hebdo") with no slug constraint. Uniqueness is enforced at the storage level via a `tripleKey` built from `name.toSlug()` — this normalization is required because the associated `Prompt.name` is derived from the same slug (`scheduled--{nameSlug}`), so two scheduled prompts whose names normalize to the same slug would produce the same prompt name and violate the `Prompt` uniqueness constraint.

The `scheduled--` prefix avoids any collision with Agentic Studio starter prompts, which follow the `{agentSlug}--{labelSlug}` convention. The agent name is deliberately excluded from the prompt name: since `ScheduledPrompt` already carries `agentConfigId` as the authoritative reference to its target agent, embedding the agent name would create a derived value that silently drifts whenever the agent is renamed.

### Scheduling logic

NextRunCalculator computes nextRunAt directly from the descriptive recurrence model — no cron expression involved.

Supports:
  - WEEK with optional day-of-week filter (defaults to startDate's weekday if no days specified)
  - MONTH anchored on startDate's day, with last-day-of-month fallback for short months (e.g. Jan 31 → Feb 28)

### Execution engine skeleton

SchedulerScanner runs on a configurable fixed-delay tick (scheduler.tick-interval-ms):
  - Queries all due ScheduledPrompts (nextRunAt ≤ now, enabled = true)
  - For each: skips if a previous run is still active (overlap policy), otherwise claims the slot
  - The insert of ScheduledPromptRun acts as a distributed lock — a UNIQUE constraint on slotKey `(scheduledPromptId|scheduledForEpochMilli)` prevents double-execution in multi-instance deployments
  - Always advances nextRunAt after claiming (self-healing: survives crashes between insert and pointer advance)
  - Execution is a no-op for now (TODO execute log), the handler will be wired in a future PR

### API

REST endpoints at `/api/scheduled-prompts`: CRUD (POST/GET/PUT/DELETE), search, effective (overlay resolution), toggle (enable/disable). The TypeScript Angular client is fully generated from the OpenAPI spec, no hand-written HTTP client.

### Frontend

Full CRUD form with two fieldsets (Recurrence and Planning), conditional day-of-week checkboxes, time picker (UTC), date pickers, end condition selector. The item card displays `nextRunAt` and `lastRunAt` for at-a-glance scheduling status.

Capture d’écran 2026-07-30 à 16 25 18


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseDefinition/CronExpressionConverter.kt
vincent-audibert-whoz, line 1: Warning, when building the schedulers for Coday, the CRON proved to be significantly unfit to business processes that are on a less strict period (2 weeks, first of month, etc...).
If it is just used to compute the next run for a scheduler engine, that’s ok short-term. But if the scheduler engine only works with cron directly, we’ll run into issues in the near future.

fdelsert-whoz: The current model is:

Recurrence(every, unit, days, timeUtc) + Planning(startDate, endType, ...): stored as-is in the database, no cron expression involved.

NextRunCalculator: computes nextRunAt: Instant directly from these descriptive fields.
SchedulerScanner: works exclusively with nextRunAt (a plain Instant), never with a cron expression.

The only cron in the system is the `@Scheduled` tick on the scanner itself (a fixed-delay heartbeat), not the scheduling logic. The engine never sees a cron expression for business schedules, it simply knows "this scheduledPrompt is due at this instant" and advances the pointer via NextRunCalculator.nextAfter(), which natively handles business cases like "every 2 months", "specific days of the week", etc.

The trigger part is in place but execution will be handled in a future PR.

agentos/agentos-service/src/main/resources/application.yml
```diff
@@ -4,6 +4,8 @@ spring:
   jackson:
     mapper:
       sort-properties-alphabetically: true
+    serialization:
+      write-dates-as-timestamps: false
```
fdelsert-whoz, line 8: Heu ...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,310 @@
...
+    // Read endpoints
+    // -------------------------------------------------------------------------
+
+    @GetMapping("/{id}")
+    @PreAuthorize("hasPermission(#id, 'ScheduledPrompt', 'READ')")
+    @HideOnAccessDenied
+    override fun getById(@PathVariable id: UUID): ScheduledPromptDto = crud.getById(id)
+
+    @PostMapping("/by-ids", consumes = [MediaType.APPLICATION_JSON_VALUE], produces = [MediaType.APPLICATION_JSON_VALUE])
+    @PreAuthorize("isAuthenticated()")
```
fdelsert-whoz, line 115: Trop léger ?

fdelsert-whoz: Les vérifications sont faites coté service.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptApi.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * POST /api/scheduled-prompts/effective — effective merged set for a user in a namespace.
+     */
+    fun effective(request: ScheduledPromptEffectiveRequest): List<ScheduledPromptDto>
+
+    /**
+     * PATCH /api/scheduled-prompts/{id}/toggle — toggle the enabled flag.
+     */
+    fun toggle(id: UUID): ScheduledPromptDto
```
thomasmartin-whoz: kinda strange semantic, a boolean parameter `enabled` would be clearer

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptApi.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * POST /api/scheduled-prompts/search — list scheduled prompts at an exact scope level.
+     */
+    fun search(request: ScheduledPromptSearchRequest): List<ScheduledPromptDto>
+
+    /**
+     * POST /api/scheduled-prompts/effective — effective merged set for a user in a namespace.
+     */
+    fun effective(request: ScheduledPromptEffectiveRequest): List<ScheduledPromptDto>
```
thomasmartin-whoz: "effective" does not sound like a list method to me, it should be more precise semantically

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
```diff
@@ -0,0 +1,130 @@
...
+ */
+@Schema(name = "Planning")
+data class PlanningDto(
+    @field:NotNull(message = "startDate must not be null")
+    val startDate: LocalDate,
+    @field:NotNull(message = "endType must not be null")
+    val endType: SchedulerEndType,
+    val endDate: LocalDate? = null,
+    @field:Positive(message = "occurrenceCount must be > 0")
+    val occurrenceCount: Int? = null,
```
thomasmartin-whoz: this should be `maxOccurencesCount` to not be confused with actual runtime count of occurences

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
```diff
@@ -0,0 +1,130 @@
...
+    @field:Valid
+    val recurrence: RecurrenceDto,
+    @field:NotNull(message = "planning must not be null")
+    @field:Valid
+    val planning: PlanningDto,
+    val enabled: Boolean = true,
+    /** Next scheduled run time (UTC). Read-only: calculated by the backend, ignored on POST/PUT. */
+    val nextRunAt: Instant? = null,
+    /** Last time this scheduled prompt was triggered. Read-only. Null until first run. */
+    val lastRunAt: Instant? = null,
```
thomasmartin-whoz, line 136: wondering if nextRunAt and lastRunAt should be exposed to the API.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
```diff
@@ -0,0 +1,130 @@
...
+    val createdBy: String? = null,
+    val createdOn: Instant? = null,
+    val updatedBy: String? = null,
+    val updatedOn: Instant? = null,
```
thomasmartin-whoz, line 140: why is this entity not compliant with Entity interface ?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptEffectiveRequest.kt
```diff
@@ -0,0 +1,33 @@
...
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val namespaceId: UUID? = null,
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val userId: UUID? = null,
+    @field:Schema(types = ["string", "null"])
+    val namespaceExternalId: String? = null,
+    @field:Schema(types = ["string", "null"])
+    val userExternalId: String? = null,
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val agentConfigId: UUID? = null,
```
thomasmartin-whoz, line 34: there should be a constraint (@AssertTrue) on either namespaceId or namespaceExternalId is not null.
Same for userId/userExternalId

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptSearchRequest.kt
```diff
@@ -0,0 +1,35 @@
...
+ * [agentConfigIds] is an optional filter: when provided, only scheduled prompts linked
+ * to one of those agents are returned. When null or empty, all scheduled prompts at the
+ * resolved scope level are returned.
+ */
+@Schema(name = "ScheduledPromptSearchRequest")
+data class ScheduledPromptSearchRequest(
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val namespaceId: UUID? = null,
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val userId: UUID? = null,
```
thomasmartin-whoz, line 39: why no userExternalId here ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/ClockConfiguration.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Declared with [ConditionalOnMissingBean] so tests can substitute a fixed clock
+ * without conflicts. The Neo4j persistence configuration also declares this bean
+ * (for the case where it is the only active configuration), but this class
+ * ensures the bean is available in all profiles including `test` (in-memory).
+ */
+@Configuration
+class ClockConfiguration {
+    @Bean
+    @ConditionalOnMissingBean(Clock::class)
+    fun utcClock(): Clock = Clock.systemUTC()
```
thomasmartin-whoz, line 20: why this ? I think the Whoz way is to handle the clock thanks to infra

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * **MONTH**:
+ * - Fires every month on the same day-of-month as `startDate`.
+ * - If the target month is shorter than `startDate.dayOfMonth`, the last day of that month is used.
+ * - [Recurrence.days] is ignored for MONTH.
+ *
+ * The calculator is **pure** (no side effects) and takes an explicit [clock] so tests can inject
+ * a fixed instant.
+ */
+object NextRunCalculator {
```
thomasmartin-whoz: should not this be a spring service ? especially when a clock is in parameters

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Compute the next run instant.
+     *
+     * @param sp the scheduled prompt whose [ScheduledPrompt.recurrence] and [ScheduledPrompt.planning] drive the calculation
+     * @param clock source of "now"; defaults to UTC system clock
+     * @return UTC instant of the next run
+     */
+    fun compute(sp: ScheduledPrompt, clock: Clock = Clock.systemUTC()): Instant {
```
thomasmartin-whoz:
```suggestion
    fun compute(scheduledPrompt: ScheduledPrompt, clock: Clock = Clock.systemUTC()): Instant {
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+     */
+    private fun computeWeek(candidate: LocalDateTime, recurrence: Recurrence, planning: Planning): LocalDateTime {
+        val timeUtc = recurrence.timeUtc
+        val startDate = planning.startDate
+        val days = recurrence.days
+
+        return if (days.isEmpty()) {
+            computeWeekSameDay(candidate, startDate, timeUtc)
+        } else {
+            computeWeekWithDayFilter(candidate, startDate, days, timeUtc)
```
thomasmartin-whoz: named params, in the entire class

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * WEEK without day filter: fires every week on the same day-of-week as [startDate].
+     */
+    private fun computeWeekSameDay(
+        candidate: LocalDateTime,
+        startDate: LocalDate,
+        timeUtc: LocalTime,
+    ): LocalDateTime {
+        val targetDow = startDate.dayOfWeek
```
thomasmartin-whoz: avoid short names like "dow"

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+        var slotDate = startDate.with(TemporalAdjusters.nextOrSame(targetDow))
+
+        while (slotDate.isBefore(candidate.toLocalDate())) {
```
thomasmartin-whoz: is a loop really necessary here ? 
if it helps, you can look at LocalDateExtensions in whoz repository, there are a bunch of helpful method to manipulate dates.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculatorService.kt
```diff
@@ -0,0 +1,188 @@
...
+    ): LocalDateTime {
+        val targetDow = startDate.dayOfWeek
+        var slotDate = startDate.with(TemporalAdjusters.nextOrSame(targetDow))
+
+        while (slotDate.isBefore(candidate.toLocalDate())) {
+            slotDate = slotDate.plusWeeks(1)
+        }
+
+        val slotDateTime = slotDate.atTime(timeUtc)
+        return if (!slotDateTime.isBefore(candidate)) slotDateTime
```
thomasmartin-whoz, line 107:
```suggestion
        return if (slotDateTime.isAfter(candidate)) slotDateTime
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculatorService.kt
```diff
@@ -0,0 +1,188 @@
...
+        days: List<DayOfWeek>,
+        timeUtc: LocalTime,
+    ): LocalDateTime {
+        val sortedDays = days.map { it.value }.toSortedSet() // 1=MON..7=SUN
+        var searchDate = if (candidate.toLocalDate().isBefore(startDate)) startDate else candidate.toLocalDate()
+
+        repeat(MAX_SEARCH_DAYS) {
+            if (searchDate.dayOfWeek.value in sortedDays) {
+                val slotDateTime = searchDate.atTime(timeUtc)
+                if (!slotDateTime.isBefore(candidate)) return slotDateTime
```
thomasmartin-whoz, line 127:
```suggestion
                if (slotDateTime.isAfter(candidate)) return slotDateTime
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        repeat(MAX_SEARCH_DAYS) {
+            if (searchDate.dayOfWeek.value in sortedDays) {
+                val slotDateTime = searchDate.atTime(timeUtc)
+                if (!slotDateTime.isBefore(candidate)) return slotDateTime
+            }
+            searchDate = searchDate.plusDays(1)
+        }
+        // Fallback: should never happen with a non-empty days list
+        return candidate.toLocalDate().plusDays(8).atTime(timeUtc)
```
thomasmartin-whoz: MAX_SEARCH_DAYS should be used here I guess ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPrompt.kt
```diff
@@ -0,0 +1,102 @@
...
+ * [endType] controls termination:
+ * - NEVER: runs indefinitely.
+ * - ON_DATE: stops after [endDate] (required, strictly after [startDate]).
+ * - OCCURRENCES: stops after [occurrenceCount] executions (required, > 0).
+ */
+data class Planning(
+    val startDate: LocalDate,
+    val endType: SchedulerEndType = SchedulerEndType.NEVER,
+    val endDate: LocalDate? = null,
+    val occurrenceCount: Int? = null,
```
thomasmartin-whoz: maxOccurencesCount

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPrompt.kt
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// ---------------------------------------------------------------------------
+// Aggregate root
+// ---------------------------------------------------------------------------
+
+/**
+ * A declarative, quasi-immutable definition of a scheduled prompt.
```
thomasmartin-whoz: what does "quasi-immutable" mean ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+) : ScheduledPromptApi {
+
+    private val crud =
+        EntityCrudDelegate(
+            service = scheduledPromptService,
+            userService = userService,
+            permissions = permissionService,
+            entityType = EntityType.SCHEDULED_PROMPT,
+            toResource = { entity ->
+                val sp = entity as ScheduledPrompt
```
thomasmartin-whoz:
```suggestion
                val scheduledPrompt = entity as ScheduledPrompt
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+        }
+        if (resolvedNs != null) {
+            val granted = permissionService.hasPermission(
+                userId = currentUser.id.toString(),
+                entityType = EntityType.NAMESPACE,
+                entityId = resolvedNs.toString(),
+                action = Action.READ,
+            )
+            if (!granted) throw AccessDeniedException("Cannot read scheduled prompts in namespace $resolvedNs")
+        }
```
thomasmartin-whoz: should not this be on PreAuthorize ?

Also there is inconsistency in design: the /by-ids endpoint let the service handle the permission part, while "/search" does it in the controller part.
You should make a choice and stick to it for the all class

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    @Operation(summary = "Effective scheduled prompts for a user in a namespace")
+    @PostMapping("/effective", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @PreAuthorize("isAuthenticated()")
+    override fun effective(@Valid @RequestBody request: ScheduledPromptEffectiveRequest): List<ScheduledPromptDto> {
+        val nsId = resolveNamespaceId(request.namespaceId, request.namespaceExternalId)
+        val uId = resolveUserId(request.userId, request.userExternalId)
+        val currentUser = userService.getCurrentUser()
+        if (uId != currentUser.id) throw BadRequestException("userId must match authenticated user")
```
thomasmartin-whoz: is there a point to have userId param if it has to be the authenticated user ? Why not resolve the authenticated user by the endpoint ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+        val granted = permissionService.hasPermission(
+            userId = currentUser.id.toString(),
+            entityType = EntityType.NAMESPACE,
+            entityId = nsId.toString(),
+            action = Action.READ,
+        )
+        if (!granted) throw AccessDeniedException("Cannot read scheduled prompts in namespace $nsId")
+        return scheduledPromptService
+            .findEffective(nsId, currentUser.id)
+            .filter { request.agentConfigId == null || it.agentConfigId == request.agentConfigId }
```
thomasmartin-whoz: this seems out of responsibility of the controller part. It should be in the service layer, and even in the repository layer.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+        if (resource.userId != null && resource.userId != me) {
+            throw BadRequestException("userId in body must match authenticated user or be omitted")
+        }
+
+        val resolvedNs: UUID? = resource.namespaceId
+        val resolvedUser: UUID? = if (resource.userId != null) me else null
+        val isPlatform = resolvedNs == null && resolvedUser == null
+
+        when {
+            isPlatform -> if (!currentUser.isAdmin) throw AccessDeniedException("Platform-level ScheduledPrompt requires Super Admin")
```
thomasmartin-whoz: here it says it requires  Super Admin, but even a super admin would get this error

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+                    userId = me.toString(),
+                    entityType = EntityType.NAMESPACE,
+                    entityId = resolvedNs.toString(),
+                    action = authzAction,
+                )
+                if (!granted) throw AccessDeniedException(
+                    "Cannot create ScheduledPrompt in namespace $resolvedNs (${authzAction.name} required)",
+                )
+            }
+        }
```
thomasmartin-whoz: again it should be in PreAuthorize.
If not possible, we should have a dedicated bean to handle permission on ScheduledPrompt instead of duplicated code into the endpoint exposing methods.

The number of lines of code for this controller looks too large for me.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+
+        val entity = ScheduledPrompt(
+            metadata = EntityMetadata(id = resource.id ?: UUID.randomUUID()),
```
thomasmartin-whoz: I do not think client should be able to choose the entity id

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+            namespaceId = resolvedNs,
+            userId = resolvedUser,
+            agentConfigId = resource.agentConfigId,
+            promptTemplateId = UUID.randomUUID(), // placeholder — overwritten by createWithPrompt
```
thomasmartin-whoz, line 170: should not this be nullable then ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+        )
+        val (saved, promptContent) = scheduledPromptService.createWithPrompt(entity, resource.promptContent)
+        return toDto(saved, promptContent)
+    }
+
+    @PutMapping("/{id}", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @PreAuthorize("hasPermission(#id, 'ScheduledPrompt', 'WRITE')")
+    @HideOnAccessDenied
+    override fun update(@PathVariable id: UUID, @Valid @RequestBody resource: ScheduledPromptDto): ScheduledPromptDto {
+        val existing = scheduledPromptService.findById(id)
```
thomasmartin-whoz, line 186: we should have scheduledPromptService.getById() returning non nullable

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
```diff
@@ -0,0 +1,308 @@
...
+
+    private fun resolveUserId(id: UUID?, externalId: String?): UUID {
+        if (id != null && externalId != null) throw BadRequestException("Provide userId or userExternalId, not both")
+        return id
+            ?: externalId?.let {
+                userService.findByExternalId(it)?.metadata?.id
+                    ?: throw ResourceNotFoundException("User not found for externalId: $it")
+            }
+            ?: throw BadRequestException("userId or userExternalId is required")
+    }
```
thomasmartin-whoz: those resolving methods must already exist elsewhere. A dedicated helper or service should handle this.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNode.kt
```diff
@@ -0,0 +1,141 @@
...
+    // --- Planning fields (flattened) ---
+    val startDate: LocalDate,
+    val endType: String,
+    val endDate: LocalDate? = null,
+    val occurrenceCount: Int? = null,
+    // --- Common ---
+    val enabled: Boolean = true,
+    val nextRunAt: Instant,
+    val lastRunAt: Instant? = null,
+    val tripleKey: String,
```
thomasmartin-whoz, line 66: tripleKey is stored?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNode.kt
```diff
@@ -0,0 +1,141 @@
...
+            enabled = enabled,
+            nextRunAt = nextRunAt,
+            lastRunAt = lastRunAt,
+        )
+
+    companion object {
+        fun computeTripleKey(namespaceId: UUID?, userId: UUID?, name: String): String =
+            OverlayKeyEncoding.activeKey(namespaceId, userId, name.toSlug())
+
+        fun tombstoneTripleKey(id: String): String = OverlayKeyEncoding.tombstoneKey(id)
```
thomasmartin-whoz, line 111: I still don't understand the `tombstone` naming

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,137 @@
...
+              AND NOT COALESCE(sp.removed, false)
+            RETURN sp ORDER BY sp.name ASC
+            """,
+    )
+    fun findActivePlatform(): List<ScheduledPromptNode>
+
+    /** Find a single non-removed scheduled prompt matched by its [ScheduledPromptNode.tripleKey] discriminator. */
+    @Query(
+        $$"""
+            MATCH (sp:ScheduledPrompt {tripleKey: $tripleKey})
```
thomasmartin-whoz, line 42: I do not understand why we store the triple key. This is a form of denormalisation, why not just search on the three fields ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNode.kt
```diff
@@ -0,0 +1,141 @@
...
+    val description: String? = null,
+    // --- Recurrence fields (flattened) ---
+    val unit: String,
+    val days: List<String> = emptyList(),
+    val timeUtc: LocalTime,
+    // --- Planning fields (flattened) ---
+    val startDate: LocalDate,
+    val endType: String,
+    val endDate: LocalDate? = null,
+    val occurrenceCount: Int? = null,
```
thomasmartin-whoz: is this the real occurence count ? If so, where is stored the max ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * Sets `nextRunAt = nextSlot` only if the current stored value equals [currentSlot].
+     * This prevents a race between two concurrent scanner ticks overwriting each other's advance.
+     *
+     * @param id the scheduled prompt to advance
+     * @param currentSlot the expected current value of nextRunAt (CAS guard)
+     * @param nextSlot the new value to write
+     * @return true if the update was applied (the CAS matched), false if another tick beat us
+     */
+    fun advance(id: UUID, currentSlot: Instant, nextSlot: Instant): Boolean
```
thomasmartin-whoz: why is this not called advanceNextRunAt ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRunNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,25 @@
...
+        $$"""
+        MATCH (r:ScheduledPromptRun)
+        WHERE r.scheduledPromptId = $scheduledPromptId
+        AND r.status IN ['CLAIMED', 'RUNNING']
+        AND NOT COALESCE(r.removed, false)
+        RETURN count(r) > 0
+        """,
+    )
+    fun existsActiveByScheduledPromptId(scheduledPromptId: String): Boolean
+}
```
thomasmartin-whoz, line 25: shouldn't we have a `countOccurences` method here ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRunSchemaInitializer.kt
```diff
@@ -0,0 +1,60 @@
...
+        ensureSlotKeyUniqueConstraint()
+        ensureScheduledPromptIdIndex()
+        ensureStatusIndex()
+    }
+
+    private fun ensureSlotKeyUniqueConstraint() {
+        neo4jClient
+            .query(
+                "CREATE CONSTRAINT scheduled_prompt_run_slot_key_unique IF NOT EXISTS " +
+                    "FOR (r:ScheduledPromptRun) REQUIRE r.slotKey IS UNIQUE",
```
thomasmartin-whoz, line 38: what is slot key ? I am lost between tombstoneKey, tripleKey, slotKey...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
```diff
@@ -0,0 +1,267 @@
...
+    // -------------------------------------------------------------------------
+    // Core CRUD
+    // -------------------------------------------------------------------------
+
+    override fun create(entity: ScheduledPrompt): ScheduledPrompt {
+        val agentConfig = agentConfigService.findById(entity.agentConfigId)
+            ?: throw ResourceNotFoundException("AgentConfig not found: ${entity.agentConfigId}")
+        if (agentConfig.metadata.version == null) {
+            throw UnprocessableEntityException(
+                "AgentConfig id=${entity.agentConfigId} is a filesystem-only agent and cannot be linked to a ScheduledPrompt",
```
thomasmartin-whoz, line 58: what has this to do with optimistic locking ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
```diff
@@ -0,0 +1,267 @@
...
+                    throw BadRequestException("endDate must be after startDate")
+                }
+            }
+            SchedulerEndType.OCCURRENCES -> {
+                val count = entity.planning.occurrenceCount
+                    ?: throw BadRequestException("occurrenceCount is required when endType is OCCURRENCES")
+                if (count <= 0) throw BadRequestException("occurrenceCount must be > 0")
+            }
+            SchedulerEndType.NEVER -> Unit
+        }
```
thomasmartin-whoz: looks like something that could be validated by jackson or the controller

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
```diff
@@ -0,0 +1,267 @@
...
+        // Recalculate nextRunAt whenever recurrence or planning may have changed.
+        val withNextRun = entity.copy(nextRunAt = NextRunCalculator.compute(entity, clock))
+        return saveOrConflict(withNextRun)
+    }
+
+    override fun findById(id: UUID, withRemoved: Boolean): ScheduledPrompt? =
+        repository.findByIds(listOf(id), withRemoved).firstOrNull()
+
+    override fun findByIds(ids: Collection<UUID>, withRemoved: Boolean): List<ScheduledPrompt> =
+        repository.findByIds(ids, withRemoved)
```
thomasmartin-whoz, line 97: I thought service was handling permissions for this according to controller comments. It looks like it is not the case

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
```diff
@@ -0,0 +1,106 @@
...
+        val now = Instant.now(clock)
+        scheduledPromptRepository.findDue(now)
+            .also { due ->
+                if (due.isEmpty()) logger.debug { "[SchedulerScanner] tick: no due prompts" }
+                else logger.info { "[SchedulerScanner] tick: ${due.size} due prompt(s)" }
+            }
+            .forEach { sp -> claim(sp) }
+    }
+
+    private fun claim(sp: ScheduledPrompt) {
```
thomasmartin-whoz:
```suggestion
    private fun claim(scheduledPrompt: ScheduledPrompt) {
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
```diff
@@ -0,0 +1,106 @@
...
+        val slot = sp.nextRunAt
+        val correlationId = "sp-${sp.id.toString().take(8)}-${slot.epochSecond}"
+
+        // Guard: disable the ScheduledPrompt if its AgentConfig is gone or disabled.
+        val agentConfig = agentConfigService.findById(sp.agentConfigId)
+        if (agentConfig == null || !agentConfig.enabled) {
+            logger.warn {
```
thomasmartin-whoz, line 62: not sure warn is the right level

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
```diff
@@ -0,0 +1,106 @@
...
+                "[SchedulerScanner] AgentConfig ${sp.agentConfigId} is ${if (agentConfig == null) "deleted" else "disabled"} " +
+                    "— disabling sp=${sp.id} to prevent further zombie ticks"
+            }
+            scheduledPromptRepository.save(sp.copy(enabled = false))
+            return
+        }
+
+        val status = when {
+            runRepository.hasActive(sp.id) -> {
+                logger.warn { "[SchedulerScanner] OVERLAP sp=${sp.id} has active run → SKIPPED" }
```
thomasmartin-whoz: info ?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
fdelsert-whoz: I display them in the list on the agentos side.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
fdelsert-whoz: Because it's not an Entity but a plain flat Dto.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
fdelsert-whoz: and on top of that, throwing BadRequestException from the service layer feels really wrong.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: Totally agree, good catch.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
fdelsert-whoz: done at commit: 3293caffc2e23802efd533a45e20d4ab77ce023f

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
thomasmartin-whoz: I don't see how it is different...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/ClockConfiguration.kt
thomasmartin-whoz: I meant the timezone, not the clock

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNodeNeo4jRepository.kt
fdelsert-whoz: It is possible, but it doesn't protect against concurrent races, two simultaneous creations both pass the pre-check, and without a DB constraint, you end up with two silent duplicates.

tripleKey is a misleading name, it doesn't just encode the (namespaceId, userId, name) triple, it has a dual role:

- active: encodes the triple key, backing the UNIQUE constraint.
- soft-deleted: rewritten to tombstone:<id> to free the unique slot, allowing immediate recreation of an entity on the same triple after deletion.

Without this rewrite, the UNIQUE constraint would block any recreation, the soft-deleted row would permanently "occupy" the key. 

We denormalize into a single non-null String (tripleKey) because a composite UNIQUE constraint on (namespaceId, userId, name) would be silently ignored by Neo4j whenever one of the fields is NULL.

It is documented here: https://github.com/whoz-oss/coday/blob/c6ed526e642c2ae2b931c9fc407f4ca68423af60/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/persistence/OverlayKeyEncoding.kt

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRunSchemaInitializer.kt
fdelsert-whoz: slotKey (which prompt, for which fire-time) is completely different, it prevents double-firing of the same scheduler tick (multi-instances).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
fdelsert-whoz: Nothing.
It's being repurposed merely as a signal that this AgentConfig reference has no real Neo4j node (filesystem-only) and its id is therefore unsafe to use as a graph relationship/unique-key target.

It should be an explicit computed field or equivalent marker.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
fdelsert-whoz: To be replaced by `ChronoUnit.WEEKS.between` from `java.time.temporal.Duration`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculator.kt
fdelsert-whoz: done at commit: 40334eab6295e144ef2271e3617f4c5dbd9a1d4c

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculatorService.kt
fdelsert-whoz: keeped : isAfter excludes the equal case, while !isBefore includes it

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/NextRunCalculatorService.kt
fdelsert-whoz: keeped : isAfter excludes the equal case, while !isBefore includes it

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptApi.kt
fdelsert-whoz: toggle replaced by enable/disable endpoints like in AgentConfigController

done at commit: 5ec45b549120cba89fca903f30b45114d1fae975

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptSearchRequest.kt
fdelsert-whoz: done at commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptEffectiveRequest.kt
fdelsert-whoz: done at commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: fixed at commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: Yes but getById() can return removed entities.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: done at commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNode.kt
fdelsert-whoz: see: https://github.com/whoz-oss/coday/pull/1062#discussion_r3675129627

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNode.kt
fdelsert-whoz: see: https://github.com/whoz-oss/coday/pull/1062#discussion_r3675129627

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
fdelsert-whoz: done at commit: 347fc02ac2d5f7de61cc514d4bd09e04edd04282

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: fixed at commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: done at commit: 96a75f344ef66d0bcbd376c293839cde83bfe617

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptDto.kt
fdelsert-whoz: To be re-examined, but I won't fix this in this PR (out of scope given the many other entities following the flat-DTO pattern not tied to the domain Entity).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: embedded in commit: 04243f431b4644f0682ac41383ccb8eed95b0aa0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: Won't fix.

Making promptTemplateId nullable would force null-checks across every other consumer that safely dereferences it today.
The placeholder hack lives one line and is never persisted or exposed. A clean fix would add significantly more complexity.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
fdelsert-whoz: fixed at commit: 85978ad086bd4bf759953019f037264a5edb9d88

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRunNodeNeo4jRepository.kt
fdelsert-whoz: countOccurrences logic depends on actual execution wiring and complete run statuses; to be addressed in the PR that implements execution.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptRepository.kt
fdelsert-whoz: done at commit: 8cd19dca3ebd490e421ac5fc500b1cbc10312ee1

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptServiceImpl.kt
fdelsert-whoz: Kdoc fixed even if the inconsistency remains.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
fdelsert-whoz: To be reworked in the next "execution wiring" PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
fdelsert-whoz: To be reworked in the next "execution wiring" PR.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: authorization logic extracted and reused (search/effective/create for Prompt and ScheduledPrompt) via a new OverlayScopeAuthorizer bean.
No `@PreAuthorize` introduced in this PR, the security model stays unchanged, only the existing logic is centralized and shared between the two controllers.

see commit 809ac052ec2314dec24c2416812e162da63e119b

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: a reading error on your part, or a misunderstanding on mine ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/ClockConfiguration.kt
fdelsert-whoz: From what I have seen on the whoz side, the difference is that we can introduce an offset relative to system time via an env var. But the rest is similar.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: This missing by-ids filtering isn't specific to ScheduledPrompt, it affects nearly every entity.
It should be fixed in a dedicated, broader PR unifying the visibility predicate across all entities, not patched here.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptApi.kt
fdelsert-whoz: resolveEffective ?
done at commit: f2a8c909ae4e9313190c723d08a02ce0a7e4c86c

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptController.kt
fdelsert-whoz: OK, I just understood, I modified the explanatory comment on the controller class.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptSearchRequest.kt
```diff
@@ -31,5 +39,17 @@ data class PromptSearchRequest(
...
     val namespaceExternalId: String? = null,
+    @field:Schema(types = ["string", "null"])
+    val userExternalId: String? = null,
     val agentConfigIds: List<UUID>? = null,
-)
+) {
+    @get:AssertTrue(message = "Provide namespaceId or namespaceExternalId, not both")
+    @get:JsonIgnore
+    val isNamespaceIdentifierValid: Boolean
+        get() = namespaceId == null || namespaceExternalId == null
```
thomasmartin-whoz, line 49: if it is not both then it should be

```suggestion
        get() = (namespaceId == null) xor (namespaceExternalId == null)
```

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptSearchRequest.kt
```diff
@@ -31,5 +39,17 @@ data class PromptSearchRequest(
...
+
+    @get:AssertTrue(message = "Provide userId or userExternalId, not both")
+    @get:JsonIgnore
+    val isUserIdentifierValid: Boolean
+        get() = userId == null || userExternalId == null
```
thomasmartin-whoz, line 54: xor here too

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/scheduledPrompt/ScheduledPromptEffectiveRequest.kt
```diff
@@ -0,0 +1,40 @@
...
+    val namespaceId: UUID? = null,
+    @field:Schema(types = ["string", "null"])
+    val namespaceExternalId: String? = null,
+    @field:Schema(types = ["string", "null"], format = "uuid")
+    val agentConfigId: UUID? = null,
+) {
+    @get:AssertTrue(message = "Provide exactly one of namespaceId or namespaceExternalId")
+    @get:JsonIgnore
+    val isNamespaceIdentifierValid: Boolean
+        get() = (namespaceId == null) != (namespaceExternalId == null)
```
thomasmartin-whoz, line 39: this one is correct

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
```diff
@@ -116,4 +116,21 @@ data class AgentConfig(
...
+     *
+     * Filesystem agents are built in-memory on every read and never go through
+     * Spring Data Neo4j's `save()`, so [EntityMetadata.version] — which SDN sets to a
+     * non-null value on first persistence — stays `null` for their entire lifetime.
+     * This is an explicit, named proxy for that fact: callers that need to reject
+     * filesystem-only agents (e.g. before linking a [io.whozoss.agentos.prompt.Prompt] or a
+     * [io.whozoss.agentos.scheduledPrompt.ScheduledPrompt]) should read this property
+     * rather than re-deriving the same check from `metadata.version == null` at each
+     * call site.
+     */
```
thomasmartin-whoz, line 133: I still don't get it.
What is filesystem only ? And why is that checked against optimistic lock field ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/entity/ExternalIdentifierResolver.kt
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Two flavours per identifier type:
+ * - `resolve*Id` — exactly one of the pair is required; throws [BadRequestException]
+ *   when both are absent. Used by endpoints where the scope is mandatory (e.g. `effective`).
+ * - `resolveOptional*Id` — both may be absent (`null` result), used where the identifier
+ *   is one axis of an otherwise-optional scope (e.g. `search`, where `(null, null)` means
+ *   platform scope).
+ */
+@Service
+class ExternalIdentifierResolver(
```
thomasmartin-whoz, line 30: ❤️

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
```diff
@@ -54,7 +53,18 @@ import io.whozoss.agentos.sdk.api.common.GetByIdsRequest as SdkGetByIdsRequest
...
+ * body itself carries the `(namespaceId, userId)` scope that determines which check applies.
+ * A `@PreAuthorize` SpEL expression can only ever produce 403 on refusal, but the
+ * mass-assignment guard on `body.userId` must produce 400 (malformed request); and the
+ * external-id resolution ([io.whozoss.agentos.entity.ExternalIdentifierResolver], a Neo4j
+ * lookup) would have to run twice — once in SpEL to authorize, once in the method body to
+ * use the resolved value.
+ * [OverlayScopeAuthorizer] resolves once and throws the correctly-typed exception, so
+ * `@PreAuthorize("isAuthenticated()")` remains only as the declarative floor on these three.
  */
 @RestController
```
thomasmartin-whoz, line 69: looks way better in terms of single responsibility :)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
```diff
@@ -75,15 +73,27 @@ class PromptServiceImpl(
...
+    // The agentConfigId filter is intentionally applied here, in memory, and not pushed into
+    // the repository query. findEffective fetches raw candidates across all four overlay
+    // layers (platform / user-global / namespace-shared / user×namespace) for a given prompt
+    // name; which layer "wins" is only known after the groupBy+priority fold below. A layer
+    // that does NOT match agentConfigId can still be the eventual winner for its name, while a
+    // lower-priority layer that DOES match loses — filtering in the Cypher query (i.e. on the
+    // raw, pre-merge rows) would silently drop the winning row whenever its non-matching layer
+    // outranks a matching one, or spuriously keep a name whose only matching layer isn't the
+    // effective one. The filter must therefore run after the merge, on the already-resolved
+    // per-name winners — never before.
```
thomasmartin-whoz, line 85: all right

thomasmartin-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
fdelsert-whoz: Why ?

Filesystem agents are loaded from YAML files and never persisted in Neo4j, so they have no node, meaning the BELONGS_TO edge to them (used by the permission system)  can never be created.
To detect filesystem agents, we reuses the `@Version` field that Spring Data sets on first write. If it's null and the entity came through a repository but not from Neo4j, we reject linking to it.

The BELONGS_TO edge to AgentConfig is used in findEffective to check that the agent exists, is enabled, and the user has access but this could equivalently be done via a scalar filter on agentConfigId. In either case, filesystem agents would still be rejected since they have no Neo4j node to match against.

In the next PR, resolving which users are targeted by a ScheduledPrompt will be derived from the linked AgentConfig: the recipients are the users who have access to that agent.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptSearchRequest.kt
fdelsert-whoz: No because both can be null for search (platform level).
We only reject it when both are provided.

vincent-audibert-whoz (approved):

PR #1066, opened 2026-06-29
feat: #1009 add dark mode to AgentOS UI

### LINKED ISSUE #1009: Dark mode possible on AgentOS-ui
Have a dark mode possible in user settings (as well as system) on AgentOS-ui

selimbensenouciep-prog:
## What

Add a user-selectable **dark mode** (Light / Dark / System) to the AgentOS UI, available from the user settings page (`/agentos/me` → **Appearance**).

Closes #1009.

## Why

The AgentOS UI runs in its own standalone layout (route `/agentos`), without the legacy Coday client's sidebar where the theme selector lives — so once inside AgentOS there was no way to change the theme. The client already had a full theming system (CSS tokens + `data-theme` on `document.documentElement`), but it wasn't reachable from AgentOS.

## How

- **Theme port (DI inversion).** `agentos-ui` defines `THEME_PORT` (InjectionToken) + a small `ThemePort` interface, consumed by the shell and the profile page. It defaults to a self-contained `AgentosThemeService` so the lib works standalone; the Coday client binds `{ provide: THEME_PORT, useExisting: ThemeService }` on the lazy `/agentos` route, so **a single service owns `document[data-theme]`** (no two theme owners fighting over it). This respects the lib boundary — `agentos-ui` keeps no dependency on `apps/client` services (per `GUIDELINES.md`).
- **Appearance control.** Light / Dark / System segmented control in `UserProfileComponent` (`/agentos/me`).
- **Token completion / polish.** Add the missing dark tokens to `apps/client/.../styles/colors.scss` (`--color-text-tertiary`, `--color-surface-hover/-secondary/-muted`, `--color-success-bg`) and document them in `_contract.scss`; replace hardcoded `rgba()` colors in `case-chat` / `case-home` / `case-drawer` with tokens so dark mode renders cleanly across the section.

## Testing

- `nx lint agentos-ui`, `nx test agentos-ui` (59 tests, incl. new `theme.service.spec.ts`), `nx build client` — all green; eager bundle unchanged (theme port wiring stays in the lazy `/agentos` chunk).
- Manual: `/agentos/me` → **Appearance** → switch Light / Dark / System. Change is immediate app-wide, persists across reload, and **System** follows the OS preference (live).


libs/agentos-ui/src/lib/services/theme.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.applyTheme(this.subject.value)
+    this.media?.addEventListener('change', this.onSystemChange)
+  }
+
+  ngOnDestroy(): void {
+    this.media?.removeEventListener('change', this.onSystemChange)
+  }
+
+  getCurrentTheme(): ThemeMode {
+    return this.subject.value
```
mathieu-dearmey: getCurrentTheme() forces consumers to keep an imperative access to the value. Exposing a Signal<ThemeMode> instead of Observable + getter would be cleaner on the consumer side.

libs/agentos-ui/src/lib/components/user-profile/user-profile.component.html
```diff
@@ -115,6 +115,27 @@ <h2 class="user-profile__section-title">Profile</h2>
           </button>
         </div>
 
+        <div class="user-profile__section">
+          <h2 class="user-profile__section-title">Appearance</h2>
```
mathieu-dearmey: "Appearance" is inside the @else (read mode) block. Is it intentional that it disappears when editing?

libs/agentos-ui/src/lib/components/user-profile/user-profile.component.html
mathieu-dearmey: French strings (Mes configurations utilisateur globales, Crée-en depuis les pages…) in an otherwise English UI.

libs/agentos-ui/src/lib/services/theme.service.ts
selimbensenouciep-prog: Good call, done in `2c3ea1a0`. `ThemePort` now exposes a single `theme: Signal<ThemeMode>` and drops both `currentTheme$` and `getCurrentTheme()`. The profile page reads the signal directly, so the `toSignal(...)` bridge in the component is gone too.

On the host side I kept the client `ThemeService` untouched: it still exposes its Observable API for the legacy sidebar and options-panel consumers. Instead of `useExisting`, the `/agentos` route now provides `THEME_PORT` through a small adapter (`useFactory` returning `ThemePort`) that wraps `currentTheme$` into a signal with `toSignal`. It's typed as `ThemePort`, so conformance is checked at compile time, and since it's built lazily on entering `/agentos` the persisted theme is already resolved (no initial-value flash).

libs/agentos-ui/src/lib/components/user-profile/user-profile.component.html
selimbensenouciep-prog: Not intentional, good catch. Fixed in `2c3ea1a0`: the Appearance section is now mode-independent. I moved it out of the read-mode `@else` to a stable position above the profile content, so the theme selector stays visible and editable in both read and edit modes instead of disappearing while editing.

libs/agentos-ui/src/lib/components/user-profile/user-profile.component.html
selimbensenouciep-prog: Done in `2c3ea1a0`. These strings actually predate this PR (they come from the Story 6.6 user-global overrides section), but since I was in the file I translated them: "Mes configurations utilisateur globales" became "My global user configurations", and the empty-state line is now "No global configuration yet. Create one from the Integrations / AI Providers / AI Models pages of your namespaces." I also updated the matching French JSDoc comment in the component.

vincent-audibert-whoz (approved):

Issue #1014, opened 2026-06-16
AgentOS Case management (rename, star, delete)


vincent-audibert-whoz:
Users should be able to : 
- rename a case
- star a case
- soft-delete a case

Issue #1010, opened 2026-06-16
AgentOS case automatic naming


vincent-audibert-whoz:
For Cases that are viewed by a user, have an automatic naming triggered at some points such as : 
- on first user message (mandatory)
- on second user message (nice to have to refine the title)

Implies a CaseUpdatedEvent sent for frontend update, and technical Cases (if any) to skip.

PR #1081, opened 2026-07-02
feat: #1014 add soft-delete action to AgentOS case list

selimbensenouciep-prog:
## What

First slice of #1014 (**AgentOS case management**): the **soft-delete** action in the case drawer.

Each case row now has a delete button. It soft-deletes the case through the **existing** `DELETE /api/cases/{id}` endpoint (which already flips a `removed` flag server-side), so this PR is front-only.

## How

- `CaseDrawerComponent` (presentational) gets a `deleteRequested` output. The row template is restructured into a flex container with a **select button** and a **danger `ds-icon-button`**, avoiding a `<button>` nested inside a `<button>`.
- `CaseShellComponent` (smart) handles the flow in `onDeleteRequested`: confirm (naming the case when it has a title), call `deleteCase`, navigate away if the **active** case was deleted, then refresh the list. Delete failures are logged and leave state intact.
- Removed a dead `(itemSelected)` binding on `ds-entity-list` (selection flows through the row's select button) and extracted a `navigateToSectionHome()` helper.

## Tests

- `case-drawer.component.spec.ts`: `deleteRequested` emits the case id.
- `case-shell.component.spec.ts`: confirm gating, delete + refresh, navigate-away-when-active, no navigate/refresh + log on failure, and the confirmation label (titled vs generic).

Verified: `nx test agentos-ui` (80/80), `nx lint agentos-ui`, and `nx build-angular client` (strictTemplates).

## Out of scope (rest of #1014)

- **rename**: coupled to the automatic case-naming work (#1010, needs a `titleManuallySet`-style guard) and will ship with that PR.
- **star / favorite**: separate PR (edge property on the user/case relation, toggle endpoint, favorite attribute on the list DTO + client regen).

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.html
+++ b/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.html
@@ -7,7 +7,6 @@
   searchPlaceholder="Filter cases…"
   cardMinWidth="220px"
   contentMaxWidth="100%"
-  (itemSelected)="onItemSelected($event)"
   (createRequested)="onCreateRequested()"
 >
   <!-- Inject the close button into the toolbar-start slot, before the title -->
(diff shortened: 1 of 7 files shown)


selimbensenouciep-prog:
Superseded by #1084, which combines soft-delete + star/favorite + related-only listing into a single PR for #1014 (both touch the same drawer/shell files, so they are merged rather than stacked).

PR #1084, opened 2026-07-02
feat: #1014 AgentOS case management (soft-delete + star/favorite + related-only listing)

selimbensenouciep-prog:
## What

Single PR for the **#1014 AgentOS case management** work, combining three pieces that all touch the same drawer/shell surface:

1. **Soft-delete** a case from the drawer (reuses the existing `DELETE /api/cases/{id}` soft-delete).
2. **Star / favorite** a case (per-user), persisted and surfaced in the list.
3. **Listing** restricted to the cases the current user has a **direct** relation with (so every listed case is starrable).

> Supersedes #1081 (soft-delete-only), which is closed in favour of this combined PR. **Rename** stays out of scope (ships with the automatic case-naming work #1010).

## How

**Backend (Kotlin / Neo4j)**
- `starred` boolean on the existing `(User)-[:ADMIN|MEMBER]->(Case)` relation, exposed via `PermissionRepository` → `PermissionService` (mirrors grant/revoke; no cache).
- `CaseController`: `PUT`/`DELETE /api/cases/{id}/star` (READ-gated); `CaseResource.favorite` populated via a `listStarredEntityIds` companion query (`withFavorites` helper, no N+1).
- New `GET /api/cases/by-parentId/{parentId}/mine` → only the current user's direct `[:ADMIN|MEMBER]` cases in the namespace (reuses `findConcerningUserInNamespace`; no admin fast-path / transitivity). Single-case URL access is unchanged, so admins can still open a case they don't own.

**Client**: regenerated OpenAPI client (`Case.favorite`, `starCase`/`unstarCase`, `listMineByParentCase`).

**Frontend (Angular)**
- Drawer row restructured into a flex container: select button + **star** toggle (`star`/`star_border`) + **danger delete** icon-button; favorites sorted first and grouped under a "Favorites" accordion.
- `CaseShellComponent`: `onDeleteRequested` (confirm using the same label the row shows, soft-delete, refresh, navigate-away-if-active, **alert on failure**), `onStarToggled` (star/unstar + refresh), list source = `listMineByParentCase`.
- Removed the dead `CaseListComponent`.

## Behavior notes
- A namespace-admin who owns no case in a namespace now sees an **empty drawer** there (the point of the related-only listing); single-case open-by-URL is unchanged.
- Delete failures are surfaced via a native `alert` (consistent with the native `confirm`); no toast infra added.

## Tests
- Backend `nx :agentos-service:test` → **1414 passed / 0 failed**.
- Frontend `nx test agentos-ui` (+ design-system) → **87 passed**; `nx lint`; `nx build-angular client` (strictTemplates) green.


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
vincent-audibert-whoz, line 1: Check in all other relation handling or update that the relation with its attributes is kept and not deleted-recreated

vincent-audibert-whoz: check done on the sync user. Broader topic about relation lifecycle at repo level and the API exposed, for later...

vincent-audibert-whoz (dismissed):

vincent-audibert-whoz (approved):

Issue #1054, opened 2026-06-25
AgentOS file service for cases and namespaces


vincent-audibert-whoz:
## Context

Coday supports file exchange between users and agents via a case-scoped folder (exchange area). AgentOS needs a proper file service layer to support:
- File uploads into a case (user drops a file, it becomes available to agents)
- A namespace-level document repository (documents injected into agent instructions)
- A pluggable storage backend (local disk for Coday dev use, replaceable for production)

## Goal

Design and implement a file service that:
- Stores files with a reference in the database (path or storage key + metadata)
- Scopes files at two levels: **case** (ephemeral, user uploads) and **namespace** (persistent, document repository)
- Exposes a simple interface so the storage backend can be swapped (local filesystem for Coday, S3/blob for production)
- Is used by the exchange area feature and the agent document injection feature

## Design notes

From the Coday convergence discussion:
- Start with a simple implementation: store files on a mounted disk, keep a DB reference to the location
- The interface must be pluggable so a secure/cloud backend can be substituted later without changing consumers
- Avoid over-engineering: the Coday use case (local dev) needs something simple that works; production hardening (quotas, abuse prevention, encryption) comes later

## Acceptance criteria

- [ ] A file can be uploaded in the context of a case and retrieved by agents in that case
- [ ] A namespace-level document can be stored and referenced by agent configs (used by #1013)
- [ ] The storage backend is behind an interface (at least one implementation: local filesystem)
- [ ] File metadata (name, size, mime type, upload timestamp, scope) is persisted in the database
- [ ] Files are not directly exposed without access control (case membership or namespace membership required)

## Related

- #1013 — AgentOS agent document injection (consumer of namespace-level file storage)
- #1022 — AgentOS image upload (consumer of case-level file storage)

vincent-audibert-whoz:
Closing, only database agent referencing namespace files remains, covered by #1013

Issue #1013, opened 2026-06-16
AgentOS agent document injection


vincent-audibert-whoz:
Add the capability to inject the content of documents into the agent's instructions. Port in a way the feature present on Coday to:
- inject a specific document `/path/to/document.txt`
- inject the list of folders `/path/to/folder/`
- inject all readable documents in a folder: `/path/to/folder/*`

Implies that local agents are resolving that relative to the optional namespace config path.

For database agents:
- depend on a file service for hosting document repository tied to a namespace
- could reference local documents if optional config path on namespace

vincent-audibert-whoz:
Depends on #1054 (file service for cases and namespaces) for the namespace-level document storage backend.

vincent-audibert-whoz:
Local yaml agents covered with a `docs` property, the database agents remain to do by leveraging the incoming namespace-exchange. AgentOS UI needed too for the database agents.

vincent-audibert-whoz:
Next steps covered by #1185

PR #1089, opened 2026-07-06
feat: #1013 filesystem agent document injection

vincent-audibert-whoz:
Implements document injection for filesystem-backed agents (issue #1013), scoped to agents defined via YAML files only.

## What changed

**`AgentConfig`** — one new optional field `docs: List<String>?` — paths whose full content is injected into instructions

**`FilesystemAgentConfigRepository`** — maps `docs` from the YAML model into `AgentConfig`. 

**`AgentDocumentResolver`** — new Spring `@Component` that resolves and formats the docs block. Supports the same three path patterns as Coday:
- explicit file path → content injected verbatim
- path ending with `/` → first-level directory listing, no content
- path ending with `/*` → all readable files in directory, content injected

All paths are resolved relative to the namespace `configPath`. Missing files are logged and skipped — a broken doc reference never aborts the agent.

**`AgentServiceImpl`** — `buildInstructions()` now calls `AgentDocumentResolver.buildDocsBlock()`. The `namespace` object is fetched once in `resolveAgentDefinition()` and passed down to both `buildNamespaceSystemPrompt()` and `buildInstructions()`, eliminating a redundant `findById` call in the process.

## What is NOT implemented

Database-backed agents (`AgentConfig` entities stored in Neo4j) do not yet support document injection — that depends on #1054 (file service for namespace document storage).

## YAML example

```yaml
name: My Agent
instructions: You are a helpful assistant.
[...]
docs:
  - docs/architecture.md
  - docs/guidelines/*
  - context/
```


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
```diff
@@ -140,6 +140,10 @@ class FilesystemAgentConfigRepository(
             modelName = model.modelName,
             integrations = model.integrations,
             subAgents = model.subAgents?.filter { it.isNotBlank() }?.takeIf { it.isNotEmpty() },
+            docs = (model.docs ?: model.mandatoryDocs)
+                ?.filter { it.isNotBlank() }
+                ?.map { entry -> file.parent.resolve(entry).toAbsolutePath().normalize().toString() }
```
selimbensenouciep-prog: **The directory-listing pattern (a path ending with `/`) can never fire, because the trailing slash is destroyed on this line.** `resolve(...).toAbsolutePath().normalize().toString()` strips trailing slashes, so:

- `./libs/` is stored as `<base>/libs` (no trailing slash)
- `./apps/client/docs/*` stays `<base>/apps/client/docs/*` (this only survives because `*` is an ordinary path segment that `normalize()` leaves alone)

Downstream, `AgentDocumentResolver.resolveEntry` picks the pattern from the string suffix (`endsWith("/")` vs `endsWith("/*")`). Since no stored entry can end with `/` anymore, `resolveDirectoryListing` is unreachable: a `/`-terminated entry falls into the single-file branch, fails `isRegularFile()` on a directory, gets logged as "Not a regular file", and is skipped. The listing silently yields nothing.

This is not hypothetical: `Frontay.yaml` lists `./libs/` in `mandatoryDocs`, so that entry is silently dropped as soon as this ships. Verified on JDK 25 that `Paths.get("/base").resolve("libs/").normalize().toString()` returns `/base/libs`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentDocumentResolver.kt
```diff
@@ -0,0 +1,129 @@
...
+            resolved.forEach { (label, content) ->
+                appendLine()
+                appendLine("File: $label")
+                appendLine()
+                appendLine(content)
+            }
+        }.trimEnd().takeIf { it.isNotBlank() }
+    }
+
+    private suspend fun resolveEntry(entry: String): List<Pair<String, String>> =
```
selimbensenouciep-prog, line 70: There's no test coverage on this component. It carries a fair amount of logic (the three path patterns, the sorting, the parallel reads, the skip-on-error), and the only test change in the PR is a relaxed mock, so `buildDocsBlock` returns null and the behaviour isn't exercised.

A few cases would cover it: the three patterns, a missing file and a missing directory, and an assertion that the resolved block lands in the final instructions. The listing case in particular would have caught the trailing-slash issue flagged above.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
```diff
@@ -87,4 +87,27 @@ data class AgentConfig(
...
+     *
+     * For filesystem-backed agents, entries are stored as absolute paths resolved at
+     * load time relative to the agent YAML file. Three path patterns are supported:
+     * - explicit file path: single file, content injected verbatim
+     * - path ending with slash: directory listing (first-level only, no content)
+     * - path ending with slash-star: all readable files in the directory, content injected
+     *
+     * Always null for Neo4j-persisted agents.
+     */
+    val docs: List<String>? = null,
```
selimbensenouciep-prog, line 106: The `docs` field has two KDoc blocks stacked one on top of the other. Only the second attaches to the field, the first is left dangling. They also don't fully agree: the first says the field is "silently ignored when configPath is absent", the second says it's "always null for Neo4j-persisted agents". Probably worth keeping just one.

agents/Frontay.yaml
```diff
@@ -5,6 +5,7 @@ integrations:
   FILES:
   AI:
   MEMORY:
+  JIRA:
```
selimbensenouciep-prog: This JIRA integration looks unrelated to the document injection feature. Could it go in its own change so the PR stays scoped to #1013?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
selimbensenouciep-prog: The file plugin already has a `BoundaryPathResolver` for exactly this: it rejects `..` traversal, keeps resolution inside a root, validates symlinks, and blocks a deny-list of secret files (`.env`, `*.key`, `*.pem`, `id_rsa`, `credentials.json`, and so on). This doc path doesn't go through it. The repository just does `resolve(entry).normalize()` and `AgentDocumentResolver` reads the file directly, so `../../..` still escapes the config dir, and a `dir/*` over a folder that happens to hold a `.env` or a `*.key` would inject its content straight into the agent's instructions.

The YAML is admin-authored so it's not wide open, but reusing that resolver (or at least its boundary check and deny-list) would keep this in line with the file tools and avoid accidentally surfacing secrets.

selimbensenouciep-prog (dismissed):
LGTM. Please fix the trailing-slash directory-listing bug before merging (details inline), it silently breaks `./libs/` in Frontay.yaml. The other comments are non-blocking.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
vincent-audibert-whoz: cleaned

agents/Frontay.yaml
vincent-audibert-whoz: removed

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentDocumentResolver.kt
vincent-audibert-whoz: tests added 👍

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
vincent-audibert-whoz: corrected.

vincent-audibert-whoz: The yaml-based agent definitions are from trusted users, this check seems overkill here for a "non existing" security risk.

fdelsert-whoz (approved):

Issue #1023, opened 2026-06-16
AgentOS ui revamp navigation and cross-page links


vincent-audibert-whoz:
All current navigation has to go from the namespace list, with very few cross-page navigation (eg: AiProvider <-> AiModel).

PR #1091, opened 2026-07-06
#1023 new navigation on coday part signal migration

benjamin-valdes-whoz:


vincent-audibert-whoz (approved):
Would be nice to have an update in `angular-guidelines.md` for that as well

PR #1092, opened 2026-07-06
feat: wz-32968 prompt call

fdelsert-whoz:
The PR introduces a **prompt command system** for AgentOS cases. Users can now type `/prompt-name [args]` in a message to invoke a named prompt template, which expands into one or more resolved text commands before being sent to the agent.

**Core mechanism — `PromptCommandParser`**

`PromptCommandParser` resolves slash-command invocations against a snapshot of effective prompts. It supports:
- `{{ARGUMENTS}}` — replaced with the raw argument string verbatim
- Named positional parameters `{{paramName}}` — mapped by declaration order, with per-parameter defaults
- Quote-aware tokenization for multi-word arguments
- Recursive resolution: a prompt line that itself starts with `/` is resolved depth-first
- Cycle detection via a call stack of `(name, resolvedArgs)` pairs — same prompt with different arguments is allowed (sibling reuse), same prompt with same arguments in ancestry is rejected
- Depth cap at 10 to catch long chains that bypass argument-based cycle detection

**Integration strategy in `CaseServiceImpl.addMessage`**

When a message contains a slash-command, the parser expands it into a list of resolved strings. Each resolved string becomes a **separate sequential agent turn**:
- The first resolved string is sent immediately as the user message
- Subsequent strings are enqueued in `CaseRuntime.commandQueue`
- After each agent turn completes (`AgentFinishedEvent`), `runTurns` drains one command from the queue, injects it as a new user message via `addUserMessage`, and resets `iterationCount` — triggering the next agent turn

Resolution failures (cycle, depth exceeded, missing required arguments) are caught as `PromptResolutionException`, surfaced as a `WarnEvent` on the SSE flow, and the original raw message is stored so conversation history remains intact. The case then proceeds normally to `IDLE`.

Capture d’écran 2026-07-07 à 10 59 22


libs/agentos-api-client/src/custom/prompt-controller.service.ts
```diff
@@ -111,6 +111,21 @@ export class PromptControllerService extends BaseService {
     })
   }
 
+  /**
+   * Returns the resolved set of prompts for the given namespace+user context.
+   * Merges platform, namespace-shared, user-global and user×namespace layers by name.
+   * Used for slash-command autocomplete in the chat composer.
+   */
+  listEffectivePrompts(namespaceId: string): Observable<Prompt[]> {
```
vincent-couturier-whoz: question: why it was automatically generated from the openapi spec ?

vincent-couturier-whoz (dismissed):

libs/agentos-api-client/src/custom/prompt-controller.service.ts
fdelsert-whoz: Yes I wasn't aware of the process when the first PR was submitted.
It needs to be truly generated from openapi spec, and old calls needs to be rewired.

fdelsert-whoz: done at commit: 8c2e4d1

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
```diff
@@ -111,29 +122,34 @@ class CaseRuntime(
...
         killRequested.set(true)
         interruptRequested.set(true)
+        commandQueue.clear()
+    }
+
+    /**
+     * Enqueue a command for sequential execution after the current agent turn.
+     * Used when a prompt resolves to multiple commands.
+     */
+    fun enqueueCommand(content: List<MessageContent>) {
```
vincent-audibert-whoz, line 151: We might need LATER to pass the userId(s) with these MessageContent to avoid the implicit user being the last one => have a ticket to track that hole ?

Idea for LATER: this method could be a subtle alternative to the existing addMessage with a different meaning:
- addMessage = add it right away and let the already running agent have it right away
- enqueueCommand = queue it to wait for the current agent run

vincent-audibert-whoz (approved):
👍 
optional comments/ideas.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseRuntime.kt
fdelsert-whoz: Yes, the idea just below of using a new structure (_PendingCommand_) was precisely to make it extensible in the future so that this kind of information can be added.

I’m not sure I understand the first point about the implicit user.

Yes an enqueueCommand  equivalent could become an entry point of the interface.

PR #1093, opened 2026-07-06
#1023 new navigation on coday part navigation

benjamin-valdes-whoz:
image
image


libs/agentos-ui/src/lib/components/case-shell/case-shell.component.html
vincent-audibert-whoz, line 1: Let’s go but that component seems too big now, to breakup later

vincent-audibert-whoz (approved):

libs/agentos-ui/src/lib/components/case-shell/case-shell.component.html
benjamin-valdes-whoz: ok, i will cut out in small parts

Issue #1019, opened 2026-06-16
Agentos UserGroup UI


vincent-audibert-whoz:
Scoped to a namespace, have UserGroup pages to:
- list/view
- create/edit
- in the view or edit, have a list of deployed agents and a form to edit the deployed agent from a list of agentconfig that are available (platform or belonging to the namespace) and not deployed already elsewhere (on namespace).

PR #1107, opened 2026-07-07
feat: #1019 Agentos UserGroup UI + admin Users & Groups console

selimbensenouciep-prog:
Implements #1019 — a UI to manage UserGroups, plus a small backend addition to support it.

## Backend

- New `GET /api/user-groups/{id}/members` returning each member with their role (`ADMIN`/`MEMBER`), guarded by `READ` on the group.
- A member's role is the User↔UserGroup permission relation. Create/update gain an `adminExternalIds` field; on submit the service reconciles the group's `MEMBER`/`ADMIN` edges to match it. `removeUsers` now unlinks both relation types so removing an admin fully detaches them.
- Covered by service unit tests, Neo4j persistence tests (role read, promote/demote, unlink) and controller integration tests.

## Frontend

- Namespace-scoped user groups (list / create / edit), reachable from the namespace ⋮ menu. The form has:
  - a deployed-agent picker (platform + namespace agents, disabling any already deployed to another group in the namespace),
  - a member list with a per-member `MEMBER`/`ADMIN` role and a type-ahead to add members from the namespace's users.
- `/admin/users` becomes a consolidated **Users & Groups** console with `Users | Groups` tabs. Each toolbar carries a namespace filter on the search line:
  - Users: narrow the user list to a namespace's users (or all),
  - Groups: pick the namespace whose groups to manage.
  - The filter only lists namespaces the user administers (super-admins see all).
- The group form returns to where it was opened from (the namespace page or the admin Groups tab) via a `returnTo` param.
- `ds-entity-list` gains a `hideTitle` option (the tabs label each section) and its toolbar can align to the content width.

## Notes

- The "an agent is deployed to at most one group per namespace" rule is enforced in the UI only; the backend does not reject it. Left as a follow-up.
- Group `ADMIN` grants manage rights on that group only — no transitive/namespace effect — consistent with the RBAC model.

## Testing

- `nx test agentos-service` (UserGroup service / controller / persistence specs), `nx test agentos-ui`, `nx lint`, `nx build client`.
- Manually: namespace ⋮ → User Groups, and ⚙️ Admin → Users & Groups — create/edit a group, set a member's role and confirm it persists on reload, filter users and groups by namespace.


vincent-audibert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupNodeNeo4jRepository.kt
```diff
@@ -55,12 +55,46 @@ interface UserGroupNodeNeo4jRepository : Neo4jRepository<UserGroupNode, String>
...
+     */
+    @Query(
+        $$"""
+        MATCH (u:User)-[a:ADMIN]->(g:UserGroup {id: $groupId})
+        WHERE NOT u.externalId IN $adminExternalIds
+        MERGE (u)-[:MEMBER]->(g)
+        DELETE a
+        """,
+    )
+    fun demoteNonAdmins(
```
vincent-audibert-whoz: Seems very brutal, really needed ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
```diff
@@ -360,6 +360,41 @@ class Neo4jPermissionRepository(
...
         }
 
+    override fun applyShareBatch(
+        entityType: EntityType,
+        entityId: String,
+        entries: List<Pair<String, PermissionRelation?>>,
+    ): List<String> {
+        if (entries.isEmpty()) return emptyList()
+        val label = entityType.label
+        val result = mutableSetOf<String>()
```
thomasmartin-whoz: this code could work without mutation

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
```diff
@@ -342,6 +342,91 @@ interface PermissionNodeNeo4jRepository : Neo4jRepository<UserNode, String> {
...
+            DELETE oldMember
+            MERGE (u)-[:ADMIN]->(e)
+            RETURN uid
+            """,
+    )
+    fun batchGrantAdmin(
+        @Param("userIds") userIds: List<String>,
+        @Param("entityId") entityId: String,
+        @Param("entityLabel") entityLabel: String,
+    ): List<String>
```
thomasmartin-whoz, line 389: shouldn't ADMIN also be MEMBER ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
```diff
@@ -227,6 +227,24 @@ class PermissionServiceImpl(
...
+    override fun applyShareBatch(
+        entityType: EntityType,
+        entityId: String,
+        entries: List<Pair<String, PermissionRelation?>>,
+    ): List<String> {
+        if (entries.isEmpty()) return emptyList()
+        return try {
+            val result = permissionRepository.applyShareBatch(entityType, entityId, entries)
+            // Sharing affects multiple users — invalidate the whole cache.
+            permissionCache.clear()
```
thomasmartin-whoz, line 239: that's a complete other topic, but this permission cache seems a bad idea

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
```diff
@@ -51,6 +51,33 @@ open class Neo4jUserGroupRepository(
             paramValue = id.toString(),
         ).one().orElse(null)
 
+    override fun findMembers(userGroupId: UUID): List<UserGroupMember> =
```
thomasmartin-whoz, line 58: a @Query annotation does not work ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupServiceImpl.kt
```diff
@@ -153,6 +195,54 @@ class UserGroupServiceImpl(
...
+        }
+    }
+
+    private fun validateAdminsAreMembers(
+        adminExternalIds: Set<String>,
+        memberExternalIds: Set<String>,
+    ) {
+        val notMembers = adminExternalIds - memberExternalIds
+        if (notMembers.isNotEmpty()) {
+            throw UnprocessableEntityException("Admin external IDs must also be members: $notMembers")
```
thomasmartin-whoz: here there is a control on admin that must be members.
Depending of the classes, admins are expected to be members or not, it is really unclear and unconsistent.

thomasmartin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/Neo4jPermissionRepository.kt
selimbensenouciep-prog: Replaced the mutable accumulator with a `flatMap` over the entries grouped by relation, plus `distinct()`. Batch order now follows first occurrence in `entries` instead of a fixed admin/member/revoke, equivalent since the service rejects overlapping sets. Done in 3ab457d1.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionNodeNeo4jRepository.kt
selimbensenouciep-prog: By design a user holds a single relation per entity (the edge is the role), so an ADMIN needs no parallel MEMBER edge; every membership read traverses `[:MEMBER|ADMIN]`. Your comment did surface a real hole though: `createAdminPermission`/`createMemberPermission` MERGEd without deleting the opposite edge, so `PUT /admins/{uid}` on an existing member left both edges, and offboarding then only dropped `[:ADMIN]`. Fixed in 3ab457d1: `create*` now replace the opposite edge, `batchGrant*` renamed to `batchSetAdminRole`/`batchSetMemberRole`, invariant documented on `PermissionRelation` + `entity-schema.md`, covering tests added.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/permissions/PermissionServiceImpl.kt
selimbensenouciep-prog: Agreed it's weak (5 min TTL, full `clear()` on every write, a hit still pays the per-call user read). This batch layer and its `clear()` come from Vincent's #1142 (case sharing), pulled in here to ease the merge, so it follows the existing invalidation pattern rather than a new one. Leaving the rework out of scope; we'll flag it against #1142 when it lands. One in-scope fix in 3ab457d1: added members now invalidate their cache entry too (`updateFromRequest` only cleared on removal).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/Neo4jUserGroupRepository.kt
selimbensenouciep-prog: Doesn't fit here: the query returns a multi-column row including a computed `role`, which SDN can't map onto a non-entity DTO (it rejects the multi-column RETURN, same limitation documented on `CaseNodeNeo4jRepository.findDirectRelations`). Kept `neo4jClient`, consistent with the two sibling projections in the same class, and added a comment in 3ab457d1.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/userGroup/UserGroupServiceImpl.kt
selimbensenouciep-prog: Naming was the problem, "member" was overloaded. Renamed to `validateAdminsBelongToGroup`: it checks roster membership (a designated admin must be in the group, holding either edge), reworded the message and added a KDoc (3ab457d1). On "depending on the classes": the graph model is uniform (ADMIN XOR MEMBER everywhere), only the request shape differs (UserGroup is a roster plus admin sublist, Namespace/Case assign a single role per user).

PR #1110, opened 2026-07-08
fix: #33235 platform-level entities visible and read-only in namespace views

vincent-audibert-whoz:
### Problem

Platform-level entities (agent configs, AI models, AI providers) were invisible to namespace admins. Only namespace-scoped entities were shown in the listing views.

### Solution

Fetch both namespace-level and platform-level entities in parallel and merge them into grouped `ds-entity-list` views. Platform entries appear first in a dedicated "Platform (read-only)" group with all mutation actions hidden.

#### `namespace-agent-configs`
- Parallel fetch via `forkJoin`: namespace configs + platform configs (`listPlatformAgentsAgentConfig`)
- Two separate Maps to track origin; `isPlatformConfig(id)` drives `[readOnly]` on the item

#### `agent-config-item`
- New `readOnly` input: when true, only the "Inspect" action is available (edit/delete hidden)

#### `namespace-ai-models`
- Parallel fetch via `forkJoin`: namespace models + platform models (`listPlatformLevelAiModel`) + providers
- Platform models grouped first; `isPlatformModel(id)` drives `[readOnly]` on the item

#### `ai-model-item`
- New `readOnly` input: when true, the kebab menu is hidden entirely

#### `ai-providers-all-scopes` + `AiProviderConfigStateService`
- `AiProviderScope` extended with `'platform'`
- `AiProviderConfigViewModel` gains a `platform` field
- `vm$` adds a 4th parallel call (`listAiProvider('none')`) for platform providers
- Platform section rendered first; `readOnly` applies on `scope === 'platform'`

#### `ai-provider-item`
- `SCOPE_BADGES` completed with `platform` entry

#### `ai-provider-form`
- `SCOPE_LABEL` completed with `platform` (exhaustive `Record<AiProviderScope, string>`)
- `VALID_SCOPES` does **not** include `platform` — it is never selectable in a form

#### `prompt`
same: show platform entities as read-only on the namespace dedicated page.

### DIFF
--- a/libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
+++ b/libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
@@ -20,6 +20,11 @@ import { IconButtonComponent, KebabMenuComponent, KebabMenuItem } from '@whoz-os
 export class AiModelItemComponent {
   readonly model = input.required<AiModel>()
   readonly namespaceId = input<string | undefined>(undefined)
+  /**
+   * When true, edit and delete actions are hidden.
+   * Used for platform-level models displayed in a namespace context (read-only visibility).
+   */
+  readonly readOnly = input(false)
 
   readonly deleteRequested = output<AiModel>()
   readonly editRequested = output<AiModel>()
@@ -30,6 +35,8 @@ export class AiModelItemComponent {
     { key: 'edit', label: 'Edit model', icon: 'edit' },
     { key: 'delete', label: 'Delete model', icon: 'delete', variant: 'danger' },
   ]
+
+  protected readonly readOnlyMenuItems: KebabMenuItem[] = []
 
   protected get displayTitle(): string {
     return this.model().alias ?? this.model().apiModelName
--- a/libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
+++ b/libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
@@ -36,8 +36,6 @@ export class AiModelItemComponent {
     { key: 'delete', label: 'Delete model', icon: 'delete', variant: 'danger' },
   ]
 
-  protected readonly readOnlyMenuItems: KebabMenuItem[] = []
-
   protected get displayTitle(): string {
     return this.model().alias ?? this.model().apiModelName
   }
--- a/libs/agentos-ui/src/lib/components/namespace-agent-configs/namespace-agent-configs.component.ts
+++ b/libs/agentos-ui/src/lib/components/namespace-agent-configs/namespace-agent-configs.component.ts
@@ -4,14 +4,19 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop'
 import { ActivatedRoute, Router } from '@angular/router'
 import { AgentConfig, AgentConfigControllerService } from '@whoz-oss/agentos-api-client'
 import { EntityListComponent, EntityListItem } from '@whoz-oss/design-system'
-import { BehaviorSubject, map, switchMap } from 'rxjs'
+import { BehaviorSubject, forkJoin, map, switchMap } from 'rxjs'
 import { AgentConfigItemComponent } from '../agent-config-item/agent-config-item.component'
+
+const GROUP_NAMESPACE = 'namespace'
+const GROUP_PLATFORM = 'platform'
 
 /**
  * NamespaceAgentConfigsComponent — list view for agent configs of a namespace.
  *
  * Loaded at /:namespaceId/agent-configs. Responsibilities:
- * - Load and display the list of AgentConfig via ds-entity-list
+ * - Load and display namespace-level AND platform-level agent configs
+ * - Merge both levels into a grouped ds-entity-list (namespace / platform)
+ * - Platform-level configs are displayed read-only (no edit/delete actions)
  * - Navigate back to the namespace list
  * - Navigate to the create form (/:namespaceId/agent-configs/new)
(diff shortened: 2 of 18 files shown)


libs/agentos-ui/src/lib/components/namespace-agent-configs/namespace-agent-configs.component.ts
```diff
@@ -35,30 +40,52 @@ export class NamespaceAgentConfigsComponent {
...
+  /**
+   * Fetches both namespace-level and platform-level configs in parallel.
+   * Platform configs are appended with a groupKey so ds-entity-list renders them
+   * in a separate collapsed-by-default group.
+   */
+  private readonly allConfigs$ = this.refresh$.pipe(
+    switchMap(() =>
+      forkJoin({
+        namespace: this.agentConfigController.listByNamespaceAgentConfig(this.namespaceId),
+        platform: this.agentConfigController.listPlatformAgentsAgentConfig(),
```
vincent-audibert-whoz: **Fail-fast with no recovery.** `forkJoin` errors immediately if any inner observable errors. If `listPlatformAgentsAgentConfig()` returns a 4xx/5xx (a namespace admin may lack permission to list platform configs), the whole stream errors and the page goes blank — namespace configs are never displayed.

The provider state service handles this correctly with per-source `catchError(() => of([]))`. Apply the same pattern here:
```ts
forkJoin({
  namespace: this.agentConfigController.listByNamespaceAgentConfig(this.namespaceId),
  platform: this.agentConfigController.listPlatformAgentsAgentConfig().pipe(catchError(() => of([] as AgentConfig[]))),
})
```
Same issue applies to `namespace-ai-models` (platform call) and `namespace-prompts` (platform call).

libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
```diff
@@ -31,6 +36,8 @@ export class AiModelItemComponent {
     { key: 'delete', label: 'Delete model', icon: 'delete', variant: 'danger' },
   ]
```
vincent-audibert-whoz, line 37: **Dead code.** `readOnlyMenuItems` is declared but never used — the template hides the kebab menu entirely via `@else if (!readOnly())` rather than switching item lists. Remove the field.

libs/agentos-ui/src/lib/components/prompt-item/prompt-item.component.ts
```diff
@@ -24,6 +24,11 @@ export class PromptItemComponent {
   @Input() namespaceId?: string
   /** When true, edit navigates to the admin platform route instead of the namespace route. */
   @Input() platformMode = false
+  /**
+   * When true, edit and delete actions are hidden.
```
vincent-audibert-whoz, line 31: **Angular guidelines inconsistency.** `prompt-item` uses legacy `@Input()` / `@Output()` decorator style while `agent-config-item` and `ai-model-item` (both modified in this same PR) use signal-based `input()` / `output()`. The angular guidelines mandate signal-based APIs for new inputs; migrating `readOnly` to `input(false)` here would make it consistent with the other item components and remove the mixed decorator+signal style within the same file.

vincent-audibert-whoz (commented):
The pattern applied across the four entity types is sound and the approach of identifying platform origin from which endpoint served the data (rather than a flag on the DTO) is the right pragmatic call. The `forkJoin` + two-Maps + `isPlatformXxx(id)` structure is consistent and readable. Three issues need to be addressed before merging.

**Blocking: `forkJoin` fail-fast with no error recovery on namespace-scoped components.** The `ai-provider-config-state.service` correctly wraps each source in `catchError(() => of([]))`, keeping the page functional when one endpoint fails. The three namespace-level components (`namespace-agent-configs`, `namespace-ai-models`, `namespace-prompts`) use a bare `forkJoin` with no error handling. If the platform endpoint returns a 4xx/5xx (e.g. the namespace admin has no permission to list platform configs), the entire page goes blank and namespace configs are never displayed. The platform call should be wrapped in `catchError` the same way the provider service does it.

**Inconsistency: `prompt-item` uses legacy `@Input()` / `@Output()` while sibling components use signal-based `input()` / `output()`.** The PR description notes this was deliberate for consistency "with the existing component", but `agent-config-item` and `ai-model-item` already use `input()` / `output()` in this same PR. `prompt-item` is the odd one out — it mixes the old decorator style with `signal()` for internal state in the same file, which is the inconsistency the guidelines are trying to avoid.

**Minor: dead field `readOnlyMenuItems` in `ai-model-item`.** The field is declared (`protected readonly readOnlyMenuItems: KebabMenuItem[] = []`) but never referenced in the template — the template uses `@else if (!readOnly())` to hide the menu entirely. The declaration is dead code and should be removed.

libs/agentos-ui/src/lib/components/namespace-agent-configs/namespace-agent-configs.component.ts
vincent-audibert-whoz: ok

libs/agentos-ui/src/lib/components/ai-model-item/ai-model-item.component.ts
vincent-audibert-whoz: cleaned.

libs/agentos-ui/src/lib/components/prompt-item/prompt-item.component.ts
vincent-audibert-whoz: migrated

fdelsert-whoz (approved):

PR #1116, opened 2026-07-08
fix: add missing config to start agentos

vincent-audibert-whoz:
... and tweak Archay on the way.


vincent-audibert-whoz:
Fake issue, closing.

PR #1117, opened 2026-07-08
fix: #33273 case search flattens tree and matches on id

vincent-audibert-whoz:
## Problem

Two issues with case search in the sidebar drawer:

1. **Case IDs were not matched** — the search only filtered on the case title, not the ID.
2. **Sub-cases were invisible to search** — the filter operated on root tree nodes only; children at any depth were never reached.

### `ds-entity-list` — migrated to signal-based API + new `disableInternalFilter` input

- All `@Input()` / `@Output()` replaced with `input()` / `output()` (signal-based, Angular 17.3+)
- Removed the `OnChanges` + internal `itemsSignal` bridge — `filteredItems` now reads `this.items()` directly, which is natively reactive
- Added `disableInternalFilter: boolean` input: when `true`, items are displayed as-is (no internal filtering). The search input remains visible and the query is emitted via `searchChanged`.
- Added `searchChanged` output: emits on every keystroke and on clear. Primarily useful when `disableInternalFilter` is true and the parent owns the filter logic.

### `CaseDrawerComponent` — two-mode rendering

- **Tree mode** (no active search): unchanged hierarchical rendering with expand/collapse.
- **Flat mode** (search active): `collectMatching()` traverses the full tree recursively at every depth and returns a flat list of all cases whose **title or ID** matches the query. The tree structure is not preserved in search results.
- Uses `disableInternalFilter="true"` + `(searchChanged)` — the drawer owns the filter state, `ds-entity-list` receives pre-filtered items.
- `buildTree()` now populates the `description` field with the case ID, as a secondary defence so the native filter also matches on ID if `disableInternalFilter` is ever removed.

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
+++ b/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
@@ -163,18 +163,12 @@ function buildTree(cases: Case[]): CaseTreeItem[] {
  * Returns a flat list — hierarchy is not preserved in search results.
  */
 function collectMatching(nodes: CaseTreeItem[], query: string): CaseTreeItem[] {
-  const results: CaseTreeItem[] = []
-  for (const node of nodes) {
-    const matchesName = node.name.toLowerCase().includes(query)
-    const matchesId = node.id.toLowerCase().includes(query)
-    if (matchesName || matchesId) {
-      results.push(node)
-    }
-    if (node.children.length) {
-      results.push(...collectMatching(node.children, query))
-    }
-  }
-  return results
+  return nodes.reduce<CaseTreeItem[]>((acc, node) => {
+    const matches = node.name.toLowerCase().includes(query) || node.id.toLowerCase().includes(query)
+    if (matches) acc.push(node)
+    if (node.children.length) acc.push(...collectMatching(node.children, query))
+    return acc
+  }, [])
 }
 
 function findAndCollectAncestors(nodes: CaseTreeItem[], targetId: string, acc: Set<string>): boolean {
--- a/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
+++ b/libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
@@ -14,7 +14,7 @@ import { NgTemplateOutlet } from '@angular/common'
 
 /**
  * EntityListItem extended with a recursive children list.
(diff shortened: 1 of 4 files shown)


libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
```diff
@@ -122,6 +157,26 @@ function buildTree(cases: Case[]): CaseTreeItem[] {
...
 }
 
+/**
+ * Recursively collect all tree nodes (at any depth) whose name or id
+ * contains the given query string (case-insensitive).
+ * Returns a flat list — hierarchy is not preserved in search results.
+ */
+function collectMatching(nodes: CaseTreeItem[], query: string): CaseTreeItem[] {
+  const results: CaseTreeItem[] = []
+  for (const node of nodes) {
```
vincent-palita-whoz: Using a reduce here is more elegant

vincent-palita-whoz (dismissed):

libs/design-system/src/lib/components/entity-list/entity-list.component.ts
```diff
@@ -45,32 +41,41 @@ export interface GroupedItems {
...
-  @Input() autoFocusSearch: boolean = false
-  @Input() cardMinWidth: string = '260px'
-  @Input() contentMaxWidth: string = '900px'
-  @Input() showCreate: boolean = false
-  @Input() itemTemplate?: TemplateRef<{ $implicit: EntityListItem }>
-
-  @Output() itemSelected = new EventEmitter<string>()
-  @Output() createRequested = new EventEmitter<void>()
+export class EntityListComponent implements AfterViewInit {
+  readonly title = input.required<string>()
```
vincent-couturier-whoz, line 45: praise: ❤️ signal. Daniel presentation has been useful...

vincent-couturier-whoz (dismissed):

libs/agentos-ui/src/lib/components/case-drawer/case-drawer.component.ts
vincent-couturier-whoz: or just a flatMap
`nodes.flatMap(node => node.name.toLowerCase().includes(query) || node.id.toLowerCase().includes(query) ? [node,...collectMatching(node.children, query)]:collectMatching(node.children, query))`

vincent-couturier-whoz (approved):

Issue #1097, opened 2026-07-06
Frontend OAuth popup and QuestionEvent UI for MCP authorization (agentos-ui)


vincent-audibert-whoz:
## Context

Sub-issue of #1018. Depends on:
- MCP OAuth flow backend endpoint (sibling issue #1096) — the callback endpoint must exist before this can be wired

The AgentOS frontend (`agentos-ui`) already subscribes to all case events via SSE, including `QuestionEvent`. However, no UI exists to render it or send back an `AnswerEvent`. Similarly, the OAuth popup flow from Coday has never been ported to `agentos-ui`.

The Coday frontend implementation (in `apps/client`) is the reference, but it accumulated complexity over time. This issue is about porting the essential behavior cleanly, not copying the implementation.

## What exists in Coday (reference)

- **`OAuthService`** — subscribes to `OAuthRequestEvent`, opens a popup window, monitors it, handles `postMessage` from callback, falls back to localStorage, sends `OAuthCallbackEvent` back to backend
- **`OAuthRequestPanelComponent`** — inline panel in the chat input area, shows "Authorize" button when auth is pending
- **`OAuthCallbackComponent`** — route `/oauth/callback`, extracts `code` + `state`, sends to parent via `postMessage` or localStorage fallback
- **Route** — `/oauth/callback` registered in app routes

## What exists in AgentOS (foundation)

- `QuestionEvent` and `AnswerEvent` TypeScript interfaces generated in `libs/agentos-api-client`
- SSE subscription in `CaseChatComponent` already listens for `QuestionEvent` (no rendering yet)
- No `/agentos/oauth/callback` route
- No API endpoint to submit `AnswerEvent`

### 1. `QuestionEvent` rendering and `AnswerEvent` submission

- Render `QuestionEvent` inline in the case chat as an interactive prompt:
  - Free-text input if `options` is null
  - Button group if `options` is a non-empty list
- On user submission, `POST /api/cases/{caseId}/events/answer` with `{ questionId, answer }`
- Add the corresponding backend REST endpoint if it doesn't exist (`AnswerEvent` submission)
- Rendered `AnswerEvent` in the chat timeline (shows what was answered)

### 2. OAuth popup flow

- **`McpOAuthCallbackComponent`** — new Angular component at route `/oauth/callback`:
  - Extracts `code` and `state` from query params
  - Sends to parent window via `window.opener.postMessage(...)` if available
[...]

Issue #1096, opened 2026-07-06
MCP OAuth flow — backend endpoint and plugin integration


vincent-audibert-whoz:
## Context

Sub-issue of #1018. Depends on:
- MCP SDK 2.0 upgrade (sibling issue #1094)
- User-scoped credential storage (sibling issue #1095)

Remote MCP servers using OAuth 2.1 require an interactive flow: the user must visit an authorization URL in a browser, authenticate, and the resulting auth code must be sent back to the MCP client to complete the token exchange.

In Coday, this was implemented with significant complexity: a custom `OAuthProvider`, event emission (`OAuthRequestEvent` → frontend opens popup), a callback route, and a `pendingAuthPromise` resolved when the callback arrived. The implementation worked but accumulated workarounds (scope injection override, manual URL manipulation, SSE fallback complexity).

The MCP Java SDK 2.0 provides cleaner primitives for this: `HttpClientStreamableHttpTransport` with pluggable `httpRequestCustomizer` for token injection, and an elicitation mechanism for interactive URL-based requests. We implement our own OAuth orchestration on top of these primitives, keeping the plugin in control.

### Flow overview

```
1. Plugin attempts to connect to remote MCP server
2. Server returns 401 (or WWW-Authenticate header with resource metadata URL)
3. Plugin performs OAuth discovery (RFC 8414 / RFC 9728)
4. Plugin checks IntegrationConfig for existing valid tokens → skip to step 7 if found
5. Plugin emits an OAuthRequestEvent (via ToolContext) with the authorization URL
6. User completes OAuth in browser popup → auth code sent to AgentOS callback endpoint
7. Plugin exchanges code for tokens (PKCE), saves tokens via IntegrationConfig service
8. Plugin retries connection with Bearer token injected in httpRequestCustomizer
9. MCP server accepts → tools available
```

### Backend changes required

1. **OAuth callback endpoint** in `agentos-service`:
   - `POST /api/mcp/oauth/callback` — receives `{ code, state, integrationName, namespaceId }`
   - Validates state parameter (PKCE anti-CSRF)
   - Routes the callback to the waiting plugin instance (via in-memory registry keyed on `state`)
   - Returns 200 OK (frontend popup can close)

2. **Pending auth registry** in the MCP plugin:
   - `McpOAuthPendingRegistry`: maps `state` → `CompletableFuture<String>` (the auth code)
   - Plugin creates entry before emitting `OAuthRequestEvent`, awaits it
[...]

Issue #1095, opened 2026-07-06
User-scoped credential storage for MCP OAuth tokens in AgentOS


vincent-audibert-whoz:
## Context

AgentOS needs a proper credential management layer to support OAuth flows (MCP remote, future integrations), API keys, and shared/individual authentication. Today, credentials live inside `IntegrationConfig.parameters` in clear text — mixing functional config with secrets, with no access control distinction.

This issue introduces two new entities (`AuthSetting`, `Credential`) and an SDK service (`AuthService`) to separate **auth configuration** (shareable) from **auth tokens** (individual), with encryption at-rest.

---

### AuthSetting

**Purpose**: Describes *how* to authenticate against an external service. Shareable across users.

```kotlin
data class AuthSetting(
    override val metadata: EntityMetadata = EntityMetadata(),
    val namespaceId: UUID? = null,
    val userId: UUID? = null,
    val name: String,
    val authType: AuthType,
    val data: JsonNode,            // encrypted at-rest, masked in API responses
) : Entity

enum class AuthType {
    OAUTH_DISCOVERABLE,   // no clientId needed, server exposes discovery
    OAUTH_REGISTERED,     // clientId + clientSecret + tokenUrl + authorizationUrl
    OAUTH_CUSTOM,         // non-standard OAuth (e.g. Basecamp)
    API_KEY,              // static API key
    BEARER_TOKEN,         // static bearer token
    BASIC_AUTH,           // username + password
}
```

**Scoping**: 4-tier, same precedence as `IntegrationConfig`:
- Platform (`ns=null, user=null`)
- User-global (`ns=null, user=Y`)
- Namespace-shared (`ns=X, user=null`)
- User×Namespace (`ns=X, user=Y`)

Precedence (low→high): **platform < user-global < namespace-shared < user×namespace**.

Same-name AuthSettings at different levels produce **shadowing** (higher-precedence level replaces the lower one for that context), identical to AiProvider behavior.

No deep-merge of `data` , pure entities shadowing the less priority ones for a given name.

**Uniqueness constraint**: `tripleKey` on `(namespaceId, userId, name)` — same pattern as `IntegrationConfig`.

**API exposure**: `data` fields containing secrets are masked in GET responses using the existing `ApiKeyMasking` pattern: first 4 + last 4 chars visible (fewer if short), middle replaced with `****`. The masked value serves as sentinel on PUT (masked → preserve persisted value, null → preserve, empty → clear, non-blank → replace).

### Credential

[...]

PR #1119, opened 2026-07-09
feat: #1018 add MCP_HTTP tool provider for remote MCP servers

### LINKED ISSUE #1018: AgentOS remote MCP plugin
Enable remote MCP use through a plugin.

From the Coday experience, 90% of the difficulty is in handling properly the auth configuration...

vincent-audibert-whoz:
## Summary

Adds remote MCP server support to AgentOS via a new `MCP_HTTP` tool provider, using the MCP Java SDK 2.0 Streamable HTTP transport. This is independent of the #1095 AuthSetting feature — it can be merged directly to master.

Closes #1018 (core transport support).

### Commit 1 — SDK: CredentialProvider + Credential types

Adds the minimal SDK surface needed for plugins to consume credentials:

- **`CredentialProvider`** — `typealias CredentialProvider = () -> Credential?`. A pre-scoped supplier that plugins call without knowing identity resolution details.
- **`Credential`** / **`CredentialType`** — data model for stored authentication results (OAuth tokens, API keys, bearer tokens, basic auth).
- **`ToolContext.credentialProvider`** — new nullable field, defaults to `null`. Fully backward-compatible — existing plugins and service code are unaffected.

The service-side wiring (AuthService, ToolResolverService) that actually populates this field will land with #1095. Until then, `credentialProvider` is always `null`, and the MCP HTTP provider falls back to static `authToken` from the IntegrationConfig.

### Commit 2 — MCP HTTP tool provider

One plugin JAR, two `@Extension` implementations:

| Extension | `integrationType` | Transport | Pooling | Auth |
|---|---|---|---|---|
| `McpToolProvider` | `MCP_STDIO` | Stdio (child process) | ✅ Pooled | N/A |
| `McpHttpToolProvider` | `MCP_HTTP` | Streamable HTTP | ❌ Ephemeral | CredentialProvider → static token → none |

**New files:**
- `HttpMcpConnection.kt` — `McpConnectionPort` impl using `HttpClientStreamableHttpTransport` (in `mcp-core`, no new deps)
- `McpHttpToolProvider.kt` — `@Extension` with HTTP-specific JSON schema and 3-tier bearer token resolution
- `McpToolProvider.kt` — extracted from `McpPlugin.kt` (no logic change)
- Unit tests for config validation and bearer token resolution (7 test cases)

**Key design decisions:**
- **Not pooled**: each HTTP connection is user-scoped (via credential token), so sharing is unsafe. JDK `HttpClient` manages its own TCP pool internally.
- **Bearer token priority**: `CredentialProvider` → static `authToken` → no auth. Fully decoupled from OAuth flow (#1096/#1097).
- **No new dependencies**: `HttpClientStreamableHttpTransport` ships in `mcp-core` 2.0.0 (already bundled).

## Configuration example

Create an `MCP_HTTP` IntegrationConfig with:
```json
{
  \"url\": \"https://mcp.example.com\",
  \"authToken\": \"optional-static-token\",
  \"timeoutSeconds\": 30,
  \"toolCallTimeoutSeconds\": 60
}
```

## Testing

- `nx run agentos-mcp-plugin:build` passes (compile + all tests)
- Bearer token resolution matrix tested: null context, static fallback, credential provider priority, all token key names

## Related issues
- #1018 — MCP remote (this PR)
- #1095 — AuthSetting & CredentialProvider wiring (separate PR, not a dependency)
- #1096 / #1097 — OAuth flow (future, independent)


agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+            return emptyList()
+        }
+
+        val bearerToken = resolveBearerToken(context, serverConfig)
+
+        val connection = HttpMcpConnection(serverConfig)
+        return try {
+            connection.connect(bearerToken)
+            if (connection.tools.isEmpty()) {
+                logger.warn { "MCP_HTTP integration '$configName': server advertises no tools" }
```
vincent-audibert-whoz, line 57: **Connection leak — should fix before merge.**

```kotlin
// The connection stays open for the duration of the agent run.
// It is closed by the GC when the McpTool references are released.
// TODO: consider explicit lifecycle management if connections are observed to leak.
connection.tools.map { McpTool(it, connection, configName) }
```

`HttpMcpConnection` does not implement `Closeable` / `AutoCloseable`, so there is no finalizer and the JVM makes no guarantee about when (or whether) GC will trigger. In practice the connection — and the underlying `JDK HttpClient`'s internal thread pool — will be held open for the lifetime of the `McpTool` references, which are typically held by an agent that may be cached. Under any sustained load this exhausts file descriptors.

The TODO acknowledges the problem but defers it. Because the issue is architectural (no lifecycle owner), it needs to be resolved now rather than observed in production. `HttpMcpConnection.close()` is not surfaced on `McpConnectionPort`, so the caller has no standard way to close it. Either:
- Make `HttpMcpConnection` implement `Closeable` and have the framework close it after the agent run, or
- Add `close()` to `McpConnectionPort` so `McpTool` (or its owner) can close the connection after execution.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        client = McpClient
+            .sync(transport)
+            .requestTimeout(Duration.ofSeconds(clientTimeout))
+            .build()
+
+        try {
+            val initResult = client.initialize()
+            logger.info { "[MCP-HTTP] Connected: ${initResult.serverInfo?.name} ${initResult.serverInfo?.version}" }
+        } catch (e: Exception) {
```
vincent-audibert-whoz, line 80: **Unnecessary cast suppression.**

```kotlin
@Suppress("UNCHECKED_CAST")
val request = CallToolRequest.builder(toolName)
    .arguments(arguments as Map<String, Any>)
    .build()
```

This pattern is copied from `StdioMcpConnection`, but the cast silently drops nullability guarantees on the values. A null-safe transformation (e.g. `arguments.mapValues { (_, v) -> v ?: error("null argument for tool '$toolName'") }` or simply filtering nulls with a documented rationale) is both safer and avoids suppressing the compiler warning. If null values are genuinely impossible at this call site, document why rather than suppressing.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/auth/CredentialProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * A pre-scoped credential supplier.
+ *
+ * Implementations capture the resolved (namespace, user, authSettingName) at construction
+ * time so that plugin code never needs to handle identity resolution or know which
+ * AuthSetting it is bound to.
+ *
+ * Returns null when the user has not yet authenticated for the bound AuthSetting.
+ */
+typealias CredentialProvider = () -> Credential?
```
vincent-audibert-whoz, line 14: **Package cohesion — consider moving to `sdk.credential`.**

This file's only content is a typealias that returns `Credential`. The `auth` package currently contains nothing else (no `AuthSetting`, no `AuthService`, no token types). The typealias' natural home is `sdk.credential` alongside `Credential` and `CredentialType`, which is where a reader would look for it. Keeping it in `sdk.auth` only makes sense if that package is expected to grow into a real auth boundary in `#1095`. If so, a brief comment explaining the forward intent would clarify the decision.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+            // TODO: consider explicit lifecycle management if connections are observed to leak.
+            connection.tools.map { McpTool(it, connection, configName) }
+        } catch (e: Exception) {
+            logger.error { "MCP_HTTP integration '$configName': could not connect — ${e.message}" }
+            connection.close()
+            emptyList()
+        }
+    }
+
+    /**
```
vincent-audibert-whoz, line 73: **Stringly-typed key lookup duplicates `CredentialType` documentation.**

```kotlin
val token = credential.data["accessToken"]
    ?: credential.data["token"]
    ?: credential.data["key"]
    ?: credential.data["apiKey"]
```

These four string literals are the authoritative key names for `CredentialType` values, but they live only in a KDoc comment on `Credential.kt` and here as raw strings. If a new `CredentialType` is added with a different key name, this lookup silently falls through to the static token fallback with no compiler signal.

The canonical approach is to drive the lookup from the type: e.g. add a `fun primaryTokenKey(): String?` (or a companion constant) to `CredentialType`, so `credential.data[credential.credentialType.primaryTokenKey()]` is the lookup. This keeps the key-name contract in one place and makes the fallback explicit only for `BASIC_AUTH` (which genuinely has no single bearer token).

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/credential/Credential.kt
```diff
@@ -0,0 +1,30 @@
...
+ * - [CredentialType.BEARER_TOKEN]: `token`
+ * - [CredentialType.BASIC_AUTH]: `username`, `password`
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class Credential(
+    override val metadata: EntityMetadata = EntityMetadata(),
+    val userId: UUID,
+    val authSettingId: UUID,
+    val credentialType: CredentialType,
+    val data: Map<String, String> = emptyMap(),
```
vincent-audibert-whoz, line 29: **`data` type offers no protection for sensitive material.**

The KDoc states this field is *encrypted at rest*, but the type is a plain `Map<String, String>`. Any `toString()`, JSON serialisation, or log statement on a `Credential` instance will expose raw token values in plaintext. The `resolveBearerToken` logger is careful not to log the token value, which is good — but there is no type-level protection against accidental exposure elsewhere.

Not a blocker for this PR, but worth addressing in `#1095`: a wrapper type (e.g. `SensitiveString` with a redacted `toString()`) or a custom `toString()` override on `Credential` that masks `data` values would prevent credential leakage in logs.

vincent-audibert-whoz (commented):
The structure and intent of this PR are solid — clean split between stdio and HTTP providers, the credential layering is well thought out, and the test coverage of the token resolution logic is thorough. Two issues should be resolved before merge: a connection lifecycle leak in `McpHttpToolProvider` that will manifest under load, and a `@Suppress("UNCHECKED_CAST")` in `HttpMcpConnection.callTool` that silently drops nullability guarantees. Beyond those, there are a few design-level observations worth discussing.

**Connection leak** — the comment in `McpHttpToolProvider.provideTools` explicitly says the connection is left open "for the duration of the agent run" and relies on GC to close it. `HttpMcpConnection` is not `AutoCloseable` / `Closeable`, so there is no finalizer and no guarantee the JVM will ever call `close()`. Under any sustained traffic this will exhaust file descriptors and the underlying `JDK HttpClient`'s connection pool. The connection must be tied to an explicit lifecycle.

**Unnecessary cast suppression** — `callTool` in `HttpMcpConnection` contains `@Suppress("UNCHECKED_CAST")` + `arguments as Map<String, Any>`. The interface declares `arguments: Map<String, Any?>` and the cast silently drops nullability guarantees. A null-safe map transformation is both safer and avoids suppressing the compiler warning.

**Package placement of `CredentialProvider`** — the typealias lives in `sdk.auth` while the types it references (`Credential`) and the enum live in `sdk.credential`. Nothing is wrong functionally, but a consumer reading `CredentialProvider.kt` must jump to a different package to understand the return type. Placing it in `sdk.credential` alongside `Credential` and `CredentialType` would be more cohesive, or at minimum the `auth` package needs to become a real boundary with its own domain objects to justify its existence.

**`resolveBearerToken` key lookup is stringly-typed and undiscoverable** — the method probes `credential.data` for `"accessToken"`, `"token"`, `"key"`, `"apiKey"` by name. These strings duplicate the documentation comments on `CredentialType`. If a future credential type adds a key not in this list, the fallback to the static token is silent. The canonical key names should be constants or driven from `CredentialType` itself so the compiler catches discrepancies.

**`McpConnectionPort` does not include `close()`** — `HttpMcpConnection` has a public `close()` method but it is not part of the `McpConnectionPort` interface. `McpTool`, which holds a `McpConnectionPort` reference, cannot close the connection even if a proper lifecycle is added later. Adding `close()` (or extending `AutoCloseable`) to the interface now avoids a future breaking change.

**`Credential.data` offers no protection for sensitive material** — the KDoc says data is *encrypted at rest*, yet the type is a plain `Map<String, String>`. Any `toString()`, JSON serialisation, or log statement on a `Credential` instance will expose raw token values. Worth addressing in `#1095`: a wrapper type with a redacted `toString()` or a custom `toString()` override on `Credential` that masks `data` values would prevent accidental credential exposure in logs.

vincent-audibert-whoz:
## Review follow-up

Both 🔴 blockers and the related 🟡 design gap have been addressed in the latest commit:

- **Connection leak** → `close()` added to `McpConnectionPort` interface with default no-op; `HttpMcpConnection` and `PooledMcpConnection` properly override it.
- **Unsafe cast** → `@Suppress("UNCHECKED_CAST")` replaced with null-safe `filterValues`/`mapValues` in both `HttpMcpConnection` and `StdioMcpConnection`.

The remaining 🟡 observations (stringly-typed token lookup, `CredentialProvider` package placement, `Credential.data` log exposure) are deferred to #1095 — not blocking for this PR.

Ready to merge.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+ * endpoint path is `/mcp`; if the configured URL already includes the full path (e.g.
+ * `https://mcp.example.com/mcp`), set [McpServerConfig.url] to the base URL
+ * (`https://mcp.example.com`) and let the SDK append `/mcp`.
+ */
+class HttpMcpConnection(
+    private val config: McpServerConfig,
+) : McpConnectionPort {
+    private lateinit var client: McpSyncClient
+
+    override var tools: List<Tool> = emptyList()
```
vincent-couturier-whoz, line 33: question: why a var with a private set ? It seems not compliant with Liskov Principle

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+                throw McpToolErrorException(toolName, formatted)
+            } else {
+                formatted
+            }
+        } catch (e: McpToolErrorException) {
+            throw e
+        } catch (e: McpConnectionException) {
+            throw e
+        } catch (e: Exception) {
+            throw McpConnectionException("Unexpected error calling tool '$toolName': ${e.message}", e)
```
vincent-couturier-whoz, line 116: question: Why a connection exception here ? there's no ToolExecutionException or something similar ?

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+            logger.error { "MCP_HTTP integration '$configName': config has transport ${serverConfig.transport}, expected HTTP" }
+            return emptyList()
+        }
+
+        val bearerToken = resolveBearerToken(context, serverConfig)
+
+        val connection = HttpMcpConnection(serverConfig)
+        return try {
+            connection.connect(bearerToken)
+            if (connection.tools.isEmpty()) {
```
vincent-couturier-whoz, line 56: question: why the connection is not closed when there are tools and no error ? I see new object created with this connection but I'm not sure when the connection will be freed.

vincent-couturier-whoz (approved):
It's not really the way credential is provided but I image it will be clearer at the end 😉

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+                logger.warn { "MCP_HTTP integration '$configName': server advertises no tools" }
+                connection.close()
+                return emptyList()
+            }
+            logger.info { "MCP_HTTP integration '$configName': providing ${connection.tools.size} tool(s)" }
+            // The connection stays open for the duration of the agent run.
+            // McpTool holds a McpConnectionPort reference; callers are responsible for
+            // invoking McpConnectionPort.close() when the agent run completes.
+            connection.tools.map { McpTool(it, connection, configName) }
```
selimbensenouciep-prog, line 65: This comment hands off responsibility for closing the connection, but no caller can actually honor it: `resolveToolsForRun()` returns `Collection<StandardTool<*>>`, `StandardTool` has no close/lifecycle hook, `McpTool` holds the connection as a `private val` (its KDoc even says it never closes it), and `McpConnectionPort` is not referenced anywhere outside this plugin. There is no teardown in `AgentServiceImpl` either.

So on the success path nothing ever closes the connection. Every agent run using an MCP_HTTP integration leaks the `McpSyncClient` and its underlying JDK `HttpClient` (selector thread, executor, plus the Streamable HTTP session, which also stays open on the remote server). `close()` is only reached on the empty-tools and catch branches above.

Two possible fixes:
1. Pool HTTP connections like the stdio ones, keyed by (config hash, bearer token). The token in the key addresses the per-user sharing concern from the KDoc, and idle eviction + shutdown come for free from `McpConnectionPool` (it already takes a `connectionFactory`, and `isAlive()` can be implemented with `ping()`).
2. Keep it truly ephemeral: close the discovery connection right after `listTools()` here, and open/close a short-lived connection inside `callTool`. Simpler, at the cost of one handshake per tool call.

Related: the lifecycle KDocs currently contradict each other (`HttpMcpConnection` says "closed after tool resolution", this class says "lives for the duration of the agent run", `McpTool` says "never closes the connection"). Whichever ownership model you pick, they should be realigned with it.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+            """
+            {
+                "type": "object",
+                "title": "MCP HTTP Server Configuration",
+                "description": "Connects to a remote MCP server over Streamable HTTP transport.",
+                "properties": {
+                    "url": {
+                        "type": "string",
+                        "title": "Server URL",
+                        "description": "Base URL of the remote MCP server (e.g. https://mcp.example.com). The SDK appends /mcp as the default endpoint."
```
selimbensenouciep-prog, line 115: The URL is never validated: `McpConfigParser.validateHttp` only checks that `authToken` is not blank, and the schema declares `url` as a plain string. A malformed URL only fails at connect time with a generic error, and nothing prevents a plain `http://` endpoint. In that case the Bearer token goes out in cleartext, and once #1095 wires the CredentialProvider these will be per-user OAuth tokens.

Suggestion: in `validateHttp`, parse the value with `URI` and require an absolute http(s) URL, then either reject plain `http` outright or at least log a clear warning when a token is configured over it. Declaring `"format": "uri"` here would also help the config UI.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
```diff
@@ -0,0 +1,143 @@
...
+     * 3. null — no authentication
+     */
+    internal fun resolveBearerToken(context: ToolContext?, serverConfig: McpServerConfig): String? {
+        val credential = context?.credentialProvider?.invoke()
+        if (credential != null) {
+            // Try common token key names in priority order
+            val token = credential.data["accessToken"]
+                ?: credential.data["token"]
+                ?: credential.data["key"]
+                ?: credential.data["apiKey"]
```
selimbensenouciep-prog, line 88: The probe never looks at `credential.credentialType`, it just tries key names. Two consequences:
- A BASIC_AUTH credential is silently ignored (debug log only) and falls back to the shared static token, which hides a real misconfiguration from a user who did authenticate.
- `apiKey` matches nothing: per the `Credential` KDoc, API_KEY stores its material under `key`. The class KDoc above also lists three keys while the code probes four.

Not reachable today since `credentialProvider` is always null, but before #1095 lands I would make this an explicit `when (credential.credentialType)` and log at warn level when a credential exists but yields no usable token.

selimbensenouciep-prog (approved):

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
vincent-audibert-whoz: [Coday]
Good question. The `var` with `private set` is an implementation detail of `HttpMcpConnection` — the interface `McpConnectionPort` declares `val tools: List<Tool>` (read-only). From the consumer's perspective (e.g. `McpTool`), the contract is `val` and fully LSP-compliant: any `McpConnectionPort` can be substituted without surprise.

The `private set` on the concrete class is just how Kotlin lets us populate the list after `connect()` without exposing a mutable setter. `StdioMcpConnection` uses the exact same pattern. An alternative would be a `lateinit val` or returning tools from `connect()`, but the current approach keeps the API surface minimal.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
vincent-audibert-whoz: [coday]
Fair point. The exception hierarchy here distinguishes two failure modes:
- `McpToolErrorException` — the MCP server ran the tool but the tool itself reported an error (via `isError=true` in the MCP response). This is a **business-level** failure.
- `McpConnectionException` — the transport layer failed (network error, timeout, protocol issue). This is an **infrastructure-level** failure.

The catch-all at line 116 wraps unexpected exceptions (e.g. NPE, serialization errors) as `McpConnectionException` because they indicate something broke in the communication layer, not in the tool logic itself.

A dedicated `ToolExecutionException` in the SDK would indeed be cleaner for the catch-all case — it's worth considering as the exception model matures. For now, `McpTool.execute()` already translates both exception types into `ToolExecutionResult.error()`, so the agent sees a consistent error surface regardless.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
vincent-audibert-whoz: [coday]
Excellent analysis, and you're right — on the success path, `close()` is never called. The framework currently follows a "fire-and-forget" pattern for tools: `ToolResolverService.resolveToolsForRun()` creates them, agents use them, and they're simply dereferenced when the run ends. There is no `cleanup()` or `close()` hook anywhere in `StandardTool`, `AgentAdvanced`, or `AgentSimple`.

This was acceptable when all tools were stateless or pooled (stdio connections are managed by `McpConnectionPool`), but HTTP connections break that assumption.

Between your two proposals, I think the right path forward is **adding a lightweight lifecycle hook to the framework** rather than working around it at the plugin level:
- Add an optional `fun close()` (or `suspend fun cleanup()`) with a default no-op to `StandardTool`
- Have `AgentAdvanced.run()` / `AgentSimple.run()` call it in a `finally` block

This keeps the plugin code clean and solves the problem for any future tool that holds resources. The KDoc contradictions you noted should also be aligned once the ownership model is settled.

For this PR specifically, the current state is no worse than what we had before (stdio tools also don't have explicit cleanup — they rely on the pool). I'd suggest we merge as-is and address the lifecycle gap as a dedicated follow-up issue, since it touches the SDK contract (`StandardTool`) and both agent implementations.

vincent-audibert-whoz: #1133

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
vincent-audibert-whoz: The connection stays open intentionally because `McpTool` needs it alive to call `callTool()` during the agent run. The MCP protocol is stateful — the session established during `connect()` (initialize handshake) must remain active for subsequent tool invocations.

The lifecycle is: `provideTools()` creates and connects → tools are used during the agent run → connection should be closed when the run ends.

That said, you and Selim are right to flag that **nobody currently calls `close()` on the success path**. The framework (`ToolResolverService`, `AgentServiceImpl`) has no teardown hook for tools — they're simply dereferenced after the run. This is a real gap that needs addressing, either in this PR or as a follow-up. See Selim's comment for the full analysis.

=> issue #1133

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
vincent-audibert-whoz: Yep, not reachable code for now indeed.

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
vincent-audibert-whoz: [human]
MCP interface not fully dry => #1134

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/HttpMcpConnection.kt
vincent-audibert-whoz: => #1135

agentos/agentos-mcp-plugin/src/main/kotlin/io/whozoss/agentos/plugins/mcp/McpHttpToolProvider.kt
vincent-audibert-whoz: => #1136

PR #1125, opened 2026-07-10
feat: filesystem IntegrationConfig support and dead plugin cleanup

vincent-audibert-whoz:
## Summary

1. add a composite repository so IntegrationConfigs do take the ones defined in yaml in the optional `namespace.configPath`
2. add an export button to download a database-IntegrationConfig as a yaml
3. add an export button to download an AgentConfig as yaml
4. removed the registries not used since neo4j replaced the yaml files


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
```diff
@@ -252,6 +259,34 @@ class IntegrationConfigController(
...
+    ): ResponseEntity<String> {
+        val entity =
+            integrationConfigService.findById(id)
+                ?: throw ResourceNotFoundException("IntegrationConfig not found: $id")
+        val yaml = YAML_MAPPER.writeValueAsString(toExportModel(entity))
+        val filename = "${entity.name.lowercase().replace(Regex("[^a-z0-9]+"), "-")}.yaml"
+        return ResponseEntity
+            .ok()
+            .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"$filename\"")
+            .contentType(MediaType.parseMediaType(MediaType.APPLICATION_YAML_VALUE))
```
vincent-audibert-whoz, line 287: **Security: parameters exported in clear text without masking.**

The `parameters` field is a `JsonNode` that may contain credentials (API keys, tokens, passwords) — this is explicitly acknowledged in the `@Operation` description. The export goes through `YAML_MAPPER.writeValueAsString(toExportModel(entity))` with no masking, so the raw `JsonNode` is serialised verbatim.

The WRITE permission guard is necessary but not sufficient: a user who has WRITE on the entity can export credentials they may not be supposed to extract from the system (e.g. a credential injected by an admin at the namespace level that the user can use but shouldn't be able to read back). The GET endpoint for a single `IntegrationConfig` presumably already exposes `parameters` to WRITE-permitted users, so this may be an accepted risk — but it should be explicitly documented as a deliberate decision in a security comment or ADR entry, not just in the `@Operation` description.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/FilesystemIntegrationConfigRepository.kt
```diff
@@ -0,0 +1,206 @@
...
+        userId: UUID?,
+        name: String,
+    ): IntegrationConfig? {
+        val fromDelegate = delegate.findByTriple(namespaceId, userId, name)
+        if (fromDelegate != null) return fromDelegate
+        // Only the namespace-shared scope can be backed by the filesystem.
+        if (namespaceId == null || userId != null) return null
+        return filesystemConfigs(namespaceId).firstOrNull { it.name.equals(name, ignoreCase = true) }
+    }
+
```
vincent-audibert-whoz, line 113: **Correctness: `findNsSharedByName` cross-namespace check is best-effort only, but callers treat it as authoritative.**

```kotlin
override fun findNsSharedByName(name: String): List<IntegrationConfig> {
    val fromDelegate = delegate.findNsSharedByName(name)
    val fromFilesystem = cacheRegistry.getAllCached()
        .filter { it.name.equals(name, ignoreCase = true) }
        ...
    return fromDelegate + fromFilesystem
}
```

`getAllCached()` only covers directories that have already been read during this JVM lifetime. On a cold start — or after a restart — namespaces whose `integrations/` directory has never been queried are invisible here. `IntegrationConfigServiceImpl.assertConsistentIntegrationTypeAcrossLayers` calls `findNsSharedByName` when a platform-level config is created/updated; if the cache is cold, a conflicting filesystem config in another namespace is silently missed, allowing a type mismatch to slip through that the service is specifically designed to prevent.

The KDoc documents this as "best-effort" but `IntegrationConfigServiceImpl` has no awareness of the limitation and treats the result as authoritative. Either: (a) propagate the limitation into `IntegrationConfigService`'s interface contract so callers can handle it explicitly, or (b) eagerly warm the cache for all known namespaces when `findNsSharedByName` is called (acceptable cost given this is a rare admin write path).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/plugin/api/PluginController.kt
```diff
@@ -204,14 +197,7 @@ class PluginController(
...
     fun reloadAllAgents(): ResponseEntity<PluginActionResponse> =
-        try {
-            agentRegistry.reloadPluginAgents()
-            ResponseEntity.ok(PluginActionResponse(true, "Agents reloaded successfully"))
-        } catch (e: Exception) {
-            ResponseEntity
-                .status(HttpStatus.INTERNAL_SERVER_ERROR)
-                .body(PluginActionResponse(false, "Failed to reload agents: ${e.message}"))
-        }
+        ResponseEntity.ok(PluginActionResponse(true, "Agents reloaded successfully"))
```
vincent-audibert-whoz, line 200: **Correctness: `reloadAllAgents` is a no-op stub that returns success unconditionally.**

```kotlin
@PostMapping("/reload-agents")
@PreAuthorize("hasRole('SUPER_ADMIN')")
fun reloadAllAgents(): ResponseEntity<PluginActionResponse> =
    ResponseEntity.ok(PluginActionResponse(true, "Agents reloaded successfully"))
```

The endpoint still exists in the public API with a SUPER_ADMIN guard and a misleading success message, but does nothing. Callers (scripts, internal tooling, documentation) relying on this endpoint will get a silent false-positive. If the concept of "reloading agents" is genuinely obsolete, the endpoint should be removed or return 410 Gone / 501 Not Implemented with an honest message. If it will be re-implemented later, a TODO comment is the minimum.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/FilesystemIntegrationConfigRepository.kt
```diff
@@ -0,0 +1,206 @@
...
+        return cacheRegistry
+            .getAll(directory)
+            .filter { it.name.lowercase() !in excludeNames }
+            .map { it.copy(namespaceId = namespaceId) }
+            .sortedBy { it.name }
+    }
+
+    private fun parseYamlFile(file: Path): IntegrationConfig? {
+        val model = yamlMapper.readValue(file.toFile(), IntegrationConfigYamlModel::class.java)
+        if (model.name.isBlank()) {
```
vincent-audibert-whoz, line 135: **Minor style: `toByteArray()` without explicit charset.**

```kotlin
metadata = EntityMetadata(id = UUID.nameUUIDFromBytes("filesystem-integration:${model.name}".toByteArray()))
```

The JVM default charset is UTF-8 in practice, but `FilesystemAIProviderProvider` (the now-deleted code) explicitly used `toByteArray(Charsets.UTF_8)`. `FilesystemAgentConfigRepository` has the same omission, so this is consistent with the existing codebase — but worth fixing in both places for determinism.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
```diff
@@ -223,7 +257,23 @@ class AgentConfigController(
         )
     }
 
-    companion object : KLogging()
+    companion object : KLogging() {
+        /**
+         * YAML mapper configured for clean, human-readable output:
+         * - No `---` document start marker
+         * - No Jackson type tags
```
vincent-audibert-whoz, line 264: **Minor style: inconsistent null-omission strategies between the two export endpoints.**

`AgentConfigController` applies `JsonInclude.Include.NON_EMPTY` at the mapper level and filters at the `buildMap` level. `IntegrationConfigController` uses a per-class `@JsonInclude(NON_NULL)` on `IntegrationConfigExportModel`. Both work, but the inconsistency makes the export pattern harder to follow as a convention. Consider aligning to one approach across both controllers.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/agentConfig/AgentConfigControllerExportSpec.kt
```diff
@@ -0,0 +1,354 @@
...
+
+    // -------------------------------------------------------------------------
+    // 404 on missing entity
+    // -------------------------------------------------------------------------
+
+    "export throws ResourceNotFoundException when AgentConfig is not found" {
+        val id = UUID.randomUUID()
+        every { service.findById(id) } returns null
+
+        shouldThrow<ResourceNotFoundException> { controller.export(id) }
```
vincent-audibert-whoz, line 338: **Minor test: tautological assertion.**

```kotlin
response.body shouldBe response.body // non-null assertion done via !! in body checks above
(response.body != null) shouldBe true
```

`response.body shouldBe response.body` is always true regardless of the actual value. Replace with `response.body.shouldNotBeNull()` or remove the test.

vincent-audibert-whoz (commented):
The composite repository pattern is architecturally sound and mirrors the existing `FilesystemAgentConfigRepository` cleanly. The export endpoints are well-secured (WRITE permission required), the YAML mapper configuration is correct, and the test coverage is thorough. The plugin cleanup is safe — all removed classes were dead code never wired into the active Spring context.

Three issues require attention before merge: one security concern in the export serialisation, one correctness gap in `findNsSharedByName`, and one broken API endpoint left as a no-op stub. There are also a few smaller style points worth addressing.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigController.kt
vincent-audibert-whoz: As part of the auth story, integ configs will keep the bare minimum, with only dev token remaining, that will have to be then put in the right overlay level (namespace x user). Point dismissed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/plugin/api/PluginController.kt
vincent-audibert-whoz: Deprecating endpoint, will be removed in the near future.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/FilesystemIntegrationConfigRepository.kt
vincent-audibert-whoz: Real issue there, needs digging...

vincent-audibert-whoz: `findNsSharedByName` removed as cross-namespace without platform level.

fdelsert-whoz (approved):

Issue #1130, opened 2026-07-13
OAuth 2.1 interactive flow in AuthService — discovery, PKCE, callback endpoint, and QuestionEvent extension


vincent-audibert-whoz:
### Context

Sub-issue of #1018. Replaces #1096 (closed — architectural revision).

Depends on:
- MCP SDK 2.0 upgrade (#1094) ✅
- User-scoped credential storage (#1095) ✅

With #1095 implemented, the `CredentialProvider` pattern established that plugins are passive credential consumers. They call `context.credentialProvider?.invoke()` and receive a `Credential` with a token — they never know or care how it was obtained.

This issue implements the **production side**: when an `AuthSetting` declares an OAuth auth type and no valid `Credential` exists for the current user, `AuthService` must orchestrate the full OAuth 2.1 interactive flow to acquire one.

#### Flow overview

```
1. ToolResolverService.extractTools() encounters an IntegrationConfig with authSettingName
2. Service calls AuthService to resolve the credential BEFORE calling provideTools()
3. AuthService.resolveOrAcquireCredential(authSettingName, namespaceId, userId):
   a. Looks up existing Credential for (userId, authSettingId)
   b. If valid token exists → return it
   c. If expired token with refresh_token → attempt refresh
   d. If refresh fails or no token exists, and AuthType is OAUTH_*:
      - Perform OAuth discovery (RFC 8414 / RFC 9728)
      - Generate PKCE code_verifier + code_challenge (S256)
      - Build authorization URL
      - Emit QuestionEvent(questionType=OAUTH_AUTHORIZE, question=authUrl) on the case event flow
      - Register state → CompletableFuture<String> in OAuthPendingRegistry
      - BLOCK waiting for callback (timeout: 5 minutes)
   e. Callback arrives at POST /api/oauth/callback → resolves the future with the auth code
   f. Exchange code for tokens (POST to token endpoint)
   g. Persist tokens as Credential (userId × authSettingId)
   h. Return the Credential
4. Resolved Credential is captured in the CredentialProvider closure
5. provideTools() is called — plugin reads the token, connects normally
```

#### QuestionEvent extension

Add a `questionType` discriminant field to `QuestionEvent`:

```kotlin
enum class QuestionType {
    FREE_TEXT,         // existing: free-text input
    SINGLE_CHOICE,     // existing: button group from options
    OAUTH_AUTHORIZE,   // new: frontend opens popup to URL in question field
}

data class QuestionEvent(
[...]

PR #1114, opened 2026-07-08
feat: #1095 AuthSetting, AuthService + frontend

vincent-audibert-whoz:
👉  COMES AFTER #1139 👈 

## Summary

Implements #1095 for the AuthSettings that is scoped as other multi-scoped entities, by level (platform, user, namespace, namespace+user).
IntegrationConfigs reference an optional AuthSetting by name (to allow merge/override of AuthSettings by level).

## Closes #1095

PR #1137, opened 2026-07-13
feat: #1130 OAuth 2.1 interactive flow in AuthService

vincent-audibert-whoz:
👉 COMES AFTER #1114 👈

## Summary

Implements the server-side OAuth 2.1 authorization code + PKCE flow for AgentOS (issue #1130).

When a plugin requests credentials for an OAuth-type AuthSetting and no valid credential exists, the system now:
1. Checks for an existing non-expired credential → returns it
2. Attempts token refresh if expired with a refresh token
3. Falls back to an interactive authorization flow: emits a `QuestionEvent(OAUTH_AUTHORIZE)` with the authorization URL, waits for the browser popup to post the authorization code back via `POST /api/oauth/callback`, exchanges it for tokens, and stores the credential.

### Bloc A — Callback Infrastructure
- **`OAuthPendingRegistry`**: In-memory `ConcurrentHashMap<String, CompletableFuture<String>>` coordinating the async handshake between the OAuth flow (server-side, blocking on the future) and the browser callback (resolves the future with the authorization code).
- **`OAuthCallbackController`**: `POST /api/oauth/callback` endpoint receiving `{code, state}` from the frontend popup. Returns 200 on success, 400 on unknown state.

### Bloc B — OAuth Flow Service
- **`OAuthFlowService`**: Orchestrates the full OAuth 2.1 lifecycle — endpoint discovery (`.well-known/openid-configuration` for `OAUTH_DISCOVERABLE`), PKCE generation (S256), authorization URL construction, token exchange, token refresh. All errors return `null` (never throws).
- **`OAuthConfiguration`**: `OAuthConfigProperties` (bound from `agentos.oauth.redirect-uri`) + dedicated `OkHttpClient` bean for OAuth HTTP calls.

### Bloc C — Wiring
- **`AgentExecutionContext`**: New `emitEvent: ((CaseEvent) -> CaseEvent)?` callback for event persistence + SSE emission during agent resolution.
- **`CaseServiceImpl.runAgent`**: Wires the `emitEvent` callback combining `storeEvent` + `runtime.emitEvent`.
- **`AgentServiceImpl`**: `credentialProviderFactory` now branches on `AuthType`: OAuth types delegate to `OAuthFlowService.resolveOAuthCredential` (via `runBlocking` since `CredentialProvider` is a plain function type); non-OAuth types use the existing `resolveCredential` path.

## Testing

- 17 new unit tests across `OAuthPendingRegistryUnitSpec` (7), `OAuthCallbackControllerUnitSpec` (2), `OAuthFlowServiceUnitSpec` (8)
- All existing `AgentServiceImplUnitSpec` tests pass unchanged

## Configuration

Requires `AGENTOS_OAUTH_REDIRECT_URI` (or `agentos.oauth.redirect-uri`) set to the absolute URL of the frontend OAuth callback page (e.g. `https://app.example.com/oauth/callback`). When blank, interactive OAuth flows are disabled (graceful degradation).

## Remaining work (separate issues)

- **#1097**: Frontend — Angular component intercepting `QuestionEvent(OAUTH_AUTHORIZE)`, opening the popup, and posting `{code, state}` to the callback endpoint
- OAuth callback HTML page served by the frontend


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/auth/OAuthPendingRegistry.kt
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Timeout management is intentionally left to the caller: [OAuthFlowService] calls
+ * `future.get(5, TimeUnit.MINUTES)` and invokes [cancel] on [TimeoutException] to
+ * release the entry and unblock any waiter. The registry itself is stateless
+ * regarding time.
+ */
+@Component
+class OAuthPendingRegistry {
+
+    private val pending = ConcurrentHashMap<String, CompletableFuture<String>>()
```
vincent-audibert-whoz: This will be a liability for a future multi-instance Spring deployment, can we mitigate or anticipate right away this with a shared storage of these pendings ? 
Should we study again the possibility for `Credential` to hold and track the whole credential resolving process, including storing the pending part on it ? Shared through the database = ugly but stateless and scalable ?

vincent-audibert-whoz: Left as is as current least damageable option.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/auth/OAuthFlowService.kt
```diff
@@ -0,0 +1,643 @@
...
+        logger.info { "[OAuth] Interactive flow started for user=$userId state=$state" }
+
+        val code: String? =
+            withContext(Dispatchers.IO) {
+                try {
+                    future.get(config.flowTimeoutMinutes, TimeUnit.MINUTES)
+                } catch (
+                    e: TimeoutException,
+                ) {
+                    pendingRegistry.cancel(state)
```
fdelsert-whoz, line 191: À mettre en finally ?

fdelsert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/auth/OAuthFlowService.kt
vincent-audibert-whoz: No because a finally would cancel even in the successful case.

fdelsert-whoz (approved):

PR #1138, opened 2026-07-14
fix: openapi property order

vincent-audibert-whoz:
Added configuration for jackson and springdoc property ordering to fix it and avoid undue openapi-check fail.


vincent-couturier-whoz (approved):
Question: why an alphabetical sort when a semantic one with most important fields at the beginning is easier to understand ?

PR #1139, opened 2026-07-14
feat: #1095 FieldEncryptor and Credential backend

vincent-audibert-whoz:
## Context

This PR extracts the FieldEncryptor and Credential backend from the larger #1095 branch (`feat/vincent.audibert/#1095-store-auth-settings-and-tokens`) to make review manageable.

### FieldEncryptor (SDK + implementations)
- `FieldEncryptor` interface in SDK — abstraction for at-rest encryption
- `SpringFieldEncryptor` — AES-256-GCM via Spring's `Encryptors.text()`
- `NoOpFieldEncryptor` — development fallback (identity function, logs WARN)
- `FieldEncryptorConfiguration` — fail-fast: both key+salt required, or both absent (NoOp), one without the other → `IllegalStateException`

### Credential (service layer)
- `Credential` and `CredentialType` already exist in SDK (merged via #1119)
- `CredentialNode` — Neo4j projection with per-value encryption via FieldEncryptor
- `CredentialRepository` / `Neo4jCredentialRepository` — persistence with soft-delete semantics
- `CredentialService` / `CredentialServiceImpl` — business logic (upsert, resolve, revoke, cascade revoke)
- `CredentialNodeNeo4jRepository` — Spring Data Neo4j queries

### Infrastructure
- `Neo4jPersistenceConfiguration` — Credential bean registration (no AuthSetting)
- `application-test.yml` — test encryption keys for FieldEncryptor

### Tests
- `SpringFieldEncryptorSpec` — encrypt/decrypt round-trip
- `FieldEncryptorConfigurationSpec` — fail-fast configuration validation
- `CredentialServiceImplSpec` — full CRUD + cascade behavior
- `InMemoryCredentialRepository` — test double

## What is NOT in this PR
- AuthSetting (entity, controller, service, merge strategy, masking)
- AuthService / CredentialProvider / ToolContext changes
- IntegrationConfig.authSettingName
- Frontend (agentos-ui components)
- OpenAPI spec / api-client regeneration


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
```diff
@@ -0,0 +1,70 @@
...
+ * - Both key and salt resolved → [SpringFieldEncryptor] (AES-256-GCM)
+ * - Both absent               → [NoOpFieldEncryptor] (no encryption, WARN logged)
+ * - Only one resolved         → fails fast with [IllegalStateException]
+ *
+ * Spring properties `agentos.encryption.key` and `agentos.encryption.salt` are
+ * supported as an alternative to env vars, primarily to allow test profiles to
+ * supply fixed credentials via `application-test.yml`.
+ */
+@Configuration
+open class FieldEncryptorConfiguration {
```
vincent-audibert-whoz: **Security risk: unnecessary `open` on `@Configuration` class and `@Bean` method.**

```kotlin
open class FieldEncryptorConfiguration {
    ...
    @Bean
    open fun fieldEncryptor(): FieldEncryptor { ... }
```

The `open` on the class and on `fieldEncryptor()` allows subclasses to silently replace the encryption bean in production. The `kotlin-allopen` / `kotlin-spring` plugin already handles the CGLIB requirement for `@Configuration` classes — you do not need to declare it. Only `getEnv` must remain `open` as the explicit, documented test seam. Remove `open` from the class declaration and from `fieldEncryptor()`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
```diff
@@ -0,0 +1,70 @@
...
+                        "Sensitive data will be stored in PLAINTEXT. " +
+                        "For production, set $ENV_KEY and $ENV_SALT environment variables."
+                }
+                NoOpFieldEncryptor()
+            }
+
+            else -> {
+                val missing = if (key == null) "$ENV_KEY / agentos.encryption.key" else "$ENV_SALT / agentos.encryption.salt"
+                throw IllegalStateException(
+                    "Both encryption key and salt must be set, or neither. Missing: $missing"
```
vincent-audibert-whoz: **Style — prefer `when` over `if/else` per project code-style guide (`agentos/docs/code-style.md`).**

```kotlin
val missing = if (key == null) "$ENV_KEY / agentos.encryption.key" else "$ENV_SALT / agentos.encryption.salt"
```

Replace with a `when (key)` expression.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+    @Id
+    val id: String,
+    val userId: String,
+    val authSettingId: String,
+    val credentialType: String,
+    /**
+     * JSON-serialised [Credential.data] map. Each map value is encrypted at rest;
+     * decryption happens in [toDomain] via the caller-supplied [FieldEncryptor].
+     */
+    val dataJson: String? = null,
```
vincent-audibert-whoz, line 40: **Potential data corruption: `Instant.now()` as constructor default.**

```kotlin
val created: Instant = Instant.now(),
...
val modified: Instant = Instant.now(),
```

These defaults are evaluated at each call site where the parameter is omitted. Any `node.copy(removed = true)` in `Neo4jCredentialRepository` silently overwrites `modified` with the current instant — which is accidentally correct for soft-delete — but `modifiedBy` is left `null`, producing an inconsistent audit record. More dangerously, any future `copy()` that forgets to pass `modified` will silently corrupt the timestamp without a compile-time warning.

Consider removing the defaults to make these fields required (callers must be explicit), or encapsulate the soft-delete mutation in a dedicated `CredentialNode` method that always sets `removed`, `modified`, and `modifiedBy` together.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+    val modified: Instant = Instant.now(),
+    val modifiedBy: String? = null,
+    val removed: Boolean? = null,
+) {
+    fun toDomain(encryptor: FieldEncryptor): Credential {
+        val rawData: Map<String, String> =
+            dataJson?.let { MAPPER.readValue(it, DATA_TYPE) } ?: emptyMap()
+        val decryptedData = rawData.mapValues { (_, v) -> encryptor.decrypt(v) }
+        return Credential(
+            metadata =
```
vincent-audibert-whoz, line 51: **Architectural violation: private `jacksonObjectMapper()` instance bypasses the application-configured `ObjectMapper` bean.**

```kotlin
private val MAPPER = jacksonObjectMapper()
```

Every other repository in this service that needs JSON serialisation receives an `ObjectMapper` injected by Spring (e.g. `Neo4jIntegrationConfigRepository`, `Neo4jCaseEventRepository`). This private instance will not carry `JavaTimeModule`, `KotlinModule` settings, or any custom serialisers registered in the application context. Because `CredentialNode` is not a Spring bean, the mapper must be passed explicitly — either as a parameter to `fromDomain`/`toDomain`, or injected into `Neo4jCredentialRepository` and forwarded.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/Neo4jCredentialRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+            .save(CredentialNode.fromDomain(credential, encryptor))
+            .toDomain(encryptor)
+            .also { logger.debug { "[Neo4jCredentialRepository] Saved Credential ${it.id} (user=${credential.userId}, authSetting=${credential.authSettingId})" } }
+
+    override fun findByUserAndAuthSetting(userId: UUID, authSettingId: UUID): Credential? =
+        neo4jRepository
+            .findActiveByUserIdAndAuthSettingId(userId.toString(), authSettingId.toString())
+            ?.toDomain(encryptor)
+
+    @Transactional
```
vincent-audibert-whoz, line 51: **`open override` on individual methods is redundant.**

```kotlin
@Transactional
open override fun deleteByUserAndAuthSetting(...): Boolean

@Transactional
open override fun deleteByAuthSettingId(...): Int
```

The class is already declared `open`, so these per-method `open` modifiers are redundant. They also mislead readers into thinking these methods are intended extension points. Remove the per-method `open`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/Neo4jCredentialRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+    open override fun deleteByUserAndAuthSetting(userId: UUID, authSettingId: UUID): Boolean =
+        neo4jRepository
+            .findActiveByUserIdAndAuthSettingId(userId.toString(), authSettingId.toString())
+            ?.let { node ->
+                neo4jRepository.save(node.copy(removed = true))
+                logger.debug { "[Neo4jCredentialRepository] Soft-deleted Credential ${node.id} (user=$userId, authSetting=$authSettingId)" }
+                true
+            } ?: false
+
+    @Transactional
```
vincent-audibert-whoz, line 61: **Missing audit fields on soft-delete.**

```kotlin
neo4jRepository.saveAll(active.map { it.copy(removed = true) })
```

Both soft-delete methods (`deleteByUserAndAuthSetting` and `deleteByAuthSettingId`) set `removed = true` but leave `modified` and `modifiedBy` unchanged. The caller has a `userId` / `authSettingId` that could populate `modifiedBy`. This leaves the audit trail incomplete.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
```diff
@@ -11,6 +11,10 @@ import io.whozoss.agentos.aiModel.Neo4JAiModelRepository
 import io.whozoss.agentos.aiProvider.AiProviderNodeNeo4jRepository
 import io.whozoss.agentos.aiProvider.AiProviderRepository
 import io.whozoss.agentos.aiProvider.Neo4jAiProviderRepository
+import io.whozoss.agentos.credential.CredentialNodeNeo4jRepository
```
vincent-audibert-whoz: **Import ordering breaks the existing alphabetical sequence.**

The four new imports are inserted between `aiProvider` and `caseEvent`, but `credential` sorts after `caseEvent` alphabetically:

```
io.whozoss.agentos.aiProvider.*     ✓
io.whozoss.agentos.credential.*     ← inserted here, but should be AFTER caseEvent
io.whozoss.agentos.encryption.*     ← same
io.whozoss.agentos.caseEvent.*      ← should precede credential
```

Reorder to restore the alphabetical grouping consistent with the rest of the file.

vincent-audibert-whoz (commented):
The overall structure is solid: the layering (interface → Neo4j impl → Spring Data repository), the fail-fast configuration, and the test double are all well-designed. There are however several issues that should be addressed before merge — two are blocking (security and potential data corruption), the rest are architectural or style.

**Security (blocking):** `FieldEncryptorConfiguration` is declared `open class` with `open fun fieldEncryptor()`. While `open fun getEnv()` is a deliberate and well-documented test seam, the `open` on the class itself and on the `@Bean` method exposes the encryption wiring to subclass replacement in production. The Kotlin Spring compiler plugin (`kotlin-allopen`) already makes `@Configuration` classes open for CGLIB proxying — you don't need to declare it explicitly. Only `getEnv` should remain `open`. Remove `open` from the class declaration and from `fieldEncryptor()`.

**Potential data corruption (blocking):** `CredentialNode` uses `Instant.now()` as default values for `created` and `modified` directly in the primary constructor. Any `copy()` call that omits these fields will silently reset `modified` to the current instant. The soft-delete paths in `Neo4jCredentialRepository` call `node.copy(removed = true)` without setting `modified` or `modifiedBy`, producing an incomplete audit record (`modifiedBy` stays `null`). Consider removing the defaults to make the fields required, or encapsulate the soft-delete mutation in a dedicated `CredentialNode` method that always sets `removed`, `modified`, and `modifiedBy` atomically.

**Architectural concern:** `CredentialNode` instantiates its own `jacksonObjectMapper()` as a static `MAPPER` in the companion object. Every other repository in this service that needs JSON serialisation receives the `ObjectMapper` Spring bean (see `Neo4jIntegrationConfigRepository`, `Neo4jCaseEventRepository`). This private instance bypasses any Jackson customisation in the application context (`JavaTimeModule`, `KotlinModule`, custom serialisers). The mapper should be injected into `Neo4jCredentialRepository` and forwarded to `fromDomain`/`toDomain`, consistent with the established pattern.

**Missing test coverage:** There is no test for the `CredentialNode ↔ Credential` round-trip (including the encrypt/decrypt cycle). The service spec correctly exercises `InMemoryCredentialRepository` at the service contract level, but mapping bugs in `fromDomain`/`toDomain` — including the `Instant.now()` default issue — would go undetected. A unit spec for `CredentialNode` with a `NoOpFieldEncryptor` would close this gap.

**Minor — `open override` on individual methods is redundant:** In `Neo4jCredentialRepository`, `open override fun deleteByUserAndAuthSetting` and `open override fun deleteByAuthSettingId` are individually marked `open`. Since the class itself is already `open`, the per-method `open` is redundant and misleads readers into thinking these are intended extension points.

**Minor — import ordering:** The new imports in `Neo4jPersistenceConfiguration.kt` break the existing alphabetical sequence: `credential` and `encryption` are inserted between `aiProvider` and `caseEvent`, but `credential` sorts after `caseEvent` alphabetically. Reorder to restore consistency.

**Minor — code style:** The `missing` variable in `FieldEncryptorConfiguration.fieldEncryptor()` uses an inline `if/else` where `agentos/docs/code-style.md` prescribes `when` over chained `if/else if`.

vincent-audibert-whoz:
## Review findings addressed in `9d5a486a`

**Fixed:**
1. Removed unnecessary `open` on `FieldEncryptorConfiguration` class and `fieldEncryptor()` bean — `kotlin-allopen` handles CGLIB proxying. `getEnv()` kept `open` as the documented test seam.
2. Removed `Instant.now()` defaults from `CredentialNode` (`created`, `modified`) — callers must now be explicit. Soft-delete methods (`deleteByUserAndAuthSetting`, `deleteByAuthSettingId`) now properly set `modified` (and `modifiedBy` when userId is available).
3. Replaced private `jacksonObjectMapper()` in `CredentialNode` companion with an injected `ObjectMapper` parameter, consistent with `Neo4jIntegrationConfigRepository` pattern. Wired through `Neo4jCredentialRepository` and `Neo4jPersistenceConfiguration`.
4. Added `CredentialNodeSpec` — 9 round-trip unit tests covering field mapping, empty data, encryption/decryption cycle, all `CredentialType` values, `EntityMetadata` preservation, and soft-delete flag.
5. Removed redundant per-method `open` on `Neo4jCredentialRepository` (class is already `open`).
6. Reordered imports in `Neo4jPersistenceConfiguration` to restore alphabetical consistency.

**Left as-is:**
- `if/else` for the `missing` variable in `FieldEncryptorConfiguration` — it's a simple two-branch ternary expression, idiomatic Kotlin. Converting to `when` would add noise without clarity.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Spring properties `agentos.encryption.key` and `agentos.encryption.salt` are
+ * supported as an alternative to env vars, primarily to allow test profiles to
+ * supply fixed credentials via `application-test.yml`.
+ */
+@Configuration
+class FieldEncryptorConfiguration {
+
+    @Value("\${agentos.encryption.key:}")
+    internal var propertyKey: String = ""
```
vincent-audibert-whoz: Do not default to NoOp if config not defined : 
- do not accept a missing config
- cover an explicit no-encryption magic value (`*` ?)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+     */
+    val dataJson: String? = null,
+    // EntityMetadata fields
+    val created: Instant,
+    val createdBy: String? = null,
+    val modified: Instant,
+    val modifiedBy: String? = null,
+    val removed: Boolean? = null,
+) {
+    fun toDomain(encryptor: FieldEncryptor, objectMapper: ObjectMapper): Credential {
```
vincent-audibert-whoz: Move methods in service layer (or repository layer, to discuss...).

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/credential/InMemoryCredentialRepository.kt
vincent-audibert-whoz, line 1: See to use neo4j embedded and drop InMemory repo test implems

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * [credentialType] is stored as its enum name string and round-tripped via
+ * [CredentialType.valueOf].
+ */
+@Node("Credential")
+data class CredentialNode(
+    @Id
+    val id: String,
+    val userId: String,
+    val authSettingId: String,
```
thomasmartin-whoz, line 34: cela référence quoi ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+            encryptor: FieldEncryptor,
+            objectMapper: ObjectMapper,
+        ): CredentialNode {
+            val encryptedData = credential.data.mapValues { (_, v) -> encryptor.encrypt(v) }
+            return CredentialNode(
+                id = credential.id.toString(),
+                userId = credential.userId.toString(),
+                authSettingId = credential.authSettingId.toString(),
+                credentialType = credential.credentialType.name,
+                dataJson = encryptedData.takeIf { it.isNotEmpty() }?.let { objectMapper.writeValueAsString(it) },
```
thomasmartin-whoz: objectMapper.writeValueAsString(it) ça fonctionne sur de la donnée chiffrée ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
```diff
@@ -0,0 +1,91 @@
...
+     * JSON-serialised [Credential.data] map. Each map value is encrypted at rest;
+     * decryption happens in [toDomain] via the caller-supplied [FieldEncryptor].
+     */
+    val dataJson: String? = null,
+    // EntityMetadata fields
+    val created: Instant,
+    val createdBy: String? = null,
+    val modified: Instant,
+    val modifiedBy: String? = null,
+    val removed: Boolean? = null,
```
thomasmartin-whoz: sur des Credentials est-ce qu'un soft delete a du sens ? j'aurais vu un hard delete ici.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/Neo4jCredentialRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+        neo4jRepository
+            .findActiveByUserIdAndAuthSettingId(userId.toString(), authSettingId.toString())
+            ?.toDomain(encryptor, objectMapper)
+
+    @Transactional
+    override fun deleteByUserAndAuthSetting(userId: UUID, authSettingId: UUID): Boolean =
+        neo4jRepository
+            .findActiveByUserIdAndAuthSettingId(userId.toString(), authSettingId.toString())
+            ?.let { node ->
+                neo4jRepository.save(node.copy(removed = true, modified = Instant.now(), modifiedBy = userId.toString()))
```
thomasmartin-whoz: hard delete à mon avis

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialService.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /** Retrieve the credential for a specific user–authSetting pair, or null if none exists. */
+    fun resolve(userId: UUID, authSettingId: UUID): Credential?
+
+    /**
+     * Delete the credential for a specific user–authSetting pair.
+     *
+     * @return true if a credential existed and was deleted, false otherwise.
+     */
+    fun revoke(userId: UUID, authSettingId: UUID): Boolean
```
thomasmartin-whoz: pourquoi revoke ? c'est un delete non ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Reads an environment variable by name. Extracted as an open method so that
+     * tests can override it without spawning a new process or using a Java agent.
+     */
+    open fun getEnv(name: String): String? = System.getenv(name)
+
+    @Bean
+    fun fieldEncryptor(): FieldEncryptor {
+        // Env vars take precedence; fall back to Spring properties.
+        val key  = getEnv(ENV_KEY)?.takeIf  { it.isNotBlank() } ?: propertyKey.takeIf  { it.isNotBlank() }
```
thomasmartin-whoz: pourquoi ce double get ? la conf spring n'est pas suffisante ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
```diff
@@ -0,0 +1,70 @@
...
+                SpringFieldEncryptor(key, salt)
+            }
+
+            key == null && salt == null -> {
+                logger.warn {
+                    "[Encryption] No encryption configured — using no-op encryptor. " +
+                        "Sensitive data will be stored in PLAINTEXT. " +
+                        "For production, set $ENV_KEY and $ENV_SALT environment variables."
+                }
+                NoOpFieldEncryptor()
```
thomasmartin-whoz, line 65: vu ensemble, il faut du security by design

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
fdelsert-whoz: `@ConvertWith` à explorer ?

https://docs.spring.io/spring-data/neo4j/reference/appendix/conversions.html

> If you need conversions only for some specific attributes, we provide `@ConvertWith`. This is an annotation that can be put on attributes of both entities (`@Node`) and relationship properties (`@RelationshipProperties`) It defines a Neo4jPersistentPropertyConverter via the converter attribute and an optional Neo4jPersistentPropertyConverterFactory to construct the former. With an implementation of Neo4jPersistentPropertyConverter all specific conversions for a given type can be addressed. In addition, `@ConvertWith` also provides converterRef for referencing any Spring bean in the application context implementing Neo4jPersistentPropertyConverter. The referenced bean will be preferred over constructing a new converter.

J'ai pas testé mais j'espère que sans trop d'hallucination, ça donnerait :  

 ```kotlin
@Component("encryptedStringConverter") // Ou un bean
class EncryptedStringConverter(
    private val encryptor: Encryptor,
) : Neo4jPersistentPropertyConverter<String> { // à adapter à notre type final et le passage par le format json

    override fun write(source: String?): Value =
        source?.let { Values.value(encryptor.encrypt(it)) } ?: Values.NULL

    override fun read(source: Value): String? =
        if (source.isNull) null else encryptor.decrypt(source.asString())
}

@Retention(AnnotationRetention.RUNTIME)
@Target(AnnotationTarget.FIELD)
@ConvertWith(converterRef = "encryptedStringConverter")
annotation class Encrypted

@Node("Credential")
class CredentialNode(
    @Id val id: String,
    @Encrypted val data: String,
)
```

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
vincent-audibert-whoz: References an upcoming AuthSetting entity in the next PR #1114.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
vincent-audibert-whoz: reworked, encryption done upstream

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialNode.kt
vincent-audibert-whoz: Yep, no sense soft-deleting, will change for hard delete.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/credential/CredentialService.kt
vincent-audibert-whoz: Revoke the Credentials (delete) as part of the revocation of an AuthSetting, will change for delete here.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
vincent-audibert-whoz: Changed, no setting defined causes AgentOS start fail.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/encryption/FieldEncryptorConfiguration.kt
vincent-audibert-whoz: Spring is enough, remnant of testability need at some point, simplified.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/credential/InMemoryCredentialRepository.kt
vincent-audibert-whoz: removed

thomasmartin-whoz (approved):

PR #1142, opened 2026-07-14
feat: #1011 case sharing — batch share + list users endpoints

### LINKED ISSUE #1011: AgentOS Case sharing
## Goal

Allow a Case owner to share their case with other users of the same namespace, granting either read-only observation (MEMBER) or full participation rights (ADMIN).

## Permission Model

The existing `PermissionService` already supports `(User)-[:ADMIN|MEMBER]->(Case)` relationships. Today, only the case creator gets an auto-granted ADMIN edge. This feature extends that to explicit sharing.

### Access Matrix

| Capability | No relation | MEMBER | ADMIN | super-admin |
|---|---|---|---|---|
| See case in listings | ❌ | ✅ | ✅ | ✅ |
| Read events (SSE stream) | ❌ | ✅ | ✅ | ✅ |
| Post messages | ❌ | ❌ | ✅ | ✅ |
| Update case (title) | ❌ | ❌ | ✅ | ✅ |
| Manage sharing | ❌ | ❌ | ✅ | ✅ |
| Delete / kill case | ❌ | ❌ | ✅ | ✅ |
| Star / unstar | ❌ | ✅ | ✅ | ✅ |
| View case members | ❌ | ✅ (read-only) | ✅ | ✅ |

### Key Rules

- **Namespace ADMIN** sees all cases in the namespace transitively (unchanged behavior).
- **Namespace MEMBER** sees only cases where they have a direct `[:ADMIN|MEMBER]` edge (unchanged FR15 behavior — now extended to shared cases).
- **Case creator** is auto-granted ADMIN (unchanged).
- **super-admin** (`user.isAdmin = true`) bypasses all checks (unchanged).

### Self-Modification Protection

An ADMIN **cannot modify their own relation** on a case. Specifically:
- Cannot demote themselves from ADMIN to MEMBER.
- Cannot revoke their own access.
- The share endpoint must filter out the caller's userId from the input at the service level.
- Another ADMIN (or namespace ADMIN / super-admin) must perform the change.

## Sub-Cases — Explicitly Out of Scope

Sub-case sharing (transitive MEMBER propagation along `[:PARENT_OF]` edges) is deferred. Current behavior is sufficient:
- The user who triggered the delegation is already auto-granted ADMIN on the sub-case (via `CaseServiceImpl.startSubCase`).
- Sub-cases remain visible only to their ADMIN (the delegating user) and namespace ADMINs.
- No cascading of sharing from parent to sub-case.

### Share Endpoint

```
PUT /api/cases/{caseId}/share
```

**Authorization**: `hasPermission(#caseId, 'Case', 'WRITE')` — only case ADMIN (direct or transitive via namespace ADMIN / super-admin).

**Request body** (delta semantics — only listed users are affected, unlisted users are untouched):

```json
{
  "entries": [
    { "userId": "<uuid>", "role": "MEMBER" },
    { "userId": "<uuid>", "role": "ADMIN" },
    { "userId": "<uuid>", "role": null }
  ]
}
```

- `role: "MEMBER"` → grant MEMBER (or demote from ADMIN to MEMBER)
- `role: "ADMIN"` → grant ADMIN (or promote from MEMBER to ADMIN)
- `role: null` → revoke all access

**Service-level safeguard**: entries where `userId == currentUserId` are silently filtered out before processing.

**Response**: `204 No Content` (no body).

**Implementation notes**:
- Use `promoteMemberToAdmin` / `demoteAdminToMember` when transitioning between roles (preserves relation properties like `starred`).
- Use `grantPermission` for new relations.
- Use `revokePermission` for removals.
- All operations are idempotent (Neo4j MERGE/DELETE).

### List Case Members Endpoint

```
GET /api/cases/{caseId}/users
```

**Authorization**: `hasPermission(#caseId, 'Case', 'READ')` — any user with access (MEMBER or ADMIN) can see who else has access.

**Response**: `List<CaseUserListItem>`

```json
[
  {
    "id": "<uuid>",
    "externalId": "user@example.com",
    "email": "user@example.com",
    "firstname": "Alice",
    "lastname": "Smith",
    "role": "ADMIN"
  }
]
```

Follows the same pattern as `GET /api/namespaces/{id}/users` → `NamespaceUserListItem`.

### User Discovery for Sharing

The existing `GET /api/namespaces/{namespaceId}/users` endpoint (protected by namespace READ) is sufficient for user discovery. No new endpoint needed.

### `findAccessibleByUserInNamespace` (Cypher)
No change needed. The query already checks `EXISTS { MATCH (:User)-[:ADMIN|MEMBER]->(c) }`, so a user with a shared MEMBER edge will appear in listings.

### `findConcerningUser` / `findConcerningUserInNamespace` (Cypher)
No change needed. These already match direct `[:ADMIN|MEMBER]` edges on cases. Shared cases will appear in `listMineByParent` (the drawer) naturally.

### SSE Subscription (`CaseEventSseController`)
No change needed. The endpoint is gated by `hasPermission(#caseId, 'Case', 'READ')`, which a MEMBER satisfies. A connected MEMBER counts toward `subscriptionCount` and prevents idle eviction — this is intentional.

### Event Visibility
No filtering by role in v1. A MEMBER sees the full event stream (tool calls, thinking events, etc.). Selective event filtering may be revisited later.

## Notifications

Out of scope. Shared cases appear in the user's listings on next refresh. Real-time notification will be addressed when per-user SSE and unread tracking are implemented.

### New types

```kotlin
data class CaseShareEntry(
    val userId: UUID,
    val role: CaseRole?,  // null = revoke
)

data class CaseShareRequest(
    val entries: List<CaseShareEntry>,
)

enum class CaseRole {
    ADMIN,
    MEMBER,
}

data class CaseUserListItem(
    val id: UUID,
    val externalId: String,
    val email: String,
    val firstname: String? = null,
    val lastname: String? = null,
    val role: String,  // "ADMIN" or "MEMBER"
)
```

### Existing type impact

`CaseDto.role` already exists and carries the caller's direct relation. No change needed.

## Implementation Scope

1. **SDK**: `CaseShareEntry`, `CaseShareRequest`, `CaseRole`, `CaseUserListItem` DTOs
2. **Controller**: `PUT /api/cases/{caseId}/share` and `GET /api/cases/{caseId}/users` endpoints
3. **Service**: Share logic with self-modification guard and delta application
4. **Tests**: Permission integration tests for share/list/revoke flows
5. **Frontend**: Share UI in case settings (user search + role assignment) — separate issue

vincent-audibert-whoz:
## Summary

Adds case sharing functionality to AgentOS — two new endpoints in `CaseController` for managing per-case user permissions.

Closes #1011

## Endpoints

| Method | Path | Auth | Response |
|--------|------|------|----------|
| `PUT` | `/api/cases/{caseId}/share` | WRITE on Case | `200` + `List<UUID>` of applied userIds |
| `GET` | `/api/cases/{caseId}/users` | READ on Case | `200` + `List<CaseUserListItem>` |

### PUT /share — Delta semantics
- Batch request with `CaseShareEntry` list: each entry is `(userId, role?)` where role is ADMIN, MEMBER, or null (revoke)
- Self-modification entries silently filtered (caller can't demote/revoke themselves)
- Non-existent userIds silently skipped (Cypher MATCH filters them)
- Returns only the userIds for which the operation was actually applied

### GET /users — List direct relations
- Returns users with direct `[:ADMIN]` or `[:MEMBER]` edges on the case
- Super-admins without a direct relation are not listed

### Permission layer — `applyShareBatch`
New batch operation across the full permission stack (`PermissionService` → `PermissionRepository` → `PermissionNodeNeo4jRepository`):

- **3 Cypher batch queries**: `batchGrantAdmin`, `batchGrantMember`, `batchRevoke`
- **Single round-trip per role group** (at most 3 round-trips for a mixed batch, not N per user)
- Role transitions (ADMIN↔MEMBER) are simple delete + merge — no property-copying logic since `starred` will move to a dedicated `[:STARRED]` relationship (see linked issue)
- Generic method signature — reusable for any entity type

### SDK DTOs
- `CaseRole` — enum ADMIN/MEMBER
- `CaseShareEntry` — (userId, role?)
- `CaseShareRequest` — validated list of entries
- `CaseUserListItem` — user info + role (mirrors `NamespaceUserListItem`)

## Tests

8 MVC integration tests against embedded Neo4j covering:
- MEMBER and ADMIN grants (200 + applied userIds)
- Revoke with no prior relation (200 + empty)
- Empty entries list (200 + empty)
- Missing case (404 via `@HideOnAccessDenied`)
- Non-existent userId silently skipped (200 + empty)
- GET users on existing case (200 + array)
- GET users on missing case (404)

## Note

OpenAPI spec regeneration needed after merge to surface the new endpoints in the TypeScript client.


vincent-audibert-whoz (commented):
The PR is structurally sound and the permission-layer stack (`PermissionService` → `PermissionRepository` → `PermissionNodeNeo4jRepository`) is extended consistently. The Cypher batch queries are correct, the self-modification guard is in the right place, and the test coverage is solid. Three issues need to be addressed before merge.

**Security — ambiguous self-modification guard (`CaseController.shareCase`).** The filter `entry.userId != currentUserId` blocks *all* self-modifications. The KDoc says "prevent accidental self-demotion or self-revocation", implying self-promotion should be allowed — but the code also silently drops a caller's self-promotion entry. If blocking all self-modifications is intentional (the safer choice), the comment must say so explicitly. If self-promotion should be allowed, the filter is over-broad and would silently swallow a legitimate ADMIN self-grant for a MEMBER caller. The intent must be documented precisely; as written it creates a latent confusion that could lead to a future incorrect relaxation of the guard.

**Correctness — `listCaseUsers` makes two sequential Neo4j round-trips where one suffices.** `listUsersWithPermission(…, ADMIN)` followed by `listUsersWithPermission(…, MEMBER)` issues two Cypher queries. The existing `findDirectRelations` query already retrieves user id + relation type in a single round-trip using the `u.id + '|' + type(r)` encoding pattern. A new `findUsersWithRoleByEntity` query on `PermissionNodeNeo4jRepository` using the same encoding would eliminate the second query and the post-join role assignment. The current double-query approach is functionally correct but inconsistent with the single-query pattern established elsewhere in the repository, and it doubles the Neo4j load for every call to `GET /users`.

**API contract gap — `shareCase` and `listCaseUsers` not declared in `CaseApi`.** Every other endpoint on `CaseController` is declared in the SDK's `CaseApi` interface so external Feign clients can bind against it. The two new endpoints are added only to the controller, breaking this contract. Any Java/Kotlin consumer using the SDK cannot call these endpoints through the interface. Add `shareCase(caseId: UUID, request: CaseShareRequest): List<UUID>` and `listCaseUsers(caseId: UUID): List<CaseUserListItem>` to `CaseApi` with `@Operation` annotations consistent with the existing declarations.

**Additional note — `batchGrantAdmin`/`batchGrantMember` silently drop `starred` on role transitions.** The single-line Cypher queries delete the old relation and MERGE the new one without `SET newRel += properties(old)`, unlike the existing `promoteAdminToMember`/`demoteAdminToMember` queries. The PR description acknowledges this as a deliberate trade-off pending `[:STARRED]` migration. A `// TODO` comment referencing the linked issue should be added to the two batch queries to prevent this silent data loss from being overlooked in production before the migration is complete.

vincent-audibert-whoz:
## Review response

Thanks for the review. Here's the disposition of each point:

### 🔴 1. Self-modification guard — comment ambiguity
**Fixed.** The KDoc now reads *"prevent any self-modification (promotion, demotion, or revocation)"* which matches the actual behavior of the filter. The intent is to block all self-targeting entries, not just demotion/revocation.

### 🟡 2. Two round-trips in `listCaseUsers`
**Acknowledged, deferred.** A single `findUsersWithRoleByEntity` query would be cleaner. The current approach is functionally correct — this is an optimization for a follow-up.

### 🔴 3. Endpoints missing from `CaseApi`
**Not applicable.** `CaseApi` is the SDK contract for external Feign clients. Several existing controller-only endpoints (`starCase`, `unstarCase`, `listMineByParent`) deliberately stay out of `CaseApi` because they are internal UI operations, not part of the external API surface. `shareCase` and `listCaseUsers` follow the same established pattern.

### 🟡 4. TODO on batch queries re: starred
**Fixed.** Added a TODO comment on `batchGrantAdmin` referencing the linked issue for the `[:STARRED]` migration, noting that `promoteMemberToAdmin` / `demoteAdminToMember` can be simplified the same way once that's done.

fdelsert-whoz (approved):

PR #1146, opened 2026-07-15
feat: wz-32968 agent prompts

fdelsert-whoz:
This PR extends the AgentOS prompt system with two new capabilities:
1. **Linking a prompt to an `AgentConfig`** — making it act as a "starter" for that agent
2. **Storing opaque external metadata** on a prompt — for external consumers like Copilot or Studio

It also refactors the prompt listing endpoints into a cleaner, more expressive API.

---

#### 1. New `agentConfigId` field on prompts
- A prompt can now be optionally linked to an `AgentConfig` at creation time. This link is **immutable post-creation**.
- In Neo4j, the link is materialised as a `(:Prompt)-[:BELONGS_TO]->(:AgentConfig)` edge.
- A Neo4j index `prompt_agent_config_id` is created at startup.
- **Cascade delete**: when an `AgentConfig` is soft-deleted, all its linked prompts are soft-deleted too.

#### 2. New `externalMetadata` field (opaque JSON)
- An arbitrary `Map<String, Any?>` stored as-is by AgentOS, without interpretation.
- Allows external consumers (Copilot, Studio) to store their own data (labels, triggers, sections, etc.).

#### 3. Prompt listing endpoint refactor

| Before | After |
|---|---|
| `GET /api/prompts?scope=PLATFORM\|NAMESPACE\|USER\|USER_NAMESPACE` | `POST /api/prompts/search` with `PromptSearchRequest` body |
| `GET /api/prompts/by-parentId/{parentId}` | Removed, replaced by `/search` |
| `GET /api/prompts/effective?namespaceId=UUID` | `POST /api/prompts/effective` with `PromptEffectiveRequest` body |

Scope is now expressed via the `(namespaceId?, userId?)` combination in the request body — no more `OverlayScope` enum in query params.

Both `/search` and `/effective` also accept `namespaceExternalId` (and `userExternalId` for `/effective`) as alternatives to UUID fields — the server resolves them internally.

#### 4. `agentConfigId` propagation in `PromptCommandParser`
- The parser now returns `ResolvedCommand(text, agentConfigId?)` instead of plain `String`.
- Each resolved line carries the `agentConfigId` of the prompt that produced it — **no inheritance** from parent to child prompts.
- In `CaseServiceImpl`, if a `ResolvedCommand` carries an `agentConfigId`, the text is automatically prefixed with `@agentName` (unless it already starts with `@`). Agent name lookups are cached per resolution to avoid redundant repository calls.

#### 5. Frontend (Angular)
- `PromptFormComponent` now exposes:
  - An optional `<select>` to link a prompt to an `AgentConfig` (read-only in edit mode if already set)
  - A monospace `<textarea>` to edit `externalMetadata` as JSON with client-side validation
- `PromptStateService` updated to use the new POST endpoints.
- `CaseChatComponent` and `CaseHomeComponent` now pass the current `userId` to `listEffective`.

---

### Testing
- Unit tests for `agentConfigId` propagation across all resolution scenarios (parent/child combinations, unknown sub-prompts, plain text)
- Integration tests for the new `/search` endpoint (scope levels, permission checks, admin bypass)
- `InMemoryPromptRepository` updated to implement `findByScope` and `softDeleteByAgentConfigId`


agentos/agentos-service/src/main/resources/application.yml
```diff
@@ -4,6 +4,13 @@ spring:
   jackson:
     mapper:
       sort-properties-alphabetically: true
+  web:
+    resources:
+      # AgentOS is a pure API service — no static resources are served.
+      # Without this, Spring's ResourceHttpRequestHandler intercepts paths
+      # containing colons (e.g. /api/prompts:search) before the controller
+      # can handle them, returning 404 ("No static resource ...").
+      add-mappings: false
```
fdelsert-whoz: À supprimer

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/Prompt.kt
```diff
@@ -31,8 +31,20 @@ data class Prompt(
     override val metadata: EntityMetadata = EntityMetadata(),
     val namespaceId: UUID? = null,
     val userId: UUID? = null,
+    /**
+     * Optional link to an [io.whozoss.agentos.agentConfig.AgentConfig].
+     * When set, this prompt acts as a Starter for that agent.
+     * Immutable post-creation — preserved on PUT by the controller.
+     * Materialised as a `(:Prompt)-[:DEPLOYED_TO]->(:AgentConfig)` edge in Neo4j.
```
fdelsert-whoz: Devrait plutôt être BELONGS_TO ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
```diff
@@ -97,99 +97,92 @@ class PromptController(
...
-                if (!granted) throw AccessDeniedException("Cannot read prompts in namespace $namespaceId")
-                promptService
-                    .findByUserId(currentUser.id)
-                    .filter { it.namespaceId == namespaceId }
-            }
-        }.map(::toDto)
+        return promptService
+            .findByScope(
+                namespaceId = resolvedNamespaceId,
+                userId = request.userId,
```
selimbensenouciep-prog, line 144: `search()` forwards `request.userId` straight to `findByScope` with no check that it matches the authenticated user, and the query below filters only on `p.userId = $userId`. Any authenticated user can `POST {"userId": "<other-user-id>"}` and read that user's private prompts (full content, parameters, externalMetadata); with READ on a shared namespace, `{"namespaceId": ns, "userId": <victim>}` exposes their user×namespace prompts too. This is a regression from the old `GET ?scope=USER`, which always used `currentUser.id`, whereas `effective()` right below already enforces `userId == currentUser.id`. Could we mirror that here (reject or 403 when `request.userId != currentUser.id`, unless the caller is admin)? Worth restoring a permission test where alice searches with bob's id.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/Neo4jAgentConfigRepository.kt
```diff
@@ -50,7 +51,8 @@ open class Neo4jAgentConfigRepository(
             ?.takeIf { it.removed != true }
             ?.let { node ->
                 neo4jRepository.save(node.copy(removed = true))
-                logger.debug { "[Neo4jAgentConfigRepository] Soft-deleted agent config $id" }
+                promptNodeNeo4jRepository.softDeleteByAgentConfigId(id.toString())
```
selimbensenouciep-prog: The agent soft-delete and the prompt cascade are two separate writes with no transaction on the call path, so a failure on the second leaves the agent removed while its prompts stay live and orphaned (and a retry short-circuits on `takeIf { it.removed != true }`, so it can't self-heal). One catch worth flagging: annotating this method with `@Transactional` wouldn't help, because this repository is hand-constructed in `Neo4jPersistenceConfiguration` and never Spring-proxied. That is also why the existing `@Transactional` on `deleteByParent` is already inert. The transaction needs to sit on a proxied layer such as `AgentConfigServiceImpl.delete`. Separately, this reaches straight into `PromptNodeNeo4jRepository`, whereas the PR already added `PromptRepository.softDeleteByAgentConfigId` for exactly this, which would keep the aggregate boundary clean.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
```diff
@@ -264,13 +256,15 @@ class PromptController(
 
         val target =
             Prompt(
-                metadata = EntityMetadata(id = UUID.randomUUID()),
+                metadata = EntityMetadata(id = resource.id ?: UUID.randomUUID()),
                 namespaceId = resolvedNs,
                 userId = resolvedUser,
+                agentConfigId = resource.agentConfigId,
```
selimbensenouciep-prog, line 267: `agentConfigId` is persisted as-is here with no validation, while the same method 404s on a missing `namespaceId`. A nonexistent id is accepted with 201: the `MATCH ... MATCH ... MERGE` link is a silent no-op when the AgentConfig node is absent, so the `BELONGS_TO` edge never gets created, and since the field is immutable on PUT the dangling link can't be fixed. There is also no same-scope check, so a prompt can link to an agent in another namespace, which `CaseServiceImpl.resolvedText` then resolves into an `@name` prefix, and whose deletion cascades across namespaces (the cascade query has no scope filter). Could we validate at creation that the AgentConfig exists, isn't removed, and belongs to the prompt's namespace or platform scope?

libs/agentos-ui/src/lib/components/prompt-form/prompt-form.component.html
```diff
@@ -41,6 +41,27 @@ <h1 class="prompt-form__title">{{ isEditMode() ? 'Edit prompt' : 'New prompt' }}
...
+      <!-- Agent Config (optional, immutable post-creation) -->
+      @if (availableAgentConfigs().length > 0) {
+        <div class="prompt-form__field">
+          <label class="prompt-form__label" for="prompt-agent-config">Agent Config</label>
+          @if (isEditMode() && agentConfigIdControl.value) {
+            <p class="prompt-form__readonly-value">{{ resolveAgentConfigName(agentConfigIdControl.value) }}</p>
+            <span class="prompt-form__field-hint">Agent Config cannot be changed after creation.</span>
+          } @else if (!isEditMode()) {
+            <select id="prompt-agent-config" class="prompt-form__select" [formControl]="agentConfigIdControl">
+              <option [value]="null">&mdash; None &mdash;</option>
```
selimbensenouciep-prog: With a `[formControl]` select, `[value]` writes the raw DOM string, so binding it to `null` produces the string `"null"`. Re-selecting "None" after picking an agent therefore sets the control to the truthy string `"null"`, and `submit()` sends `agentConfigId: "null"`, which 400s server-side (the DTO field is a `UUID?`). Using `[ngValue]` registers the typed value instead:
```suggestion
              <option [ngValue]="null">&mdash; None &mdash;</option>
```
(and `[ngValue]="agent.id"` on the `@for` options for consistency.)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/Neo4jAgentConfigRepository.kt
fdelsert-whoz: Quoi ? J'ai laissé passer ça !

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
fdelsert-whoz: done at commit: 607b136d8a48f85a30ca75ba92ddbfcc25f0c25d

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
fdelsert-whoz: Partially done at commit: 26620a9d9f5346a9f73d53d8be43fe9e719c9cca

The namespace and permission checks in search() will depend on the decisions made regarding prompt/starter overriding.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/Neo4jAgentConfigRepository.kt
fdelsert-whoz: done at commit: 68e78af10077a2ab7fced016b113f0f199a65bf4

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
fdelsert-whoz: done at commit: 11ccc37dc79fabc996a68fa369400880b0d189b2

Cross-namespace agentConfigId validation is now enforced.

The backend does accept linking a namespace prompt to a platform agent, platform agents are universally accessible by design.
The UI however only lists same-level agents in the select, so this valid combination isn't reachable from the form today. Worth deciding later whether to merge platform agents into the namespace-level select, or leave it as-is if the use case doesn't arise in practice.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
```diff
@@ -24,9 +27,15 @@ import java.util.UUID
 @Service
 class PromptServiceImpl(
     private val repository: PromptRepository,
+    private val agentConfigService: AgentConfigService,
 ) : PromptService {
     override fun create(entity: Prompt): Prompt {
         validate(entity)
+        if (entity.agentConfigId != null) {
+            val agentConfig = agentConfigService.findById(entity.agentConfigId)
```
selimbensenouciep-prog, line 36: Small regression from the new existence check: namespace agents defined on the filesystem (YAML) are merged into the dropdown by `findByParent` with a synthetic id, but `findById` here resolves through `findByIds`, which `FilesystemAgentConfigRepository` does not override, so it only hits Neo4j. Selecting a filesystem agent that the form legitimately offered now fails on create with 404 "AgentConfig not found", where before it saved (just without the edge). Could the existence check go through a lookup that includes filesystem agents (the same path `findByParent` uses), so the dropdown and the validation agree?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
```diff
@@ -108,6 +123,23 @@ class PromptServiceImpl(
...
+            // Platform agent is always valid (accessible from any scope)
+            agentConfig.namespaceId == null -> true
+            // Same namespace
+            agentConfig.namespaceId == prompt.namespaceId -> true
+            // Everything else is cross-scope
+            else -> false
+        }
+        if (!validScope) {
+            throw BadRequestException(
+                "AgentConfig '${agentConfig.name}' (id=${agentConfig.id}) belongs to a different namespace " +
```
selimbensenouciep-prog: `findById` does no permission scoping, so this message can echo the name and namespace of an agent in a tenant the caller cannot access (a random id gives "not found", an existing foreign id gives a 400 that names it). Could we keep the message generic and reference only the id the caller supplied, instead of the foreign agent's name and namespace? That keeps the cross-namespace rejection without turning it into an identity oracle.

selimbensenouciep-prog (commented):
Re-reviewed after the fix commits. The four earlier findings are addressed correctly: the `search` cross-user guard closes the IDOR (verified no bypass, maps to 403), the cascade delete is now transactional on the proxied `AgentConfigServiceImpl` and back on the `PromptRepository` abstraction, `agentConfigId` is validated for existence and namespace scope on create, and the select uses `[ngValue]`. Excluding prompts linked to disabled or removed agents from `findEffective` is a good addition, and the new tests cover the security guard and the validation matrix.

Two small regressions came in with the `agentConfigId` validation (inline below). A few test and doc nits as well: the "excludes prompts linked to deleted agents" test passes because the prompt cascade already soft-deletes the prompt, so it does not actually exercise the removed-agent branch of the new `findEffective` query; the super-admin search test asserts only status 200, not that the other user's prompt is returned; there is no PUT test for `agentConfigId` immutability; and the `PromptApi.search` KDoc still says "admin view only" although search is now available to any authenticated user.

selimbensenouciep-prog (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
fdelsert-whoz: solved at commit: 41789b615cbd8084eb0871e74ca4fce91b909ff0

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
fdelsert-whoz: Filesystem agents cannot be linked to a prompt at all, doing so would create a dangling BELONGS_TO edge in Neo4j that silently hides the prompt from findEffective. We overrode findByIds in FilesystemAgentConfigRepository so they are resolvable for read purposes, but PromptServiceImpl now explicitly rejects them with a 422 when supplied as agentConfig.

done at commit: 8a84e0f15ff199c579b4122d3a0b9a7c1aeae9dd

selimbensenouciep-prog (approved):

PR #1149, opened 2026-07-16
feat: wrap error message when too long + force delegation instead of redirection

romain-petit-whoz:
# Force delegation over redirection
Agents now use delegation instead of redirection when all target agents are delegatable. The redirect tool is filtered to exclude agents that can be reached via delegation, and is removed entirely when no redirect-only agents remain. This ensures sub-thread isolation and better traceability of agent interactions.


libs/integrations/ai/src/lib/ai.tools.ts
```diff
@@ -66,17 +66,22 @@ AVOID closed options unless the user explicitly needs to choose between specific
...
-      const agentSummaries = this.agentSummaries()
-        .map((a) => `  - ${a.name} : ${a.description}`)
-        .join('\n')
-      const redirectTool: FunctionTool<{ query: string; agentName: string }> = {
-        type: 'function',
-        function: {
-          name: `${this.name}__redirect`,
-          description: `Redirect the current query to another available agent among:
+      // If all agents are delegatable, skip redirect entirely
+      if (delegateAgentNames !== '__ALL__') {
```
vincent-audibert-whoz: Behavior change seems too drastic, but more linked to the possibility to delegate to all agents. That is in itself an agent design issue, do you have examples that would apply ?

libs/agent/src/lib/toolbox.ts
```diff
@@ -200,6 +200,16 @@ export class Toolbox implements Killable {
     // Combine all factories
     allFactories.push(...mcpFactories)
 
+    if (integrations?.has(DelegateTools.TYPE)) {
+      const delegateAgentNames = integrations.get(DelegateTools.TYPE)!
+      if (delegateAgentNames.length === 0) {
+        // Empty allow-list means all agents are delegatable
+        context.data = { ...context.data, delegateAgentNames: '__ALL__' }
```
vincent-audibert-whoz: `context.data` is a garbage object, ugly as hell but will work by the magic of js 💩 😆

apps/server/src/lib/message.routes.ts
```diff
@@ -273,23 +275,34 @@ export function registerMessageRoutes(
...
+          const thread = await threadService.getThread(<string>projectName, <string>threadId)
+          if (!thread) {
+            res.status(404).send(`Thread '${threadId}' not found`)
+            return
+          }
+          if (!hasAccess(thread, username)) {
+            res.status(403).send('Access denied: thread belongs to another user')
+            return
+          }
+          event = thread.getEventById(eventId)
```
vincent-audibert-whoz: ❤️

vincent-audibert-whoz (dismissed):
Redirect masked by full delegation seems like an over-reach, but concerns only a narrow case that can be considered a design agent issue.

PR #1150, opened 2026-07-17
#1023 new navigation on coday part chat

benjamin-valdes-whoz:


emilie-dewintre-whoz (approved):

libs/design-system/src/styles/_surfaces.scss
```diff
@@ -117,3 +117,38 @@
     color: var(--color-text, #f5f5f7);
   }
 }
+
+// ── Chat bubbles ──────────────────────────────────────────────────────────────
+
+@mixin surface-bubble-agent {
```
selimbensenouciep-prog, line 116: These four mixins hardcode their colors, whereas every other mixin in this file resolves through `var(--token, fallback)`. The header says "All mixins use CSS custom properties with fallbacks", and the `surface-mark` note just above states `#161616`/`#fff` are the only hardcoded values allowed in the design system, so this reintroduces exactly what that note rules out. A few have exact token equivalents (`#1d1d1f` = `--color-text`, `border rgba(0, 0, 0, 0.1)` = `--color-border`, dark `#f5f5f7` = `--color-text`). Could we route those through `var(--color-text, #1d1d1f)` and keep the bespoke tints (`#f0f0ee`, `#0f1115`, `#e8eaf0`, `#3a3a3c`) as the fallback inside the `var()`, the way `surface-glass`/`surface-dropdown` already do? That keeps the bubbles responsive to token overrides.

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
```diff
@@ -23,49 +23,35 @@
...
+                <span class="case-chat__role-label">{{ agentDisplayName() }}</span>
+                <div class="case-chat__bubble case-chat__message-text--markdown" [innerHTML]="streamingHtml()"></div>
               </div>
             } @else if (item.kind === 'technical') {
-              <div
-                class="case-chat__technical"
-                [class.case-chat__technical--warn]="item.item.type === 'WarnEvent'"
-                [class.case-chat__technical--error]="item.item.type === 'ErrorEvent'"
-              >
+              <details class="case-chat__log">
```
selimbensenouciep-prog, line 49: Now that the technical branch renders a native `<details>`, the old custom toggle is gone but its logic stayed behind: `toggleTechnical()` (~line 787), `isTechnicalExpanded()` (~796) and the `expandedTechnicals` signal (~209) no longer have any caller. The streaming timeline item's `text` field is also unread now that the bubble binds `streamingHtml()`. Could we drop them to keep the component lean?

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.scss
```diff
@@ -382,6 +388,111 @@
...
+
+      a {
+        color: rgba(255, 255, 255, 0.85);
+      }
+    }
+  }
+
+  // ── System log ───────────────────────────────────────────────────────────────
+
+  &__log {
```
selimbensenouciep-prog, line 349: This `&__log` block replaces the old `&__technical*` rules, but those are still in the file (~lines 289-355) and no longer match anything after the class rename. Same for the bare `&__message` (~499) and `&__message--streaming` (~583), now superseded by `&__turn` / `&__turn--streaming`. Note `&__messages` and `&__message-text--markdown` are still used and should stay. Could we drop the orphaned blocks in this PR while we're here?

selimbensenouciep-prog (dismissed):

selimbensenouciep-prog (approved):

PR #1148, opened 2026-07-16
feat: readAsImage tool to let agents read images, PDFs and PPTX visually

selimbensenouciep-prog:
## What

New `readAsImage` file tool so agents can visually read binary documents from the file exchange (or any FILE_ACCESS root): images, PDFs (one image per page) and PPTX presentations (one image per slide). The rendered pages ride as base64 `MessageContent.Image` attachments next to the textual tool summary, and reach the model as native image blocks.

Follows the team decision to start with the simple case (a few-page document) and to unify image and PDF reading behind a single dedicated tool, separate from `readFile`.

## How it works

- SDK (additive): `ToolExecutionResult.images` plus a `successWithImages` factory, and `ToolResponseEvent.images`. Existing factories keep their binary signature, legacy Neo4j nodes deserialize to an empty list, the OpenAPI change is additive.
- Plugin: `ReadAsImageTool` joins `readTools`, so it is exposed on FILE_ACCESS configs and on the case/namespace exchange grants with the usual per-agent allowlist, `BoundaryPathResolver`, deny patterns and size caps. Images are bounded to 1024px and re-encoded JPEG q80 (small originals pass through untouched). PDF pages are rendered with PDFBox at 150 DPI with lazy parsing; PPTX slides are rendered directly with POI at 2x supersampling. At most 10 pages/slides per call; the 1-based `pages` parameter selects a subset and the summary tells the model how to continue on longer documents.
- Delivery to the LLM (AgentAdvanced only): Spring AI tool responses are text-only, so images are injected at prompt-build time as a follow-up user message carrying `Media`, which the four chat models map natively. Context cost stays bounded: images are costed inside the 300k-char detailed-tool budget and only the 20 most recent images stay attached, older responses keep a text marker. `CompressingChatClient` now preserves media while compressing text, and base64 never enters any text field by construction.
- Packaging: pdfbox 3.0.7 and poi-ooxml 5.5.1 are bundled in the plugin jar (PF4J APD, same pattern as the MCP plugin); the jar grows to ~23 MB, with no Jackson or Kotlin classes bundled. `XMLSlideShow` is constructed directly on purpose: `SlideShowFactory` resolves providers through the TCCL, which cannot see bundled classes under PF4J.

## Known limitations

- AgentSimple is not wired: a simple agent calling the tool gets the textual summary only (its Spring AI ToolCallback contract returns a plain string). Follow-up work.
[...]

PR #1151, opened 2026-07-17
feat: readSpreadsheet file tool reading xlsx sheets as CSV text

selimbensenouciep-prog:
Adds a `readSpreadsheet` tool to the file plugin so agents can read Excel workbooks. Spreadsheets are data, not visuals: the tool projects sheets to CSV text, giving the model exact cell values at minimal token cost (no image rendering, which POI cannot do for xlsx anyway).

Stacked on #1148 (base branch `feat/selim/read-as-image-tool`, the description cross-references depend on it); to retarget to `master` once #1148 merges.

## What it does

- One block per selected sheet: a `## Sheet 1/3 "Name": rows 1-1000 of 5432, columns A-H` header, then RFC 4180 CSV (`\n` record ends, cheaper tokens; documented deviation).
- Values are rendered as displayed in Excel via `DataFormatter` (number and date formats applied). Formula cells return their cached last-saved result, never a `FormulaEvaluator` run (latency/CPU trap on large workbooks); error cells give Excel literals (`#DIV/0!`, `#REF!`...).
- Pagination: `sheets` (1-based) and `startRow` (single-sheet windowing). Per-call budgets: 1000 rows, 100k output chars, 256 columns, 5k chars per cell, with a `[truncated]` notice telling the model how to continue.
- Physical row numbering (aligned with what Excel shows, stable for `startRow`): interior empty rows kept as empty lines, trailing style-only rows and cells trimmed (a sheet formatted down to row 1048576 does not blow the output). Hidden sheets included and marked `(hidden)`; merged regions keep their value in the top-left cell.
- Same guardrails as the sibling tools: `BoundaryPathResolver`, size cap before parsing, POI `ZipSecureFile` defaults, 60s timeout, `logger.warn` in the generic catch.
- `XSSFWorkbook` is constructed directly, not via `WorkbookFactory` (under PF4J the TCCL does not see the bundled POI classes), same pattern as `XMLSlideShow` in #1148.
- No new dependency: `poi-ooxml` (XSSF included) was already bundled in the plugin jar.

## Routing

`readFile` and `readAsImage` descriptions now point .xlsx to `readSpreadsheet`; `readSpreadsheet` points `.csv` back to `readFile` and rejects `.xls`/`.xlsm`/`.ods` with a convert-to-xlsx hint.

## Scope notes

V1 is .xlsx only, mirroring the pptx-only scope of #1148 (.xls/.ods deferred until a need shows up). The whole workbook is materialized in memory (XSSF DOM): the KDoc documents that raising `readMaxSizeMb` raises the parse memory cost; a SAX streaming reader is the escape hatch if OOM is ever observed.

## Testing

- `ReadSpreadsheetToolSpec`: 35 tests covering CSV quoting, number/date formats, cached formulas (including staleness and cached errors), merged regions, hidden sheets, empty rows/sheets/workbooks, the three truncation modes, sheet/row selection validation, encrypted/corrupt files, size cap, deny patterns, traversal and extension routing.
- `FileToolProviderSpec`: tool registered in read-write and read-only sets, `readMaxSizeMb` and `extraDenyPatterns` propagation.
- Full `agentos-file-plugin` suite green.


agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
```diff
@@ -0,0 +1,505 @@
...
+                "File exceeds maximum size of ${readMaxSizeBytes / 1024 / 1024} MB: ${params.filePath}",
+            )
+        }
+
+        when (val extension = resolved.name.substringAfterLast('.', "").lowercase()) {
+            "xlsx" -> Unit
+            "csv" -> throw UnsupportedFormatException("'.csv' is plain text: use readFile instead.")
+            else -> throw UnsupportedFormatException(
+                "Unsupported file type '.$extension'. Supported: .xlsx only. " +
+                    "Convert .xls/.xlsm/.ods to .xlsx; for .csv use readFile.",
```
thomasmartin-whoz: since csv has its own branch, no need to write about it in the else branch

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        val selection = selectSheets(params.sheets, sheetCount, fileName)
+        if (params.startRow != null && params.startRow < 1) {
+            throw IllegalArgumentException("\"startRow\" must be at least 1 (got ${params.startRow}).")
+        }
+        if (params.startRow != null && selection.size != 1) {
+            throw IllegalArgumentException(
+                "\"startRow\" requires targeting exactly one sheet (pass a single entry in \"sheets\").",
+            )
```
thomasmartin-whoz: tous les `throw IllegalArgumentException` peuvent être remplacés par des `require` en kotlin

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
```diff
@@ -0,0 +1,505 @@
...
+
+            if (nextStartRow != null && nextStartRow <= totalRows) {
+                appendTruncationNotice(
+                    out,
+                    workbook,
+                    continueSheets = listOf(sheetNumber),
+                    nextStartRow = nextStartRow,
+                    notShown = selection.drop(position + 1),
+                )
+                break
```
thomasmartin-whoz, line 288: all those `break` and `continue` are a pain to read. A functional style would be better.

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
```diff
@@ -0,0 +1,505 @@
...
+
+    /**
+     * Used range of a sheet, scanning only physically-existing rows (bounded by what POI
+     * parsed): trailing style-only rows are excluded from the row total, and the column
+     * count covers the whole sheet so it stays stable across paged calls.
+     */
+    private fun scanUsedRange(sheet: Sheet): UsedRange {
+        var lastContentRow = -1
+        var columnCount = 0
+        var columnsTruncated = false
```
thomasmartin-whoz: lots of mutation here

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
```diff
@@ -0,0 +1,505 @@
...
+     * Formula cells read the cached last-saved result: without an evaluator,
+     * [DataFormatter.formatCellValue] would return the formula text itself. A formula
+     * saved without any cached result renders as `=formula` (POI would otherwise report
+     * a fabricated numeric 0).
+     */
+    private fun renderCell(cell: Cell, formatter: DataFormatter, date1904: Boolean): String =
+        when (cell.cellType) {
+            CellType.FORMULA -> when {
+                !hasCachedFormulaResult(cell) -> runCatching { "=${cell.cellFormula}" }.getOrDefault("")
+                else -> when (cell.cachedFormulaResultType) {
```
thomasmartin-whoz: three nested when statements is painful. Please introduce methods to avoid that

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
selimbensenouciep-prog: Trimmed the csv hint from the else branch. csv is intercepted by its own branch above, so the else path is only reached by genuinely unsupported extensions (.xls, .ods, …) where that advice was unreachable. The dedicated csv message stays.

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
selimbensenouciep-prog: Converted the validation throws to `require` (size cap, no-sheet, the three startRow checks, empty / out-of-range sheets). `require` throws IllegalArgumentException, so `execute()`'s catch still maps them to INVALID_INPUT — messages and behavior unchanged (spec green). Left the `throw IllegalArgumentException` inside the catch blocks as-is: they translate POI exceptions, there is no precondition to assert (and `check` would surface as IllegalStateException → READ_ERROR instead of INVALID_INPUT).

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
selimbensenouciep-prog: Extracted the inner row loop into `renderSheetRows(...)` returning a small `SheetRows` result; that lifts the two inner `break`s and a nesting level out of `renderWorkbook`, so the outer loop is now linear (budget check → render → append → optional truncation notice). I did not go fully functional on purpose: the budget accumulation (rows *and* chars) with early exit and the paging cursor, plus the single-oversized-row special case, do not fold into `sequence`/`fold` without a carried state that reads worse than the loop and would risk the well-tested truncation/paging behavior. Char accounting is unchanged (`out.length`, header still appended after), spec stays green (46 tests). Two outer `break`s remain (budget exhausted / sheet-level truncation) and one `continue` (empty sheet) — genuine early exits, now at a single level.

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
selimbensenouciep-prog: Rewrote `scanUsedRange` as a single `fold` over the rows with an immutable `UsedRange` accumulator — no more vars. Kept it one pass / O(1) extra memory on purpose (a `mapNotNull().toList()` would have materialized a pair per content row, a regression on large sheets); behavior is identical (the fold reproduces the per-row running max), spec green.

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadSpreadsheetTool.kt
selimbensenouciep-prog: Extracted the FORMULA branch into `renderFormulaCell(...)`: `renderCell` is now a flat 3-arm `when` (FORMULA → helper, ERROR → literal, else → `formatCellValue`), and the formula handling is a single `when` in its own method with an early return for the no-cached-result `=formula` case. Moved the formula KDoc onto the new method. No behavior change, spec green (46 tests).

thomasmartin-whoz (approved):

PR #1152, opened 2026-07-17
feat: chat composer file attachments (drag & drop, + button, exchange upload)

selimbensenouciep-prog:
Brings ChatGPT-style file attachments to the AgentOS chat input, in both composers: the case chat and the new-case home screen.

## What it does

- **Drag & drop** files anywhere on the chat (overlay "Drop files to attach"; non-file drags are ignored), or click the new **"+" button** at the left of the input to open the native picker.
- Staged files show as **removable chips** above the textarea: kind icon (reuses `getFileIcon`), middle-truncated name, extension label, X button, per-file error line. Up to 10 files per message; the cap is enforced across successive adds with a visible message.
- On submit, files are **uploaded to the case exchange first**, then the message is sent with a trailing mention so the agent knows what to read (messages have no attachment field; the exchange is the transport):
  `[Files attached to the case exchange: report.pdf, data.xlsx]`
- **Namespace targeting on explicit request**: when the message contains the word "namespace" and the user can write the namespace exchange (server-computed manifest capability, fail-closed, same gate as the drawer), the upload goes to the namespace exchange instead. A `namespace` badge on the chips previews the resolved target while typing. The heuristic is deliberately naive (documented in code); the badge is the safety net.
- **Failure handling**: uploads run sequentially and continue on failure (one duplicate does not skip the batch); any failure blocks the message send, keeps the typed text, and shows the mapped server error on the chip (type not allowed / name already exists / too large). A retry skips the files already uploaded, so nothing is double-uploaded.
- **Home screen flow**: create case, upload to the fresh case id, send the first message, navigate. The created case id is remembered so a retry after a failed upload reuses it instead of creating a duplicate case.

## Design notes

- New `composer-attachments` trio under `libs/agentos-ui`, shared by both composers: pure utils, a component-provided `ComposerAttachmentsService` (isolated instance per composer; delegates uploads to `ExchangeStateService.uploadFile`, reusing its error mapping and manifest refresh), and a signals-first presentational chip component (OnPush, `input()`/`output()`/`viewChild()`).
- Only service change: `ExchangeStateService.initializeForNamespace(namespaceId)` so the home screen can gate the namespace option before any case exists.
- No backend change: existing upload endpoints, allow-list, size caps and capability model are reused as-is.

## Testing

- New Jest specs: utils (heuristic, mention formats, truncation), service (staging, cap, drag & drop semantics, upload orchestration incl. retry-skips-uploaded), chip component (rendering, outputs, hidden input mechanics), and the first specs for `case-chat` and `case-home` (upload-then-send ordering, mention content, failure blocks send, namespace routing, case-reuse on retry). 187 tests green on `nx test agentos-ui`; `nx lint` and a full client AOT build pass.
- Manual run through the dev stack: attached a file on the home screen, chip rendered, case created, file landed in the case exchange, message ended with the mention, and the agent read the file via `case-exchange__readFile` and summarized it.

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
+++ b/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
@@ -29,6 +29,7 @@ import {
   Configuration,
   EnrichmentPhaseTrace,
   ErrorEvent,
+  ExchangeFileEntryScopeEnum,
   IntentionGeneratedEvent,
   MessageEvent as CaseMessageEvent,
   ToolRequestEvent,
@@ -47,6 +48,9 @@ import { UserStateService } from '../../services/user-state.service'
 import { ExchangeStateService } from '../../services/exchange-state.service'
 import { exchangeMutationScope } from '../../services/exchange-content.utils'
 import { ExchangeShellComponent } from '../exchange-shell/exchange-shell.component'
+import { ComposerAttachmentsComponent } from '../composer-attachments/composer-attachments.component'
+import { ComposerAttachmentsService } from '../composer-attachments/composer-attachments.service'
+import { resolveUploadScope } from '../composer-attachments/composer-attachments.utils'
 
 export interface ToolCall {
   requestId: string
@@ -97,7 +101,15 @@ const SCROLL_BOTTOM_THRESHOLD = 64
  */
 @Component({
   selector: 'agentos-case-chat',
-  imports: [IconButtonComponent, JsonPipe, DrawerComponent, ExchangeShellComponent, PromptAutocompleteComponent],
+  imports: [
+    IconButtonComponent,
+    JsonPipe,
+    DrawerComponent,
+    ExchangeShellComponent,
+    PromptAutocompleteComponent,
+    ComposerAttachmentsComponent,
+  ],
+  providers: [ComposerAttachmentsService],
   templateUrl: './case-chat.component.html',
   styleUrl: './case-chat.component.scss',
 })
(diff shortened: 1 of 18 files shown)


libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -370,7 +399,12 @@ export class CaseChatComponent implements OnInit, OnDestroy {
   }
 
   protected get canSend(): boolean {
```
benjamin-valdes-whoz, line 401: submit() est maintenant async. Le guard if (!this.canSend) return s'exécute en début de submit(), mais n'est pas revérifié après le await uploadAllAndBuildMention(). Entre le check initial et le postMessage, l'utilisateur peut cliquer Send à nouveau. La 2e invocation passe le guard car isUploading() a déjà été clearé par le finally de la 1re. Résultat : deux postMessage concurrents avec le même contenu.

benjamin-valdes-whoz (dismissed):

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
selimbensenouciep-prog: Good catch, fixed in 597cfc3b. I added a component-owned `submitting` flag, set synchronously at the start of the async branch and cleared in a `finally`, the same way case-home already guards with `isCreating`. Re-entrancy no longer relies on `isUploading` (owned by the service and cleared in its finally before `postMessage` sets `isRunning`). A test covers it: a second submit() while one is in flight yields a single upload and a single postMessage.

benjamin-valdes-whoz (approved):

PR #1153, opened 2026-07-20
feat: readDocument tool reading .docx as Markdown with embedded images

selimbensenouciep-prog:
## What

New `readDocument` file tool so agents can read Word documents (.docx). A Word document is text
with structure, not a fixed-layout visual (and POI has no engine to rasterize a .docx, unlike a
PDF), so it is projected to **Markdown** rather than rendered to images: exact, quotable,
searchable text at minimal token cost. Sibling of `readAsImage` (#1148) and `readSpreadsheet`
(#1151), same file-plugin conventions.

Stacked on #1148 (base branch `feat/selim/read-as-image-tool`; it reuses `ToolExecutionResult.images`,
`ImageProcessor` and the `MAX_SOURCE_PIXELS` decode-bomb cap). Retarget to `master` once #1148 merges.

## How it works

- Walk `XWPFDocument.getBodyElements()` in document order → Markdown: headings (styleID `Heading`/`Titre` N)
  become `#`..`######`, list paragraphs become `-` bullets, tables become Markdown tables (`|` escaped,
  newlines flattened, cut at 64 columns / 5k chars per cell). Budget of 100k characters per call; a 1-based
  `startElement` pages through a long document, with a `[truncated]` notice telling the model how to continue.
- **Embedded pictures → vision, where it is actually useful and cheap:** `getAllPictures()` yields the embedded
  bitmaps directly (no rendering). Each is checked against `ImageProcessor.MAX_SOURCE_PIXELS` from its declared
  header before decoding, re-encoded JPEG and attached (up to 10) via `ToolExecutionResult.images`. An oversized
  or unreadable picture is skipped (the text still reads), never fatal.
- Same guardrails as the sibling tools: `BoundaryPathResolver`, size cap before parsing, POI `ZipSecureFile`
  defaults, 60s timeout, `logger.warn` in the generic catch, path-scrubbed error mapping. `XWPFDocument` is
  constructed directly from an `OPCPackage` (under PF4J the TCCL does not see the bundled POI classes), same
  pattern as `XSSFWorkbook`/`XMLSlideShow`.
- **No new dependency:** `XWPFDocument` ships in the already-bundled `poi-ooxml` / `poi-ooxml-lite` (the
  wordprocessing schemas are present in the lite jar). `.docx` only; legacy binary `.doc`, `.odt` and `.rtf`
  are rejected with a typed `UNSUPPORTED_FORMAT` error.

## Why not render the whole document like a PDF

Investigated and rejected. No JVM-embeddable docx→PDF converter is faithful (docx4j's own author: "for great
fidelity … you need Microsoft's own Office layout engine"): XDocReport (POI→PDF via iText) is the only bundleable
one but drags iText 2.1.7 (LGPL) with imperfect fidelity and is largely unmaintained; LibreOffice headless is a
~300 MB native binary outside the plugin jar (an infra change + process-spawning risk on a shared service). Either
way it would re-OCR already-perfect text, exploding tokens and losing citability. The tool description points users
to export-to-PDF (Word's own layout engine, highest fidelity, zero infra) + `readAsImage` for documents whose
layout IS the content.

## Known limitations

- No inline placement of image markers in the text flow (images are attached, listed after the body); precise
  run↔picture correlation is a follow-up.
- Headers/footers, footnotes and SmartArt are not extracted. Legacy binary `.doc` (HWPF / poi-scratchpad) is out
  of scope.

## Tests

`ReadDocumentToolSpec` (14) covers headings/lists/tables → Markdown with pipe escaping, an embedded picture
attached, an oversized embedded picture skipped while the text still returns, `.doc`/`.odt` → UNSUPPORTED_FORMAT,
password-protected/corrupt → INVALID_INPUT, size cap, deny patterns, path traversal, `startElement` paging +
out-of-range. `FileToolProviderSpec` updated for the tool counts (7→8 / 4→5). Whole plugin suite green.


agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadDocumentTool.kt
```diff
@@ -0,0 +1,443 @@
...
+        const val MAX_TABLE_COLUMNS = 64
+
+        /** A single table cell longer than this is cut. */
+        const val MAX_CELL_CHARS = 5000
+
+        /** Maximum embedded pictures attached as vision images per call, newest layout order. */
+        const val MAX_ATTACHED_IMAGES = 10
+
+        /** CommonMark supports at most 6 heading levels; deeper Word headings are clamped to this. */
+        private const val MAX_HEADING_LEVEL = 6
```
thomasmartin-whoz, line 435: to put in configuration keys like PDF

thomasmartin-whoz (dismissed):

agentos/agentos-file-plugin/src/main/kotlin/io/whozoss/agentos/plugins/file/tools/ReadDocumentTool.kt
selimbensenouciep-prog: Exposed readDocument's per-call budgets as FILE_ACCESS config keys following the readAsImage mechanism: `documentMaxOutputChars`, `documentMaxAttachedImages`, `documentMaxTableColumns`, `documentMaxCellChars` (const defaults kept, read in `provideTools`, passed to the constructor; propagation test added). Kept `MAX_HEADING_LEVEL` internal because it is a CommonMark hard limit (ATX headings only go from 1 to 6; `#######` renders as literal text), so it belongs with readAsImage's internal format consts rather than being a tunable knob. `MAX_CONCURRENT_DECODES` also stays internal, matching readAsImage's `MAX_CONCURRENT_RENDERS`. readSpreadsheet's equivalents are still `const` from #1151, and I'll align them in a follow-up on that branch for consistency.

thomasmartin-whoz (approved):

PR #1155, opened 2026-07-21
WZ-33587: add lastMessageAt property into case DTO

thomasmartin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
```diff
@@ -13,11 +15,34 @@ interface CaseEventNodeNeo4jRepository : Neo4jRepository<CaseEventNode, String>
      * Returns `e, r, c` so SDN maps the [CaseEventNode.case] @Relationship field.
      */
     @Query(
-        $$"""MATCH (e:CaseEvent)
-            WHERE e.caseId = $caseId AND (e.removed IS NULL OR e.removed = false)
-            OPTIONAL MATCH (e)-[r:BELONGS_TO]->(c:Case)
-            RETURN e, r, c ORDER BY e.timestamp ASC, e.id ASC
-            """,
+        "MATCH (e:CaseEvent) " +
+            "WHERE e.caseId = ${'$'}caseId AND (e.removed IS NULL OR e.removed = false) " +
```
AlanMenantWhoz: C'était obligatoire de modifier `${'$'}`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
```diff
@@ -13,11 +15,34 @@ interface CaseEventNodeNeo4jRepository : Neo4jRepository<CaseEventNode, String>
...
+     * The result is a list of maps each holding `caseId` (String) and `lastMessageAt` (Instant),
+     * returned as a single `collect(...)` column so SDN can map it without a custom converter.
+     *
+     * Used by [io.whozoss.agentos.caseFlow.CaseController] to enrich list responses with
+     * [io.whozoss.agentos.sdk.api.case.CaseDto.lastMessageAt] without storing the value
+     * on the Case node itself.
+     */
+    @Transactional(readOnly = true)
+    @Query(
+        "UNWIND ${'$'}caseIds AS cid " +
```
AlanMenantWhoz: Pareil ici, j'ai l'impression que c'est Coday qui ne sait pas écrire une variable avec un $

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/Neo4jCaseEventRepository.kt
```diff
@@ -55,12 +59,20 @@ open class Neo4jCaseEventRepository(
...
-    open override fun deleteByParent(parentId: UUID): Int {
+    override fun deleteByParent(parentId: UUID): Int {
         val active = caseEventNodeNeo4jRepository.findActiveByCaseId(parentId.toString())
         caseEventNodeNeo4jRepository.saveAll(active.map { mapper.withRemoved(it, true) })
         logger.debug { "[Neo4jCaseEventRepository] Soft-deleted ${active.size} events for case $parentId" }
         return active.size
     }
 
+    override fun findLastMessageTimestamps(caseIds: Collection<UUID>): Map<UUID, Instant> {
+        if (caseIds.isEmpty()) return emptyMap()
```
AlanMenantWhoz: C'est ok finalement les early return ? (Moi j'aime bien donc ça me dérangera pas)

thomasmartin-whoz: non, c'est coday qui aime bien :D . merci

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/Neo4jCaseEventRepository.kt
```diff
@@ -55,12 +59,20 @@ open class Neo4jCaseEventRepository(
...
+        val rows = caseEventNodeNeo4jRepository.findLastMessageTimestamps(caseIds.map { it.toString() })
+        return rows.associate { row ->
+            UUID.fromString(row["caseId"] as String) to (row["lastMessageAt"] as Instant)
```
selimbensenouciep-prog: The underlying `@Query` method (`CaseEventNodeNeo4jRepository.findLastMessageTimestamps`) returns `List<Map<String, Any>>`. For a raw `Map` projection SDN returns the driver's native types instead of running its entity converters, so a Neo4j zoned datetime comes back as `java.time.ZonedDateTime`, not `Instant`, and `row["lastMessageAt"] as Instant` throws.

I reproduced it on embedded Neo4j (save a `MessageEvent`, then call `findLastMessageTimestamps`): `java.lang.ClassCastException: class java.time.ZonedDateTime cannot be cast to class java.time.Instant` at this line. Both list endpoints (`listByParent` / `listMineByParent`) go through `withCallerMeta`, so they 500 as soon as any case in the namespace has a message.

The sibling `findDirectRelations` avoids this by only casting String/List/Boolean.

```suggestion
            UUID.fromString(row["caseId"] as String) to (row["lastMessageAt"] as java.time.ZonedDateTime).toInstant()
```

Two small follow-ups: the query KDoc still says `lastMessageAt (Instant)`, the same wrong assumption, so it is worth correcting; and this path is only covered by a MockK stub today, so a covering test on embedded Neo4j would have caught it.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
```diff
@@ -13,11 +15,34 @@ interface CaseEventNodeNeo4jRepository : Neo4jRepository<CaseEventNode, String>
...
+            "MATCH (msg:MessageEvent {caseId: cid}) " +
+            "WITH cid, max(msg.timestamp) AS lastMessageAt " +
+            "RETURN collect({caseId: cid, lastMessageAt: lastMessageAt})",
+    )
+    fun findLastMessageTimestamps(
```
selimbensenouciep-prog, line 49: `findLastMessageTimestamps` has no integration coverage. In tests it is only reached through the MockK stub in `CaseControllerSpec` (which mocks `CaseEventService`, not the repo) and `InMemoryCaseEventRepository`, and both hand back a `java.time.Instant` without ever touching the Neo4j driver mapping, so the raw-map decode in `Neo4jCaseEventRepository` (`row["lastMessageAt"] as Instant`) never runs under test. `AbstractCaseEventPersistenceSpec` already covers every other `CaseEventRepository` method and runs against embedded Neo4j (no Docker) via `EmbeddedNeo4jCaseEventPersistenceSpec`. Adding one case to that contract (save a `MessageEvent`, assert `findLastMessageTimestamps` returns its timestamp) would run the real mapping and lock the behaviour, the way the sibling raw-map query `findDirectRelations` is covered in `EmbeddedNeo4jPermissionStarPersistenceSpec`. Could we add it?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
thomasmartin-whoz: reverted

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
thomasmartin-whoz: done

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/Neo4jCaseEventRepository.kt
thomasmartin-whoz: Cast fix (Neo4jCaseEventRepository): row["lastMessageAt"] as Instant → (row["lastMessageAt"] as ZonedDateTime).toInstant(), plus the early-return guard for empty input.

KDoc fix (CaseEventNodeNeo4jRepository): corrected the stated return type from Instant to ZonedDateTime, with an explicit note that the caller must convert.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventNodeNeo4jRepository.kt
thomasmartin-whoz: Contract tests (AbstractCaseEventPersistenceSpec): two new cases run against the real embedded Neo4j engine — one verifying that the max timestamp is returned per case and that message-less cases are absent, one verifying the empty-input short-circuit. These inherit into EmbeddedNeo4jCaseEventPersistenceSpec automatically, the same pattern Selim pointed to with findDirectRelations in the permission spec.

AlanMenantWhoz (approved):

PR #1163, opened 2026-07-23
fix: wz-33587: fix last message mapping on

thomasmartin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
```diff
@@ -353,6 +358,17 @@ class CaseController(
...
     }
 
+    /**
+     * Enrich a list of [Case]s with [CaseDto.lastMessageAt] only — no role/favorite.
+     *
+     * Used by the "list by target user" endpoints where role and favorite are caller-relative
+     * and would be misleading, but [CaseDto.lastMessageAt] is objective and always useful.
+     */
+    private fun List<Case>.withLastMessageAt(): List<CaseDto> {
+        val lastMessageTimestamps = caseEventService.findLastMessageTimestamps(map { it.id })
```
fdelsert-whoz, line 368: Not a fan of the extension with side effects, but given how local it is, it's fine.
I'd have preferred a private function taking the list as parameter.

fdelsert-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
thomasmartin-whoz: what side effect ? The parameter is not modified

fdelsert-whoz: findLastMessageTimestamps is a DB call.

caseEventService.findLastMessageTimestamps is a side effect in the broad sense, it makes a round trip to the DB, can fail, and can return different results depending on the DB state.
Variable immutability is not enough. That's why an extension on List doing this is surprising, you don't expect iterating/transforming a list to trigger a query.

But the comment is clear and honest about what the method does. It's probably just the name withLastMessageAt that gives the impression of a simple data transformation, whereas a name like fetchAndEnrichWithLastMessageAt or loadLastMessageAt would better signal the I/O.

PR #1171, opened 2026-07-28
feat: make exchange agent tools configurable by env

selimbensenouciep-prog:
## Why

The two built-in exchange integrations (`CASE_FILE_EXCHANGE`, `NAMESPACE_FILE_EXCHANGE`) are declared `configSchema = null`: no `IntegrationConfig` is ever persisted for them, and their tools are built imperatively in `AgentServiceImpl.buildExchangeTools`. Two gaps followed from that.

**Nine of twelve plugin settings were unreachable.** `FileToolProvider.provideTools` reads 12 keys; `buildExchangeTools` built a node with 3 (`rootPath`, `readOnly`, `readMaxSizeMb`). `extraDenyPatterns` and the eight image/document knobs stayed pinned to the plugin's compiled constants, with no operational lever of any kind.

**There was no instance-level enablement.** The exchange could only be turned on agent by agent, through the `integrations` map. An operator could not decide that every agent on a deployment should get it.

## What

New `ExchangeToolsConfigProperties` (prefix `agentos.exchange.tools`) and `ExchangeToolGrantService`, which owns the grant decision and turns the bound properties into the `provideTools` node. The SDK and the plugins keep no Spring dependency: the values still reach the plugin through the existing `JsonNode` contract, so nothing under `agentos-sdk/` or `agentos-file-plugin/` is touched.

Eleven new environment variables, all off or at the plugin's own value by default:

| Property | Env var | Default |
|---|---|---|
| `…tools.case-enabled-by-default` | `AGENTOS_EXCHANGE_TOOLS_CASE_ENABLED_BY_DEFAULT` | `false` |
| `…tools.namespace-enabled-by-default` | `AGENTOS_EXCHANGE_TOOLS_NAMESPACE_ENABLED_BY_DEFAULT` | `false` |
| `…tools.extra-deny-patterns` | `AGENTOS_EXCHANGE_TOOLS_EXTRA_DENY_PATTERNS` | empty |
| `…tools.image-max-dimension` | `AGENTOS_EXCHANGE_TOOLS_IMAGE_MAX_DIMENSION` | `1024` |
| `…tools.image-jpeg-quality` | `AGENTOS_EXCHANGE_TOOLS_IMAGE_JPEG_QUALITY` | `0.80` |
| `…tools.image-max-source-pixels` | `AGENTOS_EXCHANGE_TOOLS_IMAGE_MAX_SOURCE_PIXELS` | `50000000` |
| `…tools.image-pass-through-max-bytes` | `AGENTOS_EXCHANGE_TOOLS_IMAGE_PASS_THROUGH_MAX_BYTES` | `1048576` |
| `…tools.document-max-output-chars` | `AGENTOS_EXCHANGE_TOOLS_DOCUMENT_MAX_OUTPUT_CHARS` | `100000` |
| `…tools.document-max-attached-images` | `AGENTOS_EXCHANGE_TOOLS_DOCUMENT_MAX_ATTACHED_IMAGES` | `10` |
| `…tools.document-max-table-columns` | `AGENTOS_EXCHANGE_TOOLS_DOCUMENT_MAX_TABLE_COLUMNS` | `64` |
| `…tools.document-max-cell-chars` | `AGENTOS_EXCHANGE_TOOLS_DOCUMENT_MAX_CELL_CHARS` | `5000` |

One set of values is shared by both scopes. `rootPath` and `readOnly` stay computed per run, and `readMaxSizeMb` keeps deriving from `agentos.exchange.read-max-size-bytes` so the agent read cap cannot drift from the one the REST path enforces.

## Enablement is now a tri-state

Carried by `AgentConfig.integrations`, no new field:

| `integrations` entry | Result |
|---|---|
| key absent | the platform default decides |
| `CASE_FILE_EXCHANGE:` (null) | granted, every tool |
| `CASE_FILE_EXCHANGE: [readFile, ls]` | granted, restricted to those names |
| `CASE_FILE_EXCHANGE: []` | explicit opt-out |

The agent's own choice always wins, in both directions; the default only fills the gap. The empty list is a genuine opt-out rather than an empty allow-list: `isToolAllowed` would already reject every tool, but the grant creates the scope directory *before* that filter runs, so the decision short-circuits in `resolveGrant` and no phantom directory is materialised.

The default is a decision local to `ExchangeToolGrantService` and never a mutation of `AgentConfig.integrations` — writing the key into the map would leak it into the persisted config, the YAML export, and the peer descriptions `RedirectTool` puts in every agent's prompt.

## Why the frontend had to move

`IntegrationTypeDescriptor` gains `enabledByDefault`, and the agent form renders three states (*Platform default* / *Enabled* / *Disabled*) instead of a checkbox.

This is not cosmetic. The old form hydrated `enabled` from `type in integrations`, so an explicit opt-out (`[]`) displayed as **checked**; and an unchecked row was simply omitted from the map, which is indistinguishable from "never configured". On a default-on instance, an agent nobody had ever configured would therefore have been silently switched off the first time someone saved an unrelated change. A test now guards that path.

## Also in this PR

- `buildExchangeTools` drops from 74 to 49 lines, keeps only what it alone knows (the run's `caseId`, the invoking user's Namespace `WRITE`), and loses its early return.
- `ExchangeStorageService.readMaxSizeBytes` is removed — `buildExchangeTools` was its only caller.
- `agentos.exchange.read-max-size-bytes` is documented for the first time. It has existed since the exchange shipped but appeared in neither `application.yml` nor `docs/file-exchange.md`.
- `docs/file-exchange.md`: the claim that all three `agentos.exchange` keys use `${ENV_VAR:default}` placeholders was false — they rely on relaxed binding alone.

## Verification

- `agentos-service` 1957 tests, `agentos-sdk` 8, `agentos-ui` 214 — all green
- `nx lint agentos-ui` clean
- OpenAPI spec and TS client regenerated; the only delta is `enabledByDefault`

Behaviour with no environment variable set is unchanged: the nine tuning defaults replicate the plugin constants exactly, and both enablement defaults ship off.

## Known limitations, deliberately out of scope

- **A default-on instance materialises a case exchange directory per case a granted agent runs in.** The file plugin canonicalises `rootPath` at construction, so a grant must create its directory. Deferring that would mean touching `BoundaryPathResolver`. Documented under §Safety.
- **`grantTools` has no try/catch.** On a read-only filesystem or a full quota, the `IOException` escapes `findAgentByName`; with a default of on, that would break every agent instead of only those that opted in. Adding a catch is a behaviour change that deserves its own ticket.
- The generated TS client is out of sync with the committed spec on master for unrelated models (`Case.lastMessageAt`, `ToolResponseEvent.images`, `ExchangeControllerService` doc comments). I reverted those hunks to keep this diff scoped; they need a dedicated regeneration commit.

---

## Follow-up: review findings addressed (commit 2)

An adversarial review pass over this diff surfaced four defects, three of them in behaviour this PR introduced. All are fixed in the second commit.

**Namespace READ is now required on the agent grant path.** The grant derived only `readOnly` from Namespace `WRITE` and never checked `READ`, while every REST namespace-file endpoint enforces it via `@PreAuthorize`. A user holding a direct Case `ADMIN` edge without namespace membership could read the whole shared exchange through any agent, files their own REST calls deny. The gap predates this PR, but the new platform default scales it from individually configured agents to every agent behind one flag. `ExchangeCapabilityService` gains `canRead`, and the namespace grant requires it.

> **Behaviour change worth flagging:** a run without an identified user is now denied fail-closed, so `GET /api/agent-configs/{id}/definition` called with `withUserOverlay=false` no longer reports namespace exchange tools.

**The `[]` opt-out did not survive the YAML export.** The export mapper's `NON_EMPTY` inclusion also applies to map *content*, so it pruned both the `[]` opt-out and the `null` "all tools" value out of the `integrations` map. Re-importing such a file re-granted the exchange to an agent an admin had explicitly opted out, which defeats the tri-state this PR is built around. `toExportModel` already filters top-level values itself, so the mapper-level policy is dropped.

**The agent form widened a per-tool allowlist to all tools.** A non-empty array hydrated to the `on` state and was written back as `null`, so an agent restricted to `readFile` silently gained `editFiles`, `remove` and `moveFile`. The allowlist is now carried on the row, surfaced read-only, and re-emitted verbatim.

**The AgentServiceImpl to grant-service seam was unverified.** Both resolve calls were stubbed with `any()` and never verified, so a regression passing `null` instead of `config.integrations` would have kept every test green while an opted-out agent regained its tools. Pinned by exact-argument verifications plus a test composing the two services for real.

Smaller fixes: the FILE_ACCESS presence check moves up into `resolveGrant` so a plugin-less instance pays no permission query; `documentMaxTableColumns`, `documentMaxOutputChars` and `documentMaxCellChars` are clamped like `imageJpegQuality` already was, each crashing or livelocking a tool call when set out of range; `extraDenyPatterns` is documented as matching the final path segment only, so operators do not write path-shaped patterns that silently match nothing; and two stale KDoc claims are corrected (`ToolResolverService` on tool-free agents, `AgentConfig.integrations` on null semantics).

Suites after the fixes: `agentos-service` 1970, `agentos-sdk` 8, `agentos-ui` 222, all green.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
@@ -15,13 +15,14 @@ import java.util.UUID
  * by an AiProvider — resolution is deferred to the runtime layer.
  *
  * [integrations] is an optional map from integration name to an optional list of
- * allowed tool names. A null map binds no integration at all: the tool resolver
- * grants no tools, and only the platform-level exchange defaults (see the property
- * KDoc below) can still add the built-in file tools. When set, only tools whose
- * integration key matches an entry in this map are given to the agent. A null list
- * for a given key means all tools from that integration are allowed; a non-null
- * list restricts to exactly those tool names (or suffixes for multi-instance tools
- * named `CONFIG__tool`).
+ * allowed tool names. The map is what the agent declares, key by key: a plugin
+ * integration contributes tools only when its name appears as an entry, so a null
+ * map binds none of them at all. The two built-in exchange keys are the exception,
+ * and they are decided per key rather than per map: an absent key defers to the
+ * platform default (see the property KDoc below), whether the rest of the map is
+ * set or not. A null list for a given key means all tools from that integration are
+ * allowed; a non-null list restricts to exactly those tool names (or suffixes for
+ * multi-instance tools named `CONFIG__tool`).
  *
  * Examples (from a Coday-style agent YAML):
  * ```yaml
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
@@ -55,7 +55,11 @@ data class AgentConfig(
      * Reserved keys `CASE_FILE_EXCHANGE` / `NAMESPACE_FILE_EXCHANGE`
      * (see [io.whozoss.agentos.exchange.ExchangeIntegrationTypes]) enable the built-in
      * file-exchange integrations: they have no [IntegrationConfig] instance and are resolved by
-     * `AgentServiceImpl.buildExchangeTools` rather than the normal plugin path.
+     * `AgentServiceImpl.buildExchangeTools` rather than the normal plugin path. For those two keys
+     * only, an **empty list is an explicit opt-out** rather than an empty allow-list, and an
+     * **absent key does not mean "never granted"**: the platform defaults
+     * `agentos.exchange.tools.case-enabled-by-default` /
+     * `...namespace-enabled-by-default` (off by default) decide for an agent that stays silent.
      */
     val integrations: Map<String, List<String>?>? = null,
     /**
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
@@ -15,11 +15,13 @@ import java.util.UUID
  * by an AiProvider — resolution is deferred to the runtime layer.
  *
  * [integrations] is an optional map from integration name to an optional list of
- * allowed tool names. When null, the agent receives all tools available in the
- * namespace (no filtering). When set, only tools whose integration key matches an
- * entry in this map are given to the agent. A null list for a given key means all
- * tools from that integration are allowed; a non-null list restricts to exactly
- * those tool names (or suffixes for multi-instance tools named `CONFIG__tool`).
+ * allowed tool names. A null map binds no integration at all: the tool resolver
+ * grants no tools, and only the platform-level exchange defaults (see the property
+ * KDoc below) can still add the built-in file tools. When set, only tools whose
+ * integration key matches an entry in this map are given to the agent. A null list
+ * for a given key means all tools from that integration are allowed; a non-null
+ * list restricts to exactly those tool names (or suffixes for multi-instance tools
+ * named `CONFIG__tool`).
  *
  * Examples (from a Coday-style agent YAML):
  * ```yaml
@@ -47,7 +49,10 @@ data class AgentConfig(
     val instructions: String? = null,
     val modelName: String? = null,
     /**
-     * Optional tool-access filter. null = no restriction (all namespace tools).
+     * Integration bindings with an optional per-tool allowlist. null = no bindings:
+     * [io.whozoss.agentos.tool.ToolResolverService] then resolves no tools for this agent
+     * (only the platform exchange defaults described below can still grant the built-in
+     * file tools).
      * Map key = integration name (matches [IntegrationConfig.name] or
      * [ToolPlugin.integrationType] for config-less plugins).
      * Map value = allowed tool names, or null for all tools of that integration.
@@ -60,6 +65,8 @@ data class AgentConfig(
      * **absent key does not mean "never granted"**: the platform defaults
      * `agentos.exchange.tools.case-enabled-by-default` /
      * `...namespace-enabled-by-default` (off by default) decide for an agent that stays silent.
+     * The namespace key is further gated at run time: whatever this map says, the scope is only
+     * granted when the invoking user holds Namespace READ.
      */
     val integrations: Map<String, List<String>?>? = null,
     /**
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
@@ -1,6 +1,5 @@
 package io.whozoss.agentos.agentConfig
 
-import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.SerializationFeature
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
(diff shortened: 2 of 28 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
```diff
@@ -47,15 +50,24 @@ data class AgentConfig(
...
      * Map value = allowed tool names, or null for all tools of that integration.
      *
      * Reserved keys `CASE_FILE_EXCHANGE` / `NAMESPACE_FILE_EXCHANGE`
      * (see [io.whozoss.agentos.exchange.ExchangeIntegrationTypes]) enable the built-in
      * file-exchange integrations: they have no [IntegrationConfig] instance and are resolved by
-     * `AgentServiceImpl.buildExchangeTools` rather than the normal plugin path.
+     * `AgentServiceImpl.buildExchangeTools` rather than the normal plugin path. For those two keys
+     * only, an **empty list is an explicit opt-out** rather than an empty allow-list, and an
+     * **absent key does not mean "never granted"**: the platform defaults
+     * `agentos.exchange.tools.case-enabled-by-default` /
```
thomasmartin-whoz, line 67: it feels the semantic between empty list and null is rather obscure and implicit, I would rather have a less confusing pattern.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
```diff
@@ -262,7 +261,13 @@ class AgentConfigController(
          * YAML mapper configured for clean, human-readable output:
          * - No `---` document start marker
          * - No Jackson type tags
-         * - Null and empty values omitted (via [toExportModel] filtering + NON_EMPTY inclusion)
+         * - Null, blank and empty top-level values omitted by [toExportModel] itself
+         *
+         * Deliberately no mapper-level inclusion policy: `setSerializationInclusion` also sets the
+         * CONTENT inclusion, which prunes entries inside the `integrations` map. There a null value
+         * ("all tools of the integration") and an empty list (explicit opt-out) are distinct,
```
thomasmartin-whoz, line 268: again this distinction between null and empty list is touchy

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/exchange/ExchangeToolGrantService.kt
```diff
@@ -0,0 +1,186 @@
...
+        enabledByDefault: Boolean,
+        integrations: Map<String, List<String>?>?,
+    ): ExchangeGrant? {
+        val declared = integrations?.containsKey(integrationKey) == true
+        val declaration = integrations?.get(integrationKey)
+        return when {
+            toolRegistryService.findPlugin(ExchangeIntegrationTypes.FILE_ACCESS) == null -> null
+            !declared && enabledByDefault -> ExchangeGrant(allowedTools = null)
+            !declared -> null
+            declaration == null -> ExchangeGrant(allowedTools = null)
```
thomasmartin-whoz, line 132: what is the difference between 
!declared ->
and
 declaration == null -> 
?
How can declared be true while declaration is null ?

is it something like `integrationKey=null` in the configuration ?
I do not think that having a completely different semantic between having null and not having the key is a good idea.

Maybe we should have a special wildcard or something to explicit intent in the configuration.

vincent-audibert-whoz: This comes from the coday-era implicit null = all tools convention in declaring agent integration tools:

```
integrations:
  FOO:                               // no value = null, taken as "all tools" that is the problematic part
  BAR: ["tool1", "tool2"]  // explicit values
  BAZ: [].                          // specific case, not null but empty list = no tool
```

We should indeed move to a more explicit tool list declaration but that is a broader topic disconnected from that very PR.

thomasmartin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/exchange/ExchangeToolGrantService.kt
thomasmartin-whoz: logged in Jira for the future
https://whoz.atlassian.net/browse/WZ-33771

vincent-audibert-whoz (approved):

PR #1197, opened 2026-08-03
feat: agentos auto download approach

vincent-audibert-whoz:


.github/workflows/release.yml
```diff
@@ -109,7 +109,48 @@ jobs:
           echo "- **Run ID:** ${{ github.run_id }}" >> $GITHUB_STEP_SUMMARY
           echo "- **Run time:** $(date -u '+%Y-%m-%d %H:%M:%S UTC')" >> $GITHUB_STEP_SUMMARY
 
-  # Job 2: Build and upload desktop apps (Coday + Coday Twin)
+  # Job 2: Build and upload AgentOS JARs as GitHub Release assets
+  upload-agentos-jars:
```
vincent-audibert-whoz, line 113: weird, already exists elsewhere or already covered as jars are available on the github `/packages` page ?

vincent-audibert-whoz: Distinction: here we add the jar to the publicly available release assets. Different from the github maven that requires auth.

apps/server/postinstall.js
```diff
@@ -0,0 +1,119 @@
...
+ * The JARs are too large to bundle in the npm tarball (they are already-compressed
+ * ZIP archives and do not benefit from gzip). Instead, they are uploaded as assets
+ * on the GitHub Release that corresponds to the version of this package, and this
+ * script fetches them on first install.
+ *
+ * The script is intentionally dependency-free (only Node built-ins) so it works
+ * in any environment without a prior `npm install`.
+ *
+ * Environment variables:
+ *   AGENTOS_HOSTNAME        if set, an external AgentOS instance is configured — skip download
```
vincent-audibert-whoz, line 14:
```js
if (process.env.AGENTOS_HOSTNAME || process.env.AGENTOS_PORT) {
```
The skip condition checks `AGENTOS_HOSTNAME` / `AGENTOS_PORT`, but the warning message below references `CODAY_SKIP_AGENTOS_DOWNLOAD=1` as the way to suppress the warning. That env var is never actually read anywhere in this file. Either add a check for it (e.g. `|| process.env.CODAY_SKIP_AGENTOS_DOWNLOAD`) or remove it from the warning message to avoid misleading operators.

apps/server/postinstall.js
```diff
@@ -0,0 +1,119 @@
...
+ *   AGENTOS_PORT            if set, an external AgentOS instance is configured — skip download
+ *   CODAY_AGENTOS_VERSION   override the version used to find the GitHub Release (e.g. for testing)
+ */
+import { createWriteStream, existsSync, mkdirSync } from 'fs'
+import { pipeline } from 'stream/promises'
+import { dirname, resolve } from 'path'
```
vincent-audibert-whoz, line 21:
```js
import { createReadStream } from 'fs'
import { createGunzip } from 'zlib'
```
Both `createReadStream` and `createGunzip` are imported but never used. Dead imports — remove them.

apps/server/postinstall.js
```diff
@@ -0,0 +1,119 @@
...
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+
+const REPO = 'whoz-oss/coday'
+
+// JARs expected relative to this script's directory after download
+const JARS = [
+  { asset: 'agentos-service.jar', dest: 'agentos/agentos-service.jar' },
+  { asset: 'agentos-bash-plugin.jar', dest: 'agentos/plugins/agentos-bash-plugin.jar' },
+  { asset: 'agentos-file-plugin.jar', dest: 'agentos/plugins/agentos-file-plugin.jar' },
```
vincent-audibert-whoz:
```js
{ asset: 'agentos-bash-plugin.jar', dest: 'agentos/plugins/agentos-bash-plugin.jar' },
{ asset: 'agentos-file-plugin.jar', dest: 'agentos/plugins/agentos-file-plugin.jar' },
{ asset: 'agentos-mcp-plugin.jar',  dest: 'agentos/plugins/agentos-mcp-plugin.jar' },
{ asset: 'agentos-tmux-plugin.jar', dest: 'agentos/plugins/agentos-tmux-plugin.jar' },
```
The CI upload step uses a glob `agentos-bash-plugin-*.jar`, so the actual asset name on the release will include the version, e.g. `agentos-bash-plugin-1.2.3.jar`. The `assets.find((a) => a.name === asset)` lookup performs an exact match, so it will never find these four JARs. Only `agentos-service.jar` (produced by `bootJar` with a fixed output name) will download correctly. The lookup logic needs to handle versioned names, e.g. by using `a.name.startsWith('agentos-bash-plugin')` or by resolving the exact asset name from the release metadata.

apps/server/project.json
```diff
@@ -82,7 +82,7 @@
       }
     },
     "nx-release-publish": {
-      "dependsOn": ["build", "build-agentos"],
+      "dependsOn": ["build"],
```
vincent-audibert-whoz, line 86:
```json
"nx-release-publish": {
  "dependsOn": ["build"]
```
`build-agentos` was removed from `dependsOn` here. Previously, publishing the npm package guaranteed the JARs were built and included in `dist/agentos/`. Now `nx-release-publish` can succeed with an empty `dist/agentos/` directory. If the intent is that JARs are no longer bundled in the npm tarball at all (and only downloaded at postinstall time), the `build-agentos` target and the `files` entry in `package.json` may also need revisiting. If the JAR bundling is still expected for some environments, the dependency should be restored. Either way, the intent should be explicit.

vincent-audibert-whoz (commented):
The overall design — publish JARs as public GitHub Release assets, download on `postinstall` — is sound and the right direction for distributing large binaries outside the npm tarball. The implementation is clean and intentionally fail-safe. A few correctness and robustness issues need addressing before merge.

**Correctness issue in `postinstall.js`**: the skip-check uses `AGENTOS_HOSTNAME || AGENTOS_PORT` but the warning message mentions `CODAY_SKIP_AGENTOS_DOWNLOAD=1` — that env var is referenced in the message but never actually honoured. Either honour it or remove the reference.

**Asset-name mismatch for plugins**: the CI uploads plugin JARs with a glob (`agentos-bash-plugin-*.jar`) so the actual release asset name will include the version, e.g. `agentos-bash-plugin-1.2.3.jar`. The download script looks for exact names like `agentos-bash-plugin.jar` — this lookup will always fail for plugins. The service JAR (`agentos-service.jar`) is fine because `bootJar` produces a fixed name, but all four plugin JARs will be missed.

**Unused imports in `postinstall.js`**: `createReadStream` and `createGunzip` are imported but never used. These are dead code.

**`nx-release-publish` no longer depends on `build-agentos`**: the change to `project.json` removes `build-agentos` from the `nx-release-publish` dependency chain. This means publishing the npm package no longer guarantees the JARs were built and copied into `dist/agentos/`. If someone runs `nx-release-publish` locally or the CI order changes, the dist will be published without the JARs. The `build-agentos` target still exists and is still used for the `upload-agentos-jars` CI job, but the decoupling from publish is a silent regression — the intent should be made explicit (either document why it's intentional, or keep the dependency).

apps/server/postinstall.js
```diff
@@ -0,0 +1,122 @@
...
+  return res.json()
+}
+
+async function downloadFile(url, destPath) {
+  mkdirSync(dirname(destPath), { recursive: true })
+  const res = await fetch(url, { headers: { Accept: 'application/octet-stream' }, redirect: 'follow' })
+  if (!res.ok) throw new Error(`Download ${url} → ${res.status} ${res.statusText}`)
+  await pipeline(res.body, createWriteStream(destPath))
+}
+
```
vincent-audibert-whoz, line 65: **Critical — no integrity verification before executing downloaded JARs.**

The file is written directly to disk and will subsequently be executed as a JVM application. There is no checksum (SHA-256/SHA-512) or signature validation at any point. If the GitHub Release asset is replaced (compromised RELEASE_PAT, race between upload and download, GitHub-side incident), arbitrary code runs on every user machine at `npm install` time.

The CI upload step should compute and publish a checksum manifest (e.g. `checksums.sha256`) alongside the JARs. This script should download that manifest first, verify each JAR against it before writing to disk, and abort if verification fails.

Security risk: supply-chain code execution — a tampered JAR is indistinguishable from a legitimate one without integrity verification.

apps/server/postinstall.js
```diff
@@ -0,0 +1,122 @@
...
+async function main() {
+  if (process.env.AGENTOS_HOSTNAME || process.env.AGENTOS_PORT) {
+    console.log('[coday-server] External AgentOS configured — skipping JAR download')
+    return
+  }
+
+  // Read version from package.json sitting next to this script (or use override)
+  const { createRequire } = await import('module')
+  const require = createRequire(import.meta.url)
```
vincent-audibert-whoz, line 115: **Moderate — `CODAY_AGENTOS_VERSION` accepted without validation.**

```js
const version = process.env.CODAY_AGENTOS_VERSION ?? pkgVersion
const tag = `release/${version}`
```

This value is used verbatim in the GitHub API URL: `https://api.github.com/repos/${REPO}/releases/tags/${tag}`. There is no format validation (e.g. semver regex). An attacker who can inject environment variables during install can redirect the download to any release tag they choose, including a tag they control if the repo is ever forked or the naming convention is predictable.

Security risk: combined with the absent integrity check, this allows targeted delivery of a malicious JAR version to specific install environments (e.g. CI pipelines with env var injection). At minimum, validate the version string against a semver pattern before using it.

apps/server/postinstall.js
```diff
@@ -0,0 +1,122 @@
...
+async function fetchJson(url) {
+  const res = await fetch(url, {
+    headers: { Accept: 'application/vnd.github+json', 'X-GitHub-Api-Version': '2022-11-28' },
+  })
+  if (!res.ok) throw new Error(`GET ${url} → ${res.status} ${res.statusText}`)
+  return res.json()
+}
+
+async function downloadFile(url, destPath) {
+  mkdirSync(dirname(destPath), { recursive: true })
```
vincent-audibert-whoz, line 60: **Minor — `browser_download_url` followed without domain validation.**

```js
const res = await fetch(url, { headers: { Accept: 'application/octet-stream' }, redirect: 'follow' })
```

The `url` parameter comes from `found.browser_download_url` which is taken verbatim from the GitHub API JSON response. The fetch is performed with `redirect: 'follow'`, meaning any redirect chain is followed silently. The code does not assert that the resolved URL remains within `github.com` or `objects.githubusercontent.com` before streaming bytes to disk.

Security risk: if the GitHub API were ever manipulated to return a redirect to an attacker-controlled host (MITM, DNS spoofing, or a future API behaviour change), the download would proceed without warning. Assert the URL origin before fetching.

.github/workflows/release.yml
```diff
@@ -109,7 +109,48 @@ jobs:
...
+
+      - name: Build AgentOS JARs
+        working-directory: agentos
+        run: |
+          ./gradlew :agentos-service:bootJar :agentos-bash-plugin:jar :agentos-file-plugin:jar :agentos-mcp-plugin:jar :agentos-tmux-plugin:jar --no-daemon
+
+      - name: Upload JARs to GitHub Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_PAT }}
+        run: |
```
vincent-audibert-whoz, line 143: **Moderate — `RELEASE_PAT` used where `secrets.GITHUB_TOKEN` is sufficient.**

```yaml
env:
  GITHUB_TOKEN: ${{ secrets.RELEASE_PAT }}
```

The `upload-agentos-jars` job already declares `permissions: contents: write`. The built-in `GITHUB_TOKEN` with that permission is sufficient to run `gh release upload` on a release in the same repository. Using a long-lived PAT (`RELEASE_PAT`) here unnecessarily widens the blast radius: PATs typically carry permissions across multiple repositories and do not expire with the workflow run.

Security risk: unnecessary privilege escalation — replace with `secrets.GITHUB_TOKEN` (or `github.token`) to follow the principle of least privilege. The existing `release` and `desktop-release` jobs already use `RELEASE_PAT` for operations that genuinely require it (cross-repo push, npm publish); this job does not.

vincent-audibert-whoz (commented):
## Security Review

This PR introduces a postinstall script that downloads and executes JARs from GitHub Releases at `npm install` time. The overall supply-chain design has two significant security issues that should be addressed before merge.

**Critical — no integrity verification on downloaded JARs.** The postinstall script downloads executable JARs from `browser_download_url` with no checksum or signature validation. Anyone who can tamper with the GitHub Release assets (compromised PAT, TOCTOU between upload and download, or a GitHub-side compromise) can deliver arbitrary code to every user who runs `npm install`. This is a classic supply-chain attack vector for postinstall scripts.

**Moderate — `RELEASE_PAT` used where `GITHUB_TOKEN` is sufficient.** The `upload-agentos-jars` job uses `secrets.RELEASE_PAT` (a long-lived personal access token) to authenticate the `gh release upload` command. The built-in `secrets.GITHUB_TOKEN` scoped to `contents: write` is already declared on this job and is sufficient for uploading assets to a release in the same repository. Using a PAT here unnecessarily widens the blast radius if the token is ever leaked — PATs typically carry broader permissions across repositories.

**Moderate — `CODAY_AGENTOS_VERSION` env var allows arbitrary version injection.** The postinstall script accepts `CODAY_AGENTOS_VERSION` as an override for the release tag, and uses it directly to construct the GitHub API URL with no validation. An attacker who can inject environment variables during install (e.g. via a compromised parent package's install script, or a misconfigured CI environment) can redirect the download to any release tag. This is especially dangerous in the absence of integrity verification.

**Minor — `browser_download_url` from untrusted API response used verbatim.** The download URL is taken directly from the GitHub API JSON response (`found.browser_download_url`) and followed with `redirect: 'follow'`. The code trusts GitHub's API to return a safe URL, but does not validate that the URL remains within the expected domain before following redirects. A URL pointing to an attacker-controlled host would be followed silently.

apps/server/postinstall.js
```diff
@@ -0,0 +1,192 @@
...
+
+function sha256File(filePath) {
+  const hash = createHash('sha256')
+  hash.update(readFileSync(filePath))
+  return hash.digest('hex')
+}
+
+async function fetchChecksums(assets, tag) {
+  const checksumAsset = assets.find((a) => a.name === 'checksums.sha256')
+  if (!checksumAsset) {
```
pierre-fortes-whoz, line 82: It means in the frame of a supply chain attack if the jar is replaced with no checksum, the check will be skipped, has this been added for former versions of the jar i suppose ? 

I think it'd be better to remove that if statement, or eventually if you need this, to allow check skipping in a "developer mode" or something like that (maybe to complex though)

apps/server/postinstall.js
```diff
@@ -0,0 +1,192 @@
...
+      failed = true
+      continue
+    }
+
+    process.stdout.write(`[coday-server]   ↓ ${found.name}...`)
+    try {
+      await downloadFile(found.browser_download_url, destPath)
+
+      // Verify checksum if manifest is available
+      if (checksums) {
```
pierre-fortes-whoz: Similarly to my other comment, if checksums are removed in my opinion, this should fail and not skip the checks

pierre-fortes-whoz (changes requested):

apps/server/postinstall.js
vincent-audibert-whoz: ok, checksums mandatory past a certain version

apps/server/postinstall.js
vincent-audibert-whoz: ok

pierre-fortes-whoz (approved):

pierre-fortes-whoz (approved):

vincent-palita-whoz (approved):

PR #1212, opened 2026-08-04
fix: #1211 replace AGENTOS_HOSTNAME with validated AGENTOS_URL

### LINKED ISSUE #1211: fix: AGENTOS_HOSTNAME silently produces malformed proxy target
## Symptom

All requests to `/api/agentos/*` fail with **504 Gateway Timeout**, while the Express proxy log shows the request being received:

```
[AGENTOS PROXY] GET /api/users/me
```

The AgentOS Spring backend is healthy and listening on port 8124, but never receives the request.

## Root cause

Startup log reveals a malformed proxy target with a doubled scheme:

```
AgentOS proxy configured: /api/agentos → http://http://localhost:8124
```

In `apps/server/src/server.ts`:

```ts
const AGENTOS_HOSTNAME = process.env.AGENTOS_HOSTNAME ?? 'localhost'
const AGENTOS_URL = `http://${AGENTOS_HOSTNAME}:${AGENTOS_PORT}`
```

The template literal prepends `http://`, so the env var is contractually expected to hold a bare hostname. When it contains a full URL (e.g. `http://localhost`), the result is `http://http://localhost:8124`. `http-proxy-middleware` then resolves a literal host named `http:`, which never connects — hence the timeout rather than a fast, legible failure.

The variable was introduced in commit `6f3ddd6a` (`feat: add possibility to provide agentos hostname (#847)`) as a one-liner. The "bare hostname, no scheme" contract is undocumented and unvalidated.

## Why this is worth fixing structurally

A bare-hostname abstraction is inexpressive: it cannot carry a scheme (no `https` for a remote AgentOS), nor a base path, nor an implicit port. Hardening the existing variable by stripping schemes would perpetuate a leaky abstraction and set up the same incident again the day TLS is needed.

Additionally, the default AgentOS port is inconsistent across the codebase: `server.ts` defaults to `8124`, while `scripts/web-dev-tmux.sh` scans upward from `8123`. Currently harmless, but two independent sources of truth for the same value.

## Proposed fix

Replace `AGENTOS_HOSTNAME` with a single `AGENTOS_URL` env var holding a complete URL, validated fail-fast at boot:

```ts
const AGENTOS_URL = process.env.AGENTOS_URL ?? `http://localhost:${AGENTOS_PORT}`
try {
  new URL(AGENTOS_URL)
} catch {
  throw new Error(`Invalid AGENTOS_URL: "${AGENTOS_URL}"`)
}
```

`AGENTOS_PORT` is retained as a fallback for local dev orchestrated by `web-dev-tmux.sh`.

Benefits: the value is self-describing, supports scheme/host/port/path in one place, and a misconfiguration surfaces as an immediate startup error rather than a 504 discovered much later.

**Breaking change**: `AGENTOS_HOSTNAME` is removed. Any deployment setting it must migrate to `AGENTOS_URL` with the full URL including scheme.

vincent-audibert-whoz:
Closes #1211

## Problem

All `/api/agentos/*` requests were failing with **504 Gateway Timeout** while the Express proxy log showed the request arriving normally. The startup log held the answer:

```
AgentOS proxy configured: /api/agentos → http://http://localhost:8124
```

`server.ts` built the proxy target by prepending a scheme to a bare-hostname env var:

```ts
const AGENTOS_HOSTNAME = process.env.AGENTOS_HOSTNAME ?? 'localhost'
const AGENTOS_URL = `http://${AGENTOS_HOSTNAME}:${AGENTOS_PORT}`
```

When `AGENTOS_HOSTNAME` already contained a scheme, the result was a doubled one. `http-proxy-middleware` then tried to resolve a literal host named `http:`, which never connects — producing a slow timeout instead of a fast, legible error. The "bare hostname, no scheme" contract was undocumented and unvalidated.

## Approach

Rather than hardening the bare-hostname variable by stripping schemes, this replaces it with a single `AGENTOS_URL` holding a **complete URL**. A bare hostname is an inexpressive abstraction: it cannot carry a scheme (no `https` for a remote AgentOS), nor a base path. Papering over the symptom would have set up the same incident again the day TLS is needed.

```ts
const AGENTOS_URL = process.env.AGENTOS_URL ?? `http://localhost:${AGENTOS_PORT}`
validateAgentOsUrl(AGENTOS_URL)
```

Validation runs at module load, so a misconfiguration is now a startup failure with an explicit message naming the offending value — not a 504 discovered much later.

`AGENTOS_PORT` is retained as the fallback, since `scripts/web-dev-tmux.sh` allocates ports dynamically to let multiple git worktrees run concurrently.

## Validation is deliberately stricter than `new URL()`

`new URL()` alone is too permissive to catch the mistake this change most invites. Anyone migrating by copying their old value across (`AGENTOS_HOSTNAME=my-host` → `AGENTOS_URL=my-host`) would otherwise sail straight through:

- `new URL('localhost:8124')` does **not** throw — it reads `localhost:` as the scheme and `8124` as the path.
- `new URL('http://http://localhost:8124')` does **not** throw — it parses as hostname `http` with pathname `//localhost:8124`.

So `validateAgentOsUrl` additionally asserts the protocol is `http:`/`https:`, that a hostname is present, and that the pathname does not start with `//` (the reliable signal of a doubled scheme). Note it deliberately does *not* blacklist `http`/`https` as hostnames, since those are syntactically legitimate (`https://http.example.com` is a valid host).

| Input | Outcome |
|---|---|
| `http://localhost:8124` | accepted |
| `https://agentos.internal` | accepted |
| `https://example.com/agentos` | accepted |
| `my-host` | rejected |
| `localhost:8124` | rejected |
| `http://http://localhost:8124` | rejected |
| `ftp://localhost:8124` | rejected |

## ⚠️ Breaking change

`AGENTOS_HOSTNAME` is **removed**. Any deployment setting it must migrate to `AGENTOS_URL` with the full URL **including the scheme**:

```diff
- AGENTOS_HOSTNAME=agentos.internal
+ AGENTOS_URL=https://agentos.internal
```

Deployments not setting it are unaffected — the default remains `http://localhost:${AGENTOS_PORT}`.

## On the port default question

The issue flagged an apparent inconsistency: `server.ts` defaults to `8124` while `web-dev-tmux.sh` scans from `8123`. Investigated and left unchanged — it is not a real discrepancy. `8124` is genuinely AgentOS's default (`server.port` in `application.yml`, and `${AGENTOS_PORT:8124}` in `application-docker.yml`); `8123` in the script is a *scan starting point*, and the script always passes the port it actually found to Express explicitly, so neither default is consulted in that path. `agentos/docs/running.md` did state a stale `8080` and has been corrected to `8124`.

## Known limitation: the new test does not run in CI

`apps/server/src/lib/validate-agentos-url.spec.ts` passes when jest is invoked directly, but **is not picked up by `nx affected -t test`** — the `server` project has no `test` target in Nx at all, so `nx test server` fails with "Cannot find configuration for task". The pre-existing `apps/server/src/lib/resolve-username.spec.ts` has the same problem and has never run in CI either.

This is a test-infrastructure gap independent of this fix, and wiring a `test` target for `server` here risks surfacing unrelated pre-existing failures in a change that should stay narrowly scoped. Tracked separately — see the follow-up issue referenced below.

## Verification

| Command | Result |
|---|---|
| `pnpm nx run server:build` | ✅ pass |
| `pnpm nx affected -t lint` | ✅ pass (server, web) |
| `pnpm exec jest validate-agentos-url.spec.ts` | ✅ 9/9 pass |
| `pnpm nx affected -t test` | no test targets affected (see limitation above) |

## Files changed

| File | Change |
|---|---|
| `apps/server/src/server.ts` | Removed `AGENTOS_HOSTNAME`; `AGENTOS_URL` now read from env with fail-fast validation |
| `apps/server/src/lib/validate-agentos-url.ts` | New validation helper |
| `apps/server/src/lib/validate-agentos-url.spec.ts` | Unit tests covering the accept/reject matrix above |
| `agentos/.env.example` | Documented `AGENTOS_URL` and its scheme requirement |
| `agentos/docs/running.md` | Corrected stale default port 8080 → 8124 |

### DIFF
--- a/agentos/.env.example
+++ b/agentos/.env.example
@@ -13,6 +13,10 @@ MCP_ENABLED=false
 # Server Configuration
 SERVER_PORT=8080
 
+# Express proxy target for the AgentOS backend.
+# Must be a complete URL including scheme. Defaults to http://localhost:<AGENTOS_PORT> when unset.
+# AGENTOS_URL=http://localhost:8124
+
 # Plugin Configuration
 PLUGINS_DIR=plugins
 PLUGINS_AUTO_LOAD=true
--- a/agentos/docs/running.md
+++ b/agentos/docs/running.md
@@ -16,7 +16,7 @@ export ANTHROPIC_API_KEY=sk-ant-...
 nx bootRun agentos-service
 ```
 
-The service starts on **port 8080**. Ready signal in logs: `Started AgentOSApplication`.
+The service starts on **port 8124** by default (see `server.port` in `application.yml`). Ready signal in logs: `Started AgentOSApplication`.
 
 ## Filesystem Plugin Directories
 
--- a/apps/server/src/lib/validate-agentos-url.spec.ts
+++ b/apps/server/src/lib/validate-agentos-url.spec.ts
@@ -1,0 +1,45 @@
+import { validateAgentOsUrl } from './validate-agentos-url'
+
+describe('validateAgentOsUrl', () => {
+  describe('valid URLs', () => {
+    it('accepts http with localhost and port', () => {
+      expect(() => validateAgentOsUrl('http://localhost:8124')).not.toThrow()
+    })
+
+    it('accepts https with a named host', () => {
+      expect(() => validateAgentOsUrl('https://agentos.internal')).not.toThrow()
+    })
+
+    it('accepts a URL carrying a base path', () => {
+      expect(() => validateAgentOsUrl('https://example.com/agentos')).not.toThrow()
+    })
+  })
+
+  describe('invalid URLs', () => {
+    it('rejects a bare hostname without scheme', () => {
+      expect(() => validateAgentOsUrl('my-host')).toThrow('Invalid AGENTOS_URL: "my-host"')
+    })
+
+    it('rejects localhost:port without scheme (parsed as scheme=localhost:)', () => {
+      expect(() => validateAgentOsUrl('localhost:8124')).toThrow('Invalid AGENTOS_URL: "localhost:8124"')
+    })
+
+    it('rejects a doubled scheme (http://http://...)', () => {
+      expect(() => validateAgentOsUrl('http://http://localhost:8124')).toThrow(
+        'Invalid AGENTOS_URL: "http://http://localhost:8124"'
+      )
+    })
+
+    it('rejects a non-http scheme such as ftp://', () => {
+      expect(() => validateAgentOsUrl('ftp://localhost:8124')).toThrow('Invalid AGENTOS_URL: "ftp://localhost:8124"')
+    })
+
+    it('error message names the offending value', () => {
+      expect(() => validateAgentOsUrl('bad-value')).toThrow('"bad-value"')
+    })
+
+    it('error message shows a valid example', () => {
+      expect(() => validateAgentOsUrl('bad-value')).toThrow('http://localhost:8124')
+    })
+  })
+})
--- a/apps/server/src/lib/validate-agentos-url.ts
+++ b/apps/server/src/lib/validate-agentos-url.ts
@@ -1,0 +1,29 @@
+/**
+ * Validates that a string is a well-formed http(s) URL suitable for use as
+ * the AgentOS proxy target.
+ *
+ * `new URL()` alone is too permissive: `new URL('localhost:8124')` succeeds
+ * by treating `localhost:` as the scheme, which is the most likely migration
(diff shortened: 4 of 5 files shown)


vincent-audibert-whoz:
The "Known limitation" section above refers to a pre-existing test-infrastructure gap that is now tracked separately: **#1213 — chore: server project has no Nx test target, its specs never run in CI**.

PR #1214, opened 2026-08-05
feat: wz-330620 scheduler initiate cases

fdelsert-whoz:
Scheduled prompts needed two missing capabilities: creating one Case **per target user**, and **disabling themselves** when their end condition is reached (`ON_DATE` or `OCCURRENCES`). Without the latter, a scheduled prompt with an expired end date would keep firing indefinitely.

## Approach

The implementation extends the existing two-phase tick architecture rather than introducing new scheduling infrastructure:

- **Phase A** handles the fast path: insert a CLAIMED Run, then materialize all target UserRuns in a single Cypher MERGE and transition to RUNNING in one `@Transactional`. When no target users exist (e.g. platform-scope), the Run transitions directly to DONE. The insert is outside the transaction, if the instance crashes between the insert and materialize, a sweep at the start of each tick marks orphaned CLAIMED Runs as FAILED to unblock subsequent slots.
- **Phase B** handles the slow path: claim UserRuns in batches, execute each as an independent Case with structured concurrency, await each Case's completion reactively via CaseRuntime.statusFlow, and check Run completion once per batch.

End condition checks are placed at two strategic points: **before** execution (crash-recovery safety net — prevents zombie ticks when the server was down while the condition expired, or when a previous tick crashed between `advanceNextRunAt` and the disable) and **after** advancing `nextRunAt` (normal path, disable when the next slot would exceed the boundary).

## Compromises

- **At-least-once delivery, not exactly-once.** No idempotence key links a Case to its UserRun. If an instance crashes after `CaseService.create()` but before `markTerminal()`, the lease expires and another instance reclaims the UserRun, creating a second Case. Acceptable for the use case (duplicate conversation, no data corruption). Exactly-once would require a UNIQUE constraint on Case keyed by `(runId, userId)`.

- **No dedicated `ScheduledPromptTargetUserResolver`.** User resolution and UserRun materialization are fused in a single Cypher MERGE traversing the deployment graph.

- **Neither fire-and-forget nor a fully reliable messaging system.** The Executor creates Cases and waits for their completion, with lease-based crash recovery and optimistic locking, but there is no dead-letter queue, no retry policy, and no exactly-once guarantee. This is a pragmatic middle ground: significantly more robust than fire-and-forget (which would lose track of in-flight Cases on crash), without the complexity of a proper message broker. It covers the current use case, periodic automated conversations, where an occasional duplicate or missed run after a crash is acceptable, and where the operational priority is simplicity over exhaustive delivery guarantees.

## Domain model

Three entities form the execution hierarchy:

**`ScheduledPrompt`** (existing, extended):  the aggregate root declaring what to run, when, and for whom. Extended with `Planning` (startDate, endType, endDate, maxOccurrenceCount) and `Recurrence` (unit, days, timeUtc). `enabled` is toggled to `false` when the end condition is reached. `nextRunAt` is advanced by CAS Cypher after each tick.

**`ScheduledPromptRun`** (existing): one per scheduler tick per ScheduledPrompt. Represents a single firing of the schedule. (`scheduledPromptId`,`scheduledFor`) is UNIQUE in Neo4j, acts as the distributed lock preventing duplicate firings for the same slot. Status: CLAIMED-> RUNNING -> DONE/FAILED (or SKIPPED on overlap). The CLAIMED -> RUNNING transition serves as the completion guard: `checkCompletion` only inspects RUNNING Runs, so a Run still being materialized is never prematurely closed.

**`ScheduledPromptUserRun`** (new) : one per `(Run, User)` pair. Tracks the execution of a single Case for a single target user. Composite UNIQUE constraint on `(runId, userId)`, idempotent via MERGE, safe to replay on crash. `@Version` optimistic locking prevents double-claim across instances. `leaseUntil` enables crash recovery: a RUNNING UserRun whose lease has expired is reclaimed by the next `findClaimable`. Status: PENDING -> RUNNING -> DONE/FAILED.

### Case creation per user

The Executor resolves the prompt content directly via `promptService.findById(promptTemplateId)` and creates a Case per user through the following sequence:

1. `CaseService.create()`: persists the Case with the scheduled prompt's name as title (e.g. "Weekly Digest")
2. `PermissionService.grantPermission()`: grants ADMIN on the Case to the target user
3. `CaseService.addMessage()`: injects `"@agentName <resolved prompt content>"` as the user's message, which internally launches `runtime.run()` in `CaseServiceImpl.scope`

The `@agentName` is resolved by `selectAgent()` in the Case runtime. This avoids the incompatibility between `@mention` prefix and `PromptCommandParser`'s `/command` requirement, and positions the Executor as the right resolution point for future placeholder support.

All values are configurable via `SchedulerProperties`:

| Property | Env var | Default | Description |
|---|---|---|---|
| `scheduler.enabled` | `SCHEDULER_ENABLED` | `false` | Enable/disable the scheduler |
| `scheduler.tick-interval-ms` | `SCHEDULER_TICK_INTERVAL_MS` | `30000` | Interval between claim ticks (ms) |
| `scheduler.consume-interval-ms` | `SCHEDULER_CONSUME_INTERVAL_MS` | `10000` | Interval between consume ticks (ms) |
| `scheduler.batch-size` | `SCHEDULER_BATCH_SIZE` | `20` | Max UserRuns claimed and executed in parallel per consume tick |
| `scheduler.launch-timeout-seconds` | `SCHEDULER_LAUNCH_TIMEOUT_SECONDS` | `30` | Max seconds to wait for a Case to reach IDLE or terminal after launch |
| `scheduler.lease-minutes` | `SCHEDULER_LEASE_MINUTES` | `30` | Lease duration for RUNNING UserRuns (minutes) |

### End condition enforcement

- **Pre-check guard**: if `ON_DATE` endDate is already past or `OCCURRENCES` max is already reached, the ScheduledPrompt is disabled without inserting a Run. This is a crash-recovery safety net, prevents zombie ticks when the server was down or a previous disable failed.
- **Post-advance check**: after advancing nextRunAt, if the next slot exceeds endDate or the total number of non-SKIPPED runs (including the one just inserted) reaches maxOccurrenceCount, the ScheduledPrompt is disabled.

### Multi-instance safety

The scheduler runs safely across multiple AgentOS instances with no leader election:

| Mechanism | Where | Guarantee |
|---|---|---|
| UNIQUE constraint | `ScheduledPromptRun` insert | Only one instance claims a given slot, losers catch `DuplicateRunException` and skip |
| Compare-And-Set Cypher on `nextRunAt` | `advanceNextRunAt` | Auto-repairing: every instance advances the slot, even on duplicate, prevents stuck prompts |
| `@Transactional` on `materialize()` | MERGE UserRuns + CLAIMED→RUNNING | Atomic: crash cannot leave orphaned PENDING UserRuns or a RUNNING Run with no UserRuns |
| MERGE Cypher on composite key | `materialize()` | Idempotent UserRun creation, safe to replay, UNIQUE constraint deduplicates |
| Orphan sweep on CLAIMED | `recoverOrphanedClaimedRuns` in `tickClaim` | Crash between Run insert and `materialize()` leaves a CLAIMED orphan blocking the overlap guard, swept to FAILED after 5 min, unblocking subsequent slots |
| Orphan sweep on RUNNING | `recoverOrphanedRunningRuns` in `tickClaim` | Crash between last `markTerminal` and batch completion check leaves a RUNNING Run with all UserRuns terminal, swept to DONE or FAILED on next tick |
| `@Version` optimistic lock | `claimBatch()` on UserRun | `OptimisticLockingFailureException` → skip (another instance took it) |
| `leaseUntil` expiry | RUNNING UserRuns | Instance crash recovery: if the instance dies between Case creation and `closeUserRun`, the lease expires and another instance reclaims the UserRun (at-least-once — may create a duplicate Case) |

### Known limitations / follow-up

- **No purge of completed Runs and UserRuns.** `ScheduledPromptRun` and `ScheduledPromptUserRun` are append-only audit records that accumulate indefinitely. A scheduled prompt firing daily for 100 users produces ~36,500 UserRun nodes per year. A purge mechanism (retention policy, TTL-based cleanup, or archival) is needed in a future PR to prevent unbounded growth.

### Architectural note
The two-phase tick, UNIQUE-constraint locking, lease-based recovery, and completion sweeps are a hand-rolled workflow engine built on Neo4j.
As the volume grows, migrating to a dedicated workflow engine would replace most of this machinery with battle-tested infrastructure.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
@@ -47,6 +47,9 @@ open class Neo4jScheduledPromptRunRepository(
     override fun findById(id: UUID): ScheduledPromptRun? =
         neo4jRepository.findById(id.toString()).orElse(null)?.toDomain()
 
+    override fun countCompletedRuns(scheduledPromptId: UUID): Int =
+        neo4jRepository.countCompletedRuns(scheduledPromptId.toString())
+
     private fun isSlotKeyConflict(e: DataIntegrityViolationException): Boolean {
         val haystack = generateSequence<Throwable>(e) { it.cause }
             .mapNotNull { it.message }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
@@ -54,6 +54,9 @@ open class Neo4jScheduledPromptRunRepository(
         neo4jRepository.findByStatusAndCreatedBefore(RunStatus.CLAIMED.name, olderThan)
             .map { it.toDomain() }
 
+    override fun findSettledRunning(): List<ScheduledPromptRun> =
+        neo4jRepository.findSettledRunning().map { it.toDomain() }
+
     private fun isSlotKeyConflict(e: DataIntegrityViolationException): Boolean {
         val haystack = generateSequence<Throwable>(e) { it.cause }
             .mapNotNull { it.message }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
@@ -2,6 +2,7 @@ package io.whozoss.agentos.scheduledPrompt
 
 import mu.KLogging
 import org.springframework.dao.DataIntegrityViolationException
+import java.time.Instant
 import java.util.UUID
 
 /**
@@ -26,6 +27,25 @@ open class Neo4jScheduledPromptRunRepository(
 
     override fun hasActive(scheduledPromptId: UUID): Boolean =
         neo4jRepository.existsActiveByScheduledPromptId(scheduledPromptId.toString())
+
+    override fun updateStatus(
+        id: UUID,
+        status: RunStatus,
+        finishedAt: Instant?,
+        error: String?,
+    ): Boolean {
+        val updated = neo4jRepository.updateStatus(
+            id = id.toString(),
+            status = status.name,
+            finishedAt = finishedAt,
+            error = error,
+            now = Instant.now(),
+        )
+        return updated > 0
+    }
+
+    override fun findById(id: UUID): ScheduledPromptRun? =
+        neo4jRepository.findById(id.toString()).orElse(null)?.toDomain()
 
     private fun isSlotKeyConflict(e: DataIntegrityViolationException): Boolean {
         val haystack = generateSequence<Throwable>(e) { it.cause }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
@@ -50,6 +50,10 @@ open class Neo4jScheduledPromptRunRepository(
     override fun countCompletedRuns(scheduledPromptId: UUID): Int =
         neo4jRepository.countCompletedRuns(scheduledPromptId.toString())
 
+    override fun findOrphanedClaimed(olderThan: Instant): List<ScheduledPromptRun> =
+        neo4jRepository.findByStatusAndCreatedBefore(RunStatus.CLAIMED.name, olderThan)
+            .map { it.toDomain() }
+
     private fun isSlotKeyConflict(e: DataIntegrityViolationException): Boolean {
         val haystack = generateSequence<Throwable>(e) { it.cause }
             .mapNotNull { it.message }
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
@@ -9,8 +9,9 @@ import java.util.UUID
  * Neo4j-backed implementation of [ScheduledPromptRunRepository].
  *
  * [insert] wraps the SDN save in a constraint-violation detector: if Neo4j rejects the write
- * because [ScheduledPromptRunNode.slotKey] is already taken, the [DataIntegrityViolationException]
- * is caught and translated to a [DuplicateRunException] so callers can handle it gracefully.
+ * because the composite `(scheduledPromptId, scheduledFor)` constraint is violated, the
+ * [DataIntegrityViolationException] is caught and translated to a [DuplicateRunException]
+ * so callers can handle it gracefully.
  */
 open class Neo4jScheduledPromptRunRepository(
     private val neo4jRepository: ScheduledPromptRunNodeNeo4jRepository,
@@ -20,7 +21,7 @@ open class Neo4jScheduledPromptRunRepository(
         try {
             neo4jRepository.save(ScheduledPromptRunNode.fromDomain(run)).toDomain()
         } catch (e: DataIntegrityViolationException) {
-            if (!isSlotKeyConflict(e)) throw e
+            if (!isSlotConflict(e)) throw e
             logger.warn { "[Neo4jScheduledPromptRunRepository] Duplicate slot: sp=${run.scheduledPromptId} for=${run.scheduledFor}" }
             throw DuplicateRunException(run.scheduledPromptId, run.scheduledFor)
         }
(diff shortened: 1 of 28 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -0,0 +1,477 @@
...
+    ) {
+        runCatching {
+            userRunRepository.markTerminal(userRunId, UserRunStatus.FAILED, now, error)
+        }.onFailure { e ->
+            logger.error(e) { "[Executor] Could not mark UserRun=$userRunId as FAILED" }
+        }
+    }
+
+    private fun buildCaseTitle(scheduledPromptName: String, userDisplayName: String): String =
+        "$scheduledPromptName — $userDisplayName".take(MAX_TITLE_LENGTH)
```
fdelsert-whoz: TODO: non spécifié dans le ticket.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerProperties.kt
```diff
@@ -9,12 +9,21 @@ import org.springframework.boot.context.properties.ConfigurationProperties
...
-    /** Set to false to disable the scheduler (useful in tests). */
-    val enabled: Boolean = true,
+    /** Disabled by default; enable in production via env var SCHEDULER_ENABLED=true. */
+    val enabled: Boolean = false,
     /** Interval between scanner ticks in milliseconds. */
     val tickIntervalMs: Long = 60_000L,
+    /** Maximum number of UserRuns claimed per batch. */
+    val batchSize: Int = 20,
+    /** Lease duration for RUNNING UserRuns in minutes. Expired leases are reclaimed by the next tick. */
+    val leaseMinutes: Long = 30L,
```
fdelsert-whoz, line 31: FIXME: valeur beaucoup trop élevée ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -0,0 +1,469 @@
...
+            }
+
+            // Step 5: Collect the runtime's statusFlow until IDLE or terminal.
+            // The runtime is guaranteed to exist in activeRuntimes right after create().
+            val runtime = caseService.findActiveRuntime(caseId)
+            if (runtime == null) {
+                // Runtime already evicted (terminal status reached before we got here)
+                closeUserRun(userRun.id, caseService.findById(caseId)?.status, caseId)
+            } else {
+                awaitCaseCompletion(userRun.id, caseId, runtime)
```
fdelsert-whoz: mettre un timeout ici ?

fdelsert-whoz: done sur awaitCaseCompletion 30 sec

frederic-delsert-whoz:
## PR Feedback — ScheduledPromptExecutor

### BLOCKING

**1. The `materialize` query does not resolve all eligible users**

The Cypher query in `ScheduledPromptUserRunNodeNeo4jRepository.materialize()` only traverses the path `AgentConfig → DEPLOYED_TO → UserGroup ← MEMBER|ADMIN ← User`. It misses users who have access to the namespace through a direct path (`User → Namespace`) or via the AgentConfig without going through a deployment UserGroup. The query should be aligned with the pattern used by `findEffective` in `ScheduledPromptNodeNeo4jRepository`, which handles `isAdmin`, accessible groups, and direct links to capture the full set of target users.

**2. Remove context resolution guards before Case launch**

In `ScheduledPromptExecutor.processUserRun()`, the "Resolve context" section chains 7 checks (`Run` → `ScheduledPrompt` → `namespaceId` → `User` → `Prompt` → `promptContent` → `AgentConfig`) before launching the Case. Most of these resolutions are repeated downstream by the runtime (`addMessage` resolves the agent, the prompt is already injected as plain text). Furthermore, the entry scope will evolve (currently AgentConfig, potentially something else tomorrow) — making these guards brittle. They should be reduced to the strict minimum needed to create the Case and inject the message.

**3. Set the claim tick default to 30 seconds instead of 60**

`SchedulerProperties.tickIntervalMs` currently defaults to 60,000 ms. Change it to 30,000 ms in `SchedulerProperties` (default value) and in `SchedulerScanner.tickClaim()` (`fixedDelayString`). A ScheduledPrompt due at 09:00:00 can currently wait up to 09:01:00 before being detected — 30 seconds halves this latency.

**4. Restore a delay between each Case launch (50ms)**

The `staggerDelayMs` was removed in this PR, but CI tests show flaky OOM failures when all Cases in a batch are launched simultaneously. Restore a `delay(50)` between each `launch` in `consumeAvailable()`. Root cause investigation (memory pressure on CI) should be handled in a separate ticket.

**5. Enable the scheduler by default **

`application.yml` currently sets `scheduler.enabled: ${SCHEDULER_ENABLED:false}`

### NON-BLOCKING

**6. Add a TIMEOUT status on UserRuns**

When `monitorLaunch()` reaches `launchTimeoutSeconds` without the Case completing, the UserRun is currently closed as `DONE`. Introduce a `UserRunStatus.TIMEOUT` to distinguish "the Case finished cleanly" from "the Case was still running when monitoring was released." This status would be considered terminal (does not prevent the parent Run from transitioning to DONE in `checkCompletion()`), but provides better audit visibility.

**7. Log a batch progress summary at the end of each batch**

At the end of each batch in `consumeAvailable()`, after the `coroutineScope` and before `checkCompletion`, log a summary per impacted Run: number of UserRuns processed, count by outcome (DONE/FAILED/TIMEOUT), and Runs involved. This enables progress tracking without digging through individual log entries.

**8. Handle resumption after a production outage exceeding the lease duration**

If production is down longer than `leaseMinutes` (30 min), RUNNING UserRuns have their lease expired. On restart, `claimBatch` reclaims them via `findClaimable` (expired lease) — creating a duplicate Case (at-least-once). The idea would be to detect on restart that a UserRun's Case already exists and is still active, and resume monitoring (`monitorLaunch`) rather than creating a new Case. Impacts `claimBatch()`, `processUserRun()`, and would require a persisted link between UserRun and Case (currently absent).

vincent-audibert-whoz:
About the 8. ☝️ , it is to discuss and assess as leaving behind the interrupted case could have some benefits in avoiding continuation on a case filled with potentially stale data (or incompatible post-migration).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
```diff
@@ -19,23 +21,55 @@ open class Neo4jScheduledPromptRunRepository(
...
+        return SLOT_CONSTRAINT_NAME in haystack ||
+            SLOT_PROPERTY_A in haystack ||
+            SLOT_PROPERTY_B in haystack
     }
 
     companion object : KLogging() {
-        private const val SLOT_KEY_CONSTRAINT_NAME = "scheduled_prompt_run_slot_key_unique"
-        private const val SLOT_KEY_PROPERTY = "slotKey"
+        private const val SLOT_CONSTRAINT_NAME = "scheduled_prompt_run_slot_unique"
+        private const val SLOT_PROPERTY_A = "scheduledPromptId"
```
vincent-audibert-whoz: Rename ?

vincent-audibert-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/Neo4jScheduledPromptRunRepository.kt
fdelsert-whoz: Remove

vincent-audibert-whoz (approved):

PR #1230, opened 2026-08-08
fix: wz-33620 improve coroutine handling in scheduled prompt UserRun consumption + other fixes

fdelsert-whoz:
This is a follow-up to https://github.com/whoz-oss/coday/pull/1214

Move the `try`/`catch` from inside processUserRun up to the launch site in the caller, so exceptions propagate correctly through the `supervisorScope` boundary.
Replace `coroutineScope` with `supervisorScope` so a failing UserRun does not cancel its siblings in the same batch.
Re-throw `CancellationException` explicitly to avoid silently swallowing coroutine cancellation from the parent scope.

Group UserRuns by runId so each group's completion check runs immediately after its UserRuns settle, removing the need for a separate post-batch sweep.

Extract resolveContext, createAndInjectCase and awaitLaunch to give each step a clear single responsibility.

## Follow-up fixes identified:

Fix materialize swallowing Neo4j failures: remove runCatching { }.getOrDefault(0) so exceptions propagate out of the `@Transactional` boundary. claim() catches the exception, marks the Run FAILED immediately with the error message, and logs the cause. Previously a transient Neo4j error silently produced count = 0 → DONE, losing the slot with no trace.

Isolate `claim()` failures per prompt: wrap each `claim(sp)` call and each orphan sweep in `runCatching` inside `tickClaim`. A single failing prompt no longer aborts the entire tick — subsequent prompts in the same tick are unaffected.

Fix checkEndConditionAfterAdvance overwriting the scheduler advance: replace `save(scheduledPrompt.copy(enabled = false))` with a targeted `updateEnabled(id, false)` Cypher query that touches only the `enabled` field. The full-aggregate save was writing the stale `nextRunAt` captured at the start of `claim()`, silently undoing the optimistic advance performed just above.

Add startup invariant for `leaseMinutes > launchTimeoutSeconds`: enforce via `require()` in `SchedulerScanner.logStartup()`. A lease shorter than the launch timeout allows UserRuns to be reclaimed and double-executed before `monitorLaunch` completes.

Collapse duplicate `when` branches in `monitorLaunch`: the two non-null branches both called `closeUserRun` — unified into a single `observedStatus != null` arm.

Remove forgotten `logger.info { "[DEBUG] slot=…" }` left in `claim()`.

Document `countCompletedRuns` intent: FAILED runs intentionally consume an occurrence of `maxOccurrenceCount` — the quota tracks "how many times the prompt fired", not "how many times it succeeded".

Fix KDoc divergences: `supervisorScope` vs `coroutineScope` in `consumeAvailable`, `batchSize` bounds concurrency (not `groupBy`), correct `monitorLaunch` StateFlow semantics, correct completion sweep description.

## Late additions

- **Occurrence counting on reschedule**: when a scheduled prompt is modified, only runs whose `scheduledFor` falls within the new planning window (`>= planning.startDate`) count toward the `maxOccurrenceCount` quota. Past executions outside the window are ignored.
- **LLM-generated conversation title**: conversation name is left `null` on creation, allowing the LLM to generate it lazily on first interaction.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
@@ -298,7 +298,6 @@ class ScheduledPromptExecutor(
         }
         return UserRunContext(
             namespaceId = namespaceId,
-            caseTitle = scheduledPrompt.name,
             actor = Actor(id = userRun.userId.toString(), displayName = user.displayName(), role = ActorRole.USER),
             // Inject resolved content with @mention — selectAgent resolves the agent,
             // PromptCommandParser sees no /command and passes text through unchanged.
@@ -316,7 +315,7 @@ class ScheduledPromptExecutor(
      * a UNIQUE constraint on Case keyed by (runId, userId).
      */
     private fun createAndInjectCase(userRun: ScheduledPromptUserRun, context: UserRunContext): UUID {
-        val case = caseService.create(Case(namespaceId = context.namespaceId, title = context.caseTitle))
+        val case = caseService.create(Case(namespaceId = context.namespaceId))
         permissionService.grantPermission(
             userRun.userId.toString(),
             EntityType.CASE,
@@ -350,7 +349,6 @@ class ScheduledPromptExecutor(
     /** Resolved execution context for a single [ScheduledPromptUserRun]. */
     private data class UserRunContext(
         val namespaceId: UUID,
-        val caseTitle: String,
         val actor: Actor,
         val message: String,
     )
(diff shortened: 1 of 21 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -114,8 +114,12 @@ class ScheduledPromptExecutor(
...
      *
      * Note: the Run insert happens in [SchedulerScanner.claim], **outside** this transaction.
-     * A crash between the insert and this call leaves an orphaned CLAIMED Run, which
-     * [SchedulerScanner.recoverOrphanedClaimedRuns] detects and marks FAILED on the next tick.
+     * A crash between the insert and this call, or an exception thrown by
+     * [ScheduledPromptUserRunRepository.materialize], leaves an orphaned CLAIMED Run —
+     * Spring rolls back the transaction so no partial UserRuns are written.
+     * [SchedulerScanner.recoverOrphanedClaimedRuns] detects such orphans and marks them
+     * FAILED on the next tick. The caller ([SchedulerScanner.claim]) is protected by
+     * `runCatching` so one failure does not block other prompts in the same tick.
```
vincent-couturier-whoz, line 122: question: I didn't see runCatching in the current code. It was in the previous version but it seems to have been removed. Where is it ?

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
fdelsert-whoz: The `runCatching` is at line 270 of `SchedulerScanner`.

vincent-couturier-whoz (dismissed):

vincent-couturier-whoz (approved):

PR #1231, opened 2026-08-08
fix: wz-33620 restrict UserRun materialisation to UserGroup deployment only

fdelsert-whoz:
This is a follow-up to https://github.com/whoz-oss/coday/pull/1214

---

The Cypher query was targeting all namespace members when the agent was deployed to the namespace, too broad. Now only users who are members of a UserGroup explicitly deployed on the agent are targeted.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptUserRunNodeNeo4jRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptUserRunNodeNeo4jRepository.kt
@@ -13,18 +13,20 @@ interface ScheduledPromptUserRunNodeNeo4jRepository : Neo4jRepository<ScheduledP
      * Traverses the deployment graph and materialises PENDING UserRuns in a single query.
      *
      * Resolves all non-removed Users that are **deployment targets** of [agentConfigId]
-     * in [namespaceId] via two paths:
+     * in [namespaceId] via UserGroup deployment only:
      *
-     * 1. **UserGroup deployment**: `AgentConfig -[:DEPLOYED_TO]-> UserGroup <-[:MEMBER|ADMIN]- User`,
-     *    where the UserGroup `[:BELONGS_TO]` the namespace.
-     * 2. **Namespace deployment**: `AgentConfig -[:DEPLOYED_TO]-> Namespace <-[:MEMBER|ADMIN]- User`.
+     * `AgentConfig -[:DEPLOYED_TO]-> UserGroup <-[:MEMBER|ADMIN]- User`,
+     * where the UserGroup `[:BELONGS_TO]` the namespace.
+     *
+     * Namespace-level deployment is intentionally excluded: being a member of the namespace
+     * does not mean the user is a target audience of this specific agent. Only users
+     * belonging to a UserGroup explicitly deployed to the agent are targeted.
      *
      * Super-admins are intentionally **excluded** unless they are also members of a
-     * deployed UserGroup or namespace. `isAdmin` grants the ability to *use* any agent
-     * (access control), but scheduled prompts target users who are *deployed to* the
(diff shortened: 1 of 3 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptUserRunNodeNeo4jRepository.kt
fdelsert-whoz, line 38: J'ai un doute sur la requête, je reteste ça via Neo4j Browser.

fdelsert-whoz: OK je retrouve bien mes users 👍

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptUserRunNodeNeo4jRepository.kt
```diff
@@ -37,19 +39,10 @@ interface ScheduledPromptUserRunNodeNeo4jRepository : Neo4jRepository<ScheduledP
...
-                    MATCH (a)-[:DEPLOYED_TO]->(g)
-                }
-                OR (
-                    EXISTS { MATCH (a)-[:DEPLOYED_TO]->(ns) }
-                    AND EXISTS { MATCH (u)-[:MEMBER|ADMIN]->(ns) }
-                )
-            )
         WITH DISTINCT u.id AS userId
         MERGE (ur:ScheduledPromptUserRun {runId: $runId, userId: userId})
         ON CREATE SET
```
vincent-couturier-whoz, line 48: suggestion: if there's lots of runs to create it could failed. apoc has a method to have intermediate commit to avoid to much resource consumption on neo4j. The issue here is that in case of error in one of the intermediate transaction you will have an inconsistent data.... so I think we can let this code as is but we should do some tests to see how much run can be supported by the production by trying to execute this request in preproduction.

vincent-couturier-whoz: question: it won't works in case of an agent deployed at the namespace level so it's strange. Perhaps we should indicate this information somewhere.

vincent-couturier-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptUserRunNodeNeo4jRepository.kt
fdelsert-whoz: The user scope accessible only through the user-group intermediary has been a need from the start. This ticket fixes the regression of adding support for the "deployed to namespace" targeting (to avoid targeting all users in the namespace).

Or do you mean that it would be a necessary condition (an AND rather than an OR as previously)?

fdelsert-whoz: OK, a UI evolution will be needed on the AgentOS side, as the available agents will also need to be restricted to those accessible via user-group.
For example, I think that Astra is selectable but will not trigger for any user.
I will create an AgentOS ticket for this point.

fdelsert-whoz: Or I leave it as is. Decline this PR ?
What I didn't have in mind is whether in practice the deployment methods are mutually exclusive?

frederic-delsert-whoz:
I am declining this PR as we want to preserve the ability to target an agent deployed via namespace.

PR #1232, opened 2026-08-13
feat(agentos): add star endpoints and list my cases by parent on SDK #33653

vincent-couturier-whoz:
## Context

`CaseApi` (`agentos/agentos-sdk/.../api/case/CaseApi.kt`) is the SDK's hand-written HTTP contract: it is implemented server-side by `CaseController`, and by external consumers — notably the whoz Copilot module — as a Feign client. It is therefore the only API surface an external consumer sees.

Three endpoints existed in `CaseController` but had never been declared in the interface:

- `PUT /api/cases/{id}/star`
- `DELETE /api/cases/{id}/star`
- `GET /api/cases/by-parentId/{parentId}/mine`

As a result, these routes were unreachable for any Feign consumer, and nothing guaranteed parity between the controller and the contract — the gap could grow silently.

## Changes

Declare the three missing operations on `CaseApi`, with `@Operation` describing their semantics:

- `starCase(id: UUID)` / `unstarCase(id: UUID)` — the description states that the call requires a direct user↔case relation, otherwise it is rejected (409).
- `listMineByParent(parentId: UUID): List<CaseDto>` — placed right after `listByParent` to keep both listing variants together; the description spells out the difference, namely that namespace-admin transitivity is excluded, which is what makes every returned case starrable by the caller.

The three matching `CaseController` methods become `override`.

## Design notes

The `override` is the real point of this PR: any future divergence between the controller and the SDK contract now breaks compilation instead of going unnoticed.

Existing `CaseApi` conventions are preserved: no Spring annotations, no `@PreAuthorize` and no `@Valid` on the interface — these are service-layer concerns, as documented in the interface KDoc. Parameter naming follows the controller (`id` for star/unstar, `parentId` for listings), which is a prerequisite for the `override` to compile.

## Side effect on the spec and the TS client

`agentos-openapi.yaml` and `case-controller.service.ts` are regenerated (`nx run agentos-service:regenerate`), since the spec is derived from the controller and verified in CI. The diff is **documentation only**: the `summary`/`description` inherited from the `@Operation` annotations. No signature, no `operationId` and no TS method changes — the TypeScript client already exposed `starCase`, `unstarCase` and `listMineByParentCase`, because it is generated from the controller and not from the interface. That very gap between the two layers is what hid the problem.

## Verification

- `agentos-sdk` + `agentos-service` compile
- `CaseControllerSpec` and `CaseControllerMvcIntegrationSpec` green (tests for all three endpoints, including the 409 cases, already existed)
- Spec regenerated and committed, consistent with the `check-openapi-spec` CI check

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
@@ -123,7 +123,7 @@ class CaseController(
      */
     @GetMapping("/by-parentId/{parentId}/mine")
     @PreAuthorize("hasPermission(#parentId, 'Namespace', 'READ')")
-    fun listMineByParent(
+    override fun listMineByParent(
         @PathVariable parentId: UUID,
     ): List<CaseDto> {
         val user = userService.getCurrentUser()
@@ -316,7 +316,7 @@ class CaseController(
     @PutMapping("/{id}/star")
     @ResponseStatus(HttpStatus.OK)
     @PreAuthorize("hasPermission(#id, 'Case', 'READ')")
-    fun starCase(
+    override fun starCase(
         @PathVariable id: UUID,
     ) {
         val userId = userService.getCurrentUser().id.toString()
@@ -328,7 +328,7 @@ class CaseController(
     @DeleteMapping("/{id}/star")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     @PreAuthorize("hasPermission(#id, 'Case', 'READ')")
-    fun unstarCase(
+    override fun unstarCase(
         @PathVariable id: UUID,
     ) {
         val userId = userService.getCurrentUser().id.toString()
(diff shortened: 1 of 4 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
```diff
@@ -123,7 +123,7 @@ class CaseController(
      */
     @GetMapping("/by-parentId/{parentId}/mine")
     @PreAuthorize("hasPermission(#parentId, 'Namespace', 'READ')")
-    fun listMineByParent(
+    override fun listMineByParent(
```
fdelsert-whoz, line 126: Hors scope car déjà en place mais on pourrait tendre vers quelque chose du genre :
/api/me/cases?parentId=

fdelsert-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseController.kt
vincent-couturier-whoz: à revoir mais je ne suis de toute façon pas convaincu par ce endpoint qui me semble bancal au niveau fonctionnel. Il faudrait dans nos échanges avenir sur nos PRs être plus en conventional comments en anglais car ce repo a une visibilité publique.

PR #1233, opened 2026-08-13
feat: add UserContextProvider SPI and resolver for scheduled case context enrichment

fdelsert-whoz:
When a PF4J plugin provides a `UserContextProvider`, the scheduled prompt executor resolves the user's session context and forwards it to the case as user context on the injected message, giving the agent the same context it would receive from a manually-started conversation.

Context resolution is non-fatal: a missing provider or a provider failure results in null sessionContext and execution continues normally.

- Add `UserContextProvider` SPI to _agentos-sdk_ (ExtensionPoint)
- Add `UserContextProviderResolver` in plugin/ (PF4J discovery, consistent with ToolRegistryService pattern)
- Add `PluginBeansConfiguration` to expose the resolved `UserContextProvider` as an optional Spring bean (resolved once at startup, not on every call)
- Wire provider directly into `ScheduledPromptExecutor` constructor — replaces `@Autowired(required = false)` field injection of the resolver
- `UserContextProviderResolverSpec`: no provider / single provider / multiple providers (first wins)
- `ScheduledPromptExecutorUnitSpec`: user context forwarded when provider returns a map, graceful fallback to null when provider throws

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
@@ -12,7 +12,7 @@ import io.whozoss.agentos.sdk.actor.Actor
 import io.whozoss.agentos.sdk.actor.ActorRole
 import io.whozoss.agentos.sdk.caseEvent.MessageContent
 import io.whozoss.agentos.sdk.caseFlow.CaseStatus
-import io.whozoss.agentos.plugin.UserContextProviderResolver
+import io.whozoss.agentos.sdk.scheduledPrompt.UserContextProvider
 import io.whozoss.agentos.user.UserService
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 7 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -300,13 +309,26 @@ class ScheduledPromptExecutor(
         val agentName = checkNotNull(agentConfigService.findById(scheduledPrompt.agentConfigId)?.name) {
             "AgentConfig ${scheduledPrompt.agentConfigId} not found"
         }
+        val sessionContext = runCatching {
+            userContextProviderResolver
+                ?.resolve()
```
vincent-couturier-whoz: suggestion: should be resolved in a configuration to provide the bean to spring once to avoid resolution on each call

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
fdelsert-whoz: done at commit: 8fbe3baf83ff5151598936a5e2482c216a8bb765

vincent-couturier-whoz (approved):

PR #1248, opened 2026-08-25
feat: #33829 share a case with other users (multi-user case)

vincent-audibert-whoz:
## WZ-33829 — Share a case with other users (multi-user case)

Frontend implementation for case membership management. The AgentOS backend was already ready.

### What's new

**`CaseMemberStateService`** — thin facade over the generated `CaseMembershipControllerService` + `UserControllerService`, following the two-layer pattern (same as `NamespaceMemberStateService`).

**`CaseMembersComponent`** — presentational panel to manage a case's ADMIN/MEMBER roster:
- List of current members with role selector (ADMIN / MEMBER) and remove button
- Add-member autocomplete (visible to super-admins only — backend enforces `GET /api/users` as SUPER_ADMIN)
- Own row is read-only: role shown as a badge, no remove button
- Anti-lockout guard: save is blocked if no ADMIN would remain
- Reuses `computeMemberDiff`, `hasAtLeastOneAdmin`, `memberLabel` from `namespace-members.util.ts`

**Integration in `CaseChatComponent`**:
- New `group_add` button in the case header opens the existing right-side drawer on the members panel
- The drawer now switches between two panels (`'files' | 'members'`) via a signal
- Files button explicitly sets the panel back to `'files'` before opening

### No new route
The members panel lives inside the drawer — no separate page, no router navigation.

### DIFF
--- a/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
+++ b/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
@@ -55,6 +55,7 @@ import { USER_PREFERENCES_PORT } from '../../services/user-preferences.service'
 import { ExchangeStateService } from '../../services/exchange-state.service'
 import { exchangeMutationScope } from '../../services/exchange-content.utils'
 import { ExchangeShellComponent } from '../exchange-shell/exchange-shell.component'
+import { CaseMembersComponent } from '../case-members/case-members.component'
 import { ComposerAttachmentsService } from '../composer-attachments/composer-attachments.service'
 import { ComposerAttachmentsComponent } from '../composer-attachments/composer-attachments.component'
 import { isNamespaceTargeted, resolveUploadScope } from '../composer-attachments/composer-attachments.utils'
@@ -119,6 +120,7 @@ function hasActiveSelection(): boolean {
     JsonPipe,
     DrawerComponent,
     ExchangeShellComponent,
+    CaseMembersComponent,
     PromptAutocompleteComponent,
     AgentAutocompleteComponent,
     BlueprintDirective,
@@ -232,6 +234,19 @@ export class CaseChatComponent implements OnInit, OnDestroy {
 
   /** Whether the delete confirmation inline is showing. */
   protected readonly confirmingDelete = signal(false)
+
+  /** Which panel is active in the right drawer: files exchange or case members. */
+  protected readonly drawerPanel = signal<'files' | 'members'>('files')
+
+  protected openFilesPanel(): void {
+    this.drawerPanel.set('files')
+    this.exchangeOpen.set(true)
+  }
+
+  protected openMembersPanel(): void {
+    this.drawerPanel.set('members')
+    this.exchangeOpen.set(true)
+  }
 
   // Header action outputs — handled by CaseShellComponent
   readonly starToggled = output<{ id: string; starred: boolean }>()
--- a/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
+++ b/libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
@@ -771,6 +771,7 @@ export class CaseChatComponent implements OnInit, OnDestroy {
     this.streamingText.set('')
     this.collapsedTools.set(new Set())
     this.isAtBottom.set(true)
+    this.drawerPanel.set('files')
     this.autocomplete.reset()
     this.attachments.reset()
     this.connectSse()
--- a/libs/agentos-ui/src/lib/components/case-members/case-members.component.html
+++ b/libs/agentos-ui/src/lib/components/case-members/case-members.component.html
@@ -27,30 +27,37 @@
         } @else {
           <ul class="case-members__members">
             @for (member of selectedMembers(); track member.userId) {
-              <li class="case-members__member">
+              <li
+                class="case-members__member"
+                [class.case-members__member--self]="member.userId === currentUserId()?.id"
+              >
                 <div class="case-members__member-info">
                   <span class="case-members__member-name">{{ member.label }}</span>
                   @if (member.email && member.email !== member.label) {
                     <span class="case-members__member-email">{{ member.email }}</span>
                   }
                 </div>
-                <select
-                  class="case-members__member-role"
-                  [value]="member.role"
-                  [attr.aria-label]="'Role of ' + member.label"
-                  (change)="setMemberRole(member.userId, $any($event.target).value)"
-                >
-                  <option value="MEMBER">Member</option>
-                  <option value="ADMIN">Admin</option>
-                </select>
-                <button
-                  class="case-members__member-remove"
-                  type="button"
-                  [title]="'Remove ' + member.label"
-                  (click)="removeMember(member.userId)"
-                >
-                  &#215;
-                </button>
+                @if (member.userId === currentUserId()?.id) {
+                  <span class="case-members__member-role-badge">{{ member.role | lowercase }}</span>
+                } @else {
+                  <select
+                    class="case-members__member-role"
+                    [value]="member.role"
+                    [attr.aria-label]="'Role of ' + member.label"
+                    (change)="setMemberRole(member.userId, $any($event.target).value)"
+                  >
+                    <option value="MEMBER">Member</option>
+                    <option value="ADMIN">Admin</option>
+                  </select>
+                  <button
+                    class="case-members__member-remove"
+                    type="button"
+                    [title]="'Remove ' + member.label"
+                    (click)="removeMember(member.userId)"
+                  >
+                    &#215;
+                  </button>
+                }
               </li>
             }
           </ul>
--- a/libs/agentos-ui/src/lib/components/case-members/case-members.component.html
+++ b/libs/agentos-ui/src/lib/components/case-members/case-members.component.html
@@ -1,0 +1,89 @@
+<div class="case-members">
+  <header class="case-members__header">
+    <h2 class="case-members__title">Share case</h2>
+    <button class="case-members__close" type="button" title="Close" (click)="cancel()">
+      <span class="material-icons" aria-hidden="true">close</span>
+    </button>
+  </header>
+
+  @if (isLoading()) {
+    <p class="case-members__loading">Loading&hellip;</p>
+  } @else {
+    <div class="case-members__body">
+      @if (!isSuperAdmin()) {
+        <p class="case-members__hint">
+          You can change roles or remove existing members. Only a platform super-admin can add a new user to this case.
+        </p>
+      }
+
+      <fieldset class="case-members__section">
+        <legend class="case-members__label">
+          Members
+          <span class="case-members__label-hint">ADMIN or MEMBER</span>
+        </legend>
+
+        @if (selectedMembers().length === 0) {
+          <p class="case-members__empty-hint">No members yet.</p>
+        } @else {
+          <ul class="case-members__members">
+            @for (member of selectedMembers(); track member.userId) {
+              <li class="case-members__member">
+                <div class="case-members__member-info">
+                  <span class="case-members__member-name">{{ member.label }}</span>
+                  @if (member.email && member.email !== member.label) {
+                    <span class="case-members__member-email">{{ member.email }}</span>
+                  }
+                </div>
+                <select
+                  class="case-members__member-role"
+                  [value]="member.role"
+                  [attr.aria-label]="'Role of ' + member.label"
+                  (change)="setMemberRole(member.userId, $any($event.target).value)"
+                >
+                  <option value="MEMBER">Member</option>
+                  <option value="ADMIN">Admin</option>
+                </select>
+                <button
+                  class="case-members__member-remove"
+                  type="button"
+                  [title]="'Remove ' + member.label"
+                  (click)="removeMember(member.userId)"
+                >
+                  &#215;
+                </button>
+              </li>
+            }
+          </ul>
+        }
+
+        @if (wouldLockOutAdmins()) {
+          <p class="case-members__warning">At least one ADMIN must remain on the case.</p>
+        }
+
+        @if (isSuperAdmin()) {
+          <ds-autocomplete-input
+            [dataSource]="memberDataSource"
+            placeholder="Add a member…"
+            (itemSelected)="onMemberSelected($event)"
+          />
+        }
+      </fieldset>
+
+      @if (errorMessage()) {
+        <p class="case-members__error">{{ errorMessage() }}</p>
+      }
+
+      <div class="case-members__actions">
+        <button
+          class="case-members__btn case-members__btn--primary"
+          type="button"
+          [disabled]="isSubmitting() || wouldLockOutAdmins()"
+          (click)="submit()"
+        >
+          Save
+        </button>
+        <button class="case-members__btn" type="button" (click)="cancel()">Cancel</button>
+      </div>
+    </div>
+  }
+</div>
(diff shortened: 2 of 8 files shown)


vincent-audibert-whoz (commented):
The overall structure is sound — the two-layer pattern is respected, signals are used for new state, control flow uses block syntax, and the utilities are correctly reused from `namespace-members`. Several issues still need addressing before merge, ranging from a latent reactivity bug to an orphaned method and a naming inconsistency.

**Correctness issues**: `wouldLockOutAdmins` is declared as a plain arrow function property, not a `computed()`. With `ChangeDetectionStrategy.OnPush` this means the template will not automatically re-evaluate it when `selectedMembers` changes — the anti-lockout guard and the disabled state of the Save button may be stale. It must be a `computed()`. Additionally, `reinitialise()` in `CaseChatComponent` does not reset `drawerPanel`, so switching cases while the members panel is open leaves it open on the new case.

**Awkward observable/signal interop in `loadData()`**: the condition `this.userState.currentUser() ? of(this.userState.currentUser()) : this.userState.loadMe()` reads a signal inside a method and wraps the result in `of()`, meaning `currentUser$` is always an `Observable<Signal<User|null>|null>` rather than `Observable<User|null>` — the value passed to `.subscribe()` is the signal itself, not the user. The intent is correct but the implementation is broken; use `of(this.userState.currentUser())` (reading the signal value) or simply call `loadMe()` unconditionally and let the service deduplicate.

**Orphaned method**: `toggleExchange()` in `CaseChatComponent` is no longer called from the template after the refactor (both buttons now call `openFilesPanel()` or `openMembersPanel()`), but the method is still present. Dead code should be removed.

**Naming inconsistency**: `CaseMemberStateService` holds no reactive state — it is a pure API facade, structurally identical to what the project calls an `*-api.service.ts`. Naming it a state service is misleading and contradicts the two-layer pattern documented in the architecture. Either add actual state (e.g. a cached member list signal) or rename it to `CaseMemberApiService` and introduce a thin state wrapper if needed.

**Minor issues**: `listAllUsers()` in `CaseMemberStateService` is missing an explicit return type (`Observable<User[]>`), inconsistent with `listMembers`. The `var(--red)` token used in the remove-button hover style is a raw CSS variable name that does not follow the `--color-*` design-token convention used everywhere else in the file — verify it exists in the design system or replace it with `var(--color-error)`.

libs/agentos-ui/src/lib/components/case-members/case-members.component.ts
```diff
@@ -0,0 +1,219 @@
...
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (refreshed) => {
+          this.applyMembers(refreshed)
+          this.isSubmitting.set(false)
+          this.closeRequested.emit()
+        },
+        error: (err: HttpErrorResponse) => {
+          this.isSubmitting.set(false)
+          this.errorMessage.set(this.describeError(err))
```
vincent-audibert-whoz, line 214: **Bug — plain function is not reactive with OnPush**

```ts
protected readonly wouldLockOutAdmins = () => !hasAtLeastOneAdmin(this.selectedMembers())
```

This is a regular arrow function, not a `computed()`. Angular's `ChangeDetectionStrategy.OnPush` only schedules re-renders when a signal notifies the component. Because this function is not wrapped in `computed()`, the template reads it as a static value at the last render cycle — changes to `selectedMembers()` will not trigger re-evaluation of `wouldLockOutAdmins` in the template.

The Save button's `[disabled]` binding and the warning paragraph both depend on this, so the anti-lockout guard can appear stale after a role change or a member removal. It must be declared as a `computed()`:

```ts
protected readonly wouldLockOutAdmins = computed(() => !hasAtLeastOneAdmin(this.selectedMembers()))
```

(And `computed` must be added to the Angular import list.)

libs/agentos-ui/src/lib/components/case-members/case-members.component.ts
```diff
@@ -0,0 +1,219 @@
...
+    this.isLoading.set(true)
+    this.errorMessage.set(null)
+
+    const currentUser$ = this.userState.currentUser() ? of(this.userState.currentUser()) : this.userState.loadMe()
+
+    currentUser$
+      .pipe(
+        catchError((err) => {
+          console.error('[CaseMembers] Failed to load current user', err)
+          return of(null)
```
vincent-audibert-whoz, line 103: **Bug — signal value not read correctly in `of()` branch**

```ts
const currentUser$ = this.userState.currentUser() ? of(this.userState.currentUser()) : this.userState.loadMe()
```

`this.userState.currentUser` is a `Signal<User | null>`. Calling `this.userState.currentUser()` reads the signal and returns `User | null`. When the truthy branch fires, `of(this.userState.currentUser())` wraps that `User | null` value correctly — but TypeScript infers the type of the whole expression as `Observable<User | null> | Observable<User>`, and the `.subscribe()` callback receives `User | null`, not a signal. So the logic is *technically* correct when the signal holds a user.

However, the pattern is fragile and misleading:
1. The signal is read **twice** (once for the condition, once inside `of()`), creating a tiny window where it could change between reads.
2. If `currentUser()` is `null` but `loadMe()` also fails, the `catchError` swallows the error and `isSuperAdmin` stays `false` silently — which is acceptable, but the comment should say so explicitly.
3. The cleaner idiom is to read once: `const cached = this.userState.currentUser(); const currentUser$ = cached ? of(cached) : this.userState.loadMe()`

Fix by reading the signal once into a local variable before branching.

libs/agentos-ui/src/lib/components/case-members/case-members.component.ts
```diff
@@ -0,0 +1,219 @@
...
+ * Roles are EXCLUSIVE (ADMIN xor MEMBER xor absent), each row has a single role <select>.
+ * Adding a new user requires SUPER_ADMIN (user directory is SUPER_ADMIN-only on the backend).
+ */
+@Component({
+  selector: 'agentos-case-members',
+  imports: [AutocompleteInputComponent, LowerCasePipe],
+  templateUrl: './case-members.component.html',
+  styleUrl: './case-members.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
```
vincent-audibert-whoz, line 53: **`ngOnChanges` fires on any input change — should guard on `caseId`**

```ts
ngOnChanges(): void {
  this.loadData()
}
```

`ngOnChanges` is called whenever *any* input changes. Currently `caseId` is the only input, so this is harmless. But the pattern is fragile: if a second `@input()` is ever added, `loadData()` will fire unnecessarily (e.g. on a `closeRequested` parent state change that rebinds an unrelated input).

Prefer the explicit guard:
```ts
ngOnChanges(changes: SimpleChanges): void {
  if (changes['caseId']) {
    this.loadData()
  }
}
```

(`SimpleChanges` needs to be imported from `@angular/core`.)

libs/agentos-ui/src/lib/components/case-members/case-members.component.html
```diff
@@ -0,0 +1,96 @@
...
+                @if (member.userId === currentUserId()?.id) {
+                  <span class="case-members__member-role-badge">{{ member.role | lowercase }}</span>
+                } @else {
+                  <select
+                    class="case-members__member-role"
+                    [value]="member.role"
+                    [attr.aria-label]="'Role of ' + member.label"
+                    (change)="setMemberRole(member.userId, $any($event.target).value)"
+                  >
+                    <option value="MEMBER">Member</option>
```
vincent-audibert-whoz, line 49: **`$any()` cast is an anti-pattern — use a typed event handler instead**

```html
(change)="setMemberRole(member.userId, $any($event.target).value)"
```

`$any()` bypasses TypeScript's type system in the template and is listed as an anti-pattern in the Angular guidelines. Extract the value in the component method:

```html
(change)="setMemberRole(member.userId, $event)"
```

```ts
protected setMemberRole(userId: string, event: Event): void {
  const value = (event.target as HTMLSelectElement).value
  // ...
}
```

This keeps the cast inside TypeScript where it is visible to the compiler and linters.

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -233,6 +235,19 @@ export class CaseChatComponent implements OnInit, OnDestroy {
   /** Whether the delete confirmation inline is showing. */
   protected readonly confirmingDelete = signal(false)
 
+  /** Which panel is active in the right drawer: files exchange or case members. */
```
vincent-audibert-whoz, line 238: **`reinitialise()` does not reset `drawerPanel`**

When the user switches cases (via `queryParams` subscription in `ngOnInit`), `reinitialise()` resets all component state — but `drawerPanel` is not included. If the members panel was open on the previous case, it will still be open (and showing the *previous* case's members, since `CaseMembersComponent.ngOnChanges` fires only when `[caseId]` actually changes) on the new case until the user manually dismisses it.

Add `this.drawerPanel.set('files')` to `reinitialise()` alongside the other signal resets.

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -233,6 +235,19 @@ export class CaseChatComponent implements OnInit, OnDestroy {
   /** Whether the delete confirmation inline is showing. */
   protected readonly confirmingDelete = signal(false)
```
vincent-audibert-whoz, line 236: **Orphaned `toggleExchange()` method**

The `toggleExchange()` method defined earlier in the file is no longer referenced in the template — both the files button and the members button now call `openFilesPanel()` / `openMembersPanel()` respectively. The method is dead code and should be removed to avoid confusion about the intended open/close contract.

libs/agentos-ui/src/lib/services/case-member-state.service.ts
```diff
@@ -0,0 +1,60 @@
+import { inject, Injectable } from '@angular/core'
```
vincent-audibert-whoz, line 1: **Naming inconsistency — `*-state.service.ts` holds no state**

`CaseMemberStateService` is a pure HTTP delegation layer with no signals, no `BehaviorSubject`, and no reactive state of any kind. Its structure is identical to what the project calls an `*-api.service.ts` (compare with `NamespaceApiService` or `ThreadApiService`). The two-layer architecture doc is explicit: state services own reactive state and business logic; API services are pure HTTP.

Naming this a state service violates the project convention and will mislead future contributors who expect a state service to expose reactive state.

Options:
1. **Rename to `CaseMemberApiService`** (accurate) and have `CaseMembersComponent` inject it directly (acceptable since there is no shared state to encapsulate).
2. **Add actual state** — e.g. a `members` signal caching the last-loaded member list — which would justify the state-service name and potentially remove the need for `selectedMembers` + `originalMembers` living inside the component.

Also: `listAllUsers()` is missing an explicit return type annotation (`Observable<User[]>`), inconsistent with `listMembers` and `updateMembers` in the same class.

libs/agentos-ui/src/lib/components/case-members/case-members.component.scss
```diff
@@ -0,0 +1,209 @@
...
+  &:focus {
+    outline: none;
+    border-color: var(--color-accent);
+  }
+}
+
+.case-members__member-role-badge {
+  font-size: 0.6875rem;
+  font-weight: 700;
+  text-transform: uppercase;
```
vincent-audibert-whoz, line 148: **`var(--red)` does not follow the design-token convention**

```scss
&:hover {
  color: var(--red);
  background: color-mix(in srgb, var(--red) 12%, transparent);
}
```

Every other color reference in this file uses `var(--color-*)` tokens (e.g. `--color-divider`, `--color-text`, `--color-accent`). `var(--red)` is a raw CSS variable that likely does not exist in the design-system token layer, which would silently fall back to the browser default (transparent/unset).

Replace with `var(--color-error)` (or whatever the design system's destructive-action token is) to stay consistent and guarantee the token resolves.

vincent-audibert-whoz (commented):
File-level comments attached above. Summary of items to address before merge:

1. **`wouldLockOutAdmins` must be `computed()`** — plain arrow function is not reactive with OnPush (correctness bug).
2. **Signal read in `loadData()` must be stored in a local variable** before branching into `of()` to avoid double-read and clarify intent.
3. **`ngOnChanges` should guard on `changes['caseId']`** to be resilient to future additional inputs.
4. **`$any($event.target).value` in template** — replace with a typed event handler in the component class.
5. **`reinitialise()` must reset `drawerPanel`** — otherwise switching cases leaves the members panel open on the new case.
6. **Remove orphaned `toggleExchange()` method** — dead code after the panel refactor.
7. **`CaseMemberStateService` naming** — it holds no state; rename to `*ApiService` or add actual reactive state to justify the name.
8. **`listAllUsers()` missing explicit return type** — inconsistent with the rest of the class.
9. **`var(--red)` in SCSS** — not a valid design-system token; replace with `var(--color-error)` or equivalent.

libs/agentos-ui/src/lib/components/case-members/case-members.component.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  protected readonly memberDataSource = new NamespaceMemberAutocompleteDataSource(
+    () => this.candidateUsers(),
+    () => new Set(this.selectedMembers().map((m) => m.userId))
+  )
+
+  constructor() {
+    effect(() => {
+      // Re-load whenever caseId changes (replaces ngOnChanges).
+      this.caseId()
```
vincent-audibert-whoz: **Bug introduced by the `effect()` migration — signal writes inside an effect body**

```ts
constructor() {
  effect(() => {
    this.caseId()
    this.loadData()   // ← calls isLoading.set(true) and errorMessage.set(null) synchronously
  })
}
```

`loadData()` immediately writes to two signals (`isLoading.set(true)` and `this.errorMessage.set(null)`) as its first two statements. Both calls execute synchronously, inside the `effect()` reactive context. Angular 17+ forbids writing to signals inside an `effect` and throws `NG0600: Writing to signals is not allowed in a computed or an effect` at runtime (unless `allowSignalWrites: true` is explicitly passed as the second argument to `effect()`).

The fix is either:
- Pass `{ allowSignalWrites: true }` as the second argument: `effect(() => { … }, { allowSignalWrites: true })` — straightforward but opts out of the safety guard.
- Or move the writes out of the synchronous effect path, e.g. by wrapping `loadData()` in `untracked(() => this.loadData())` — this also prevents `loadData`'s internal signal reads from accidentally registering extra dependencies on the effect.

The `untracked` approach is preferable: it makes the intent explicit ("only `caseId()` is the tracked dependency; everything inside `loadData` runs untracked") and keeps the safety guard active.

```ts
import { effect, untracked } from '@angular/core'

constructor() {
  effect(() => {
    const id = this.caseId()          // tracked — re-runs effect when caseId changes
    untracked(() => this.loadData())  // untracked — signal writes inside are allowed
  })
}
```

vincent-audibert-whoz (commented):
Follow-up pass on the 5 corrected points. Results:

✅ **Point 1 — `wouldLockOutAdmins` → `computed()`**: correctly done. `computed` is imported, the property is declared with `computed(() => !hasAtLeastOneAdmin(this.selectedMembers()))`, and the template calls it as a function `wouldLockOutAdmins()`.

✅ **Point 2 — `reinitialise()` resets `drawerPanel`**: correctly done. `this.drawerPanel.set('files')` is present in `reinitialise()` alongside the other signal resets.

✅ **Point 3 — local variable before branching in `loadData()`**: correctly done. `const currentUser = this.userState.currentUser()` is read once, then used in the ternary. No double-read.

✅ **Point 4 — `$any()` removed from template**: correctly done. The template now passes `$event`, and `setMemberRole` casts with `(event.target as HTMLSelectElement).value` in TypeScript.

⚠️ **Point 5 — `ngOnChanges` → `effect()`**: the direction is correct but the implementation introduces a new runtime error. `loadData()` writes to signals (`isLoading.set(true)`, `errorMessage.set(null)`) synchronously inside the `effect()` body, which Angular 17+ forbids (throws `NG0600`). The fix requires wrapping the `loadData()` call in `untracked()` — see the inline comment for details.

thomas-martin-whoz (approved):

PR #1250, opened 2026-08-26
feat(agentos): add actuator endpoint to disable / enable agentos schedulers #34024

vincent-couturier-whoz:
## Motivation

During incident response or maintenance, operators need a way to halt scheduler processing without restarting the service. Previously the two scheduler phases (claim and consume) had no runtime control mechanism.

### SchedulerScanner — pause guards

Two independent `AtomicBoolean` flags (`claimPaused`, `consumePaused`) are added with their symmetric public API:

- `pauseClaim()` / `resumeClaim()` — gate on `tickClaim()`
- `pauseConsume()` / `resumeConsume()` — gate on `tickConsume()`
- `isClaimPaused()` / `isConsumePaused()` — state accessors for the endpoint

When a phase is paused the `@Scheduled` tick still fires at its configured interval but returns immediately as a no-op. Resuming restores normal processing on the next tick with no side-effects.

`tickClaim()` is refactored: its body is extracted into `processClaim()` so the scheduled entry point remains a pure activation guard.

### SchedulerEndpoint — new Actuator endpoint

`SchedulerEndpoint` exposes the pause controls as a Spring Boot Actuator endpoint (`id = "scheduler"`), available over both HTTP and JMX:

- `GET  /management/scheduler`          → current paused state of each phase
- `POST /management/scheduler/{phase}`  → pause or resume (`phase`: claim / consume / all)

The endpoint is conditional on `agentos.prompt.scheduler.enabled=true`, consistent with `SchedulerScanner` itself.

### application.yml — endpoint exposure

`scheduler` is added to both the HTTP (`web`) and JMX Actuator exposure lists so the endpoint is reachable out of the box.

### Tests

**SchedulerScannerUnitSpec** — 6 new tests for `consumePaused` mirroring the existing `claimPaused` coverage:
- default state, pause, resume, idempotent calls
- guard isolation: pausing one phase does not affect the other

**SchedulerEndpointUnitSpec** — new spec, data-driven via `listOf/forEach` in `StringSpec`:
- `status`: 4 combinations of `(claimPaused, consumePaused)`
- `control`: 6 routing cases (claim/consume/all × pause/resume), each asserting the expected scanner method is called and no cross-phase call leaks
- invalid inputs: 6 cases covering unknown phase, unknown action, and empty strings — all must throw `IllegalArgumentException`

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
@@ -1,0 +1,93 @@
+package io.whozoss.agentos.scheduledPrompt
+
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation
+import org.springframework.boot.actuate.endpoint.annotation.Selector
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+/**
+ * Actuator endpoint for operational control of the [SchedulerScanner].
+ *
+ * Exposed both as an HTTP endpoint and as a JMX MBean by Spring Boot Actuator:
+ * - **HTTP**: `GET /management/scheduler` (read), `POST /management/scheduler/{phase}` (write)
+ * - **JMX**: MBean `org.springframework.boot:type=Endpoint,name=Scheduler`
+ *
+ * Spring Boot Admin discovers this endpoint automatically via Actuator discovery
+ * and can invoke it via Jolokia (HTTP→JMX bridge).
+ *
+ * ### Operations
+ *
+ * **Read** — returns the current paused/active state of each scheduler phase.
+ *
+ * **Write** — pause or resume a phase:
+ * - `phase`: `"claim"`, `"consume"`, or `"all"`
+ * - `action`: `"pause"` or `"resume"`
+ *
+ * Pausing a phase causes the corresponding `@Scheduled` tick to return immediately
+ * without processing. The Spring scheduling thread continues to fire ticks at the
+ * configured interval; they simply become no-ops. Resuming restores normal processing
+ * on the next tick. This is safe for crash recovery: sweeps resume naturally once
+ * the phase is unpaused.
+ */
+@Component
+@ConditionalOnProperty(name = ["agentos.prompt.scheduler.enabled"], havingValue = "true")
+@Endpoint(id = "scheduler")
+class SchedulerEndpoint(
+    private val schedulerScanner: SchedulerScanner,
+) {
+    /**
+     * Read current scheduler status.
+     *
+     * - HTTP:  `GET /management/scheduler`
+     * - JMX:   `Endpoint.Scheduler → status()`
+     */
+    @ReadOperation
+    fun status(): Map<String, Any> =
+        mapOf(
+            "claimPaused" to schedulerScanner.isClaimPaused(),
+            "consumePaused" to schedulerScanner.isConsumePaused(),
+        )
+
+    /**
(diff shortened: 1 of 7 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
```diff
@@ -0,0 +1,93 @@
...
+     * Pause or resume a scheduler phase.
+     *
+     * - HTTP:  `POST /management/scheduler/{claimOrConsumeOrAllPhase}` with body `{"pauseOrResumeAction": "pause"}` or `{"pauseOrResumeAction": "resume"}`
+     * - JMX:   `Endpoint.Scheduler → control(claimOrConsumeOrAllPhase, pauseOrResumeAction)`
+     *
+     * @param claimOrConsumeOrAllPhase  `"claim"`, `"consume"`, or `"all"`
+     * @param pauseOrResumeAction `"pause"` or `"resume"`
+     * @return updated status after applying the action
+     */
+    @WriteOperation
```
thomasmartin-whoz, line 63: shouldn't there be JMX documentation to help OPS ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
```diff
@@ -0,0 +1,93 @@
...
+        val apply: (String) -> Unit =
+            when (pauseOrResumeAction) {
+                "pause" -> { p -> if (p == "claim") schedulerScanner.pauseClaim() else schedulerScanner.pauseConsume() }
+
+                "resume" -> { p -> if (p == "claim") schedulerScanner.resumeClaim() else schedulerScanner.resumeConsume() }
+
+                else -> throw IllegalArgumentException(
+                    "Invalid action '$pauseOrResumeAction' (expected: pause, resume)",
+                )
+            }
```
thomasmartin-whoz, line 89: If feel 2 when with if conditions on claim into the pause/resume when plus lambda definitions is overengineered.

why not a simple when with guard conditions like 

```
require(pauseOrResumeAction in listOf("pause","resume")) {"Invalid action '$pauseOrResumeAction' (expected: pause, resume)"}
when (claimOrConsumeOrAllPhase) {
                "all" if(pauseOrResumeAction=="pause")-> {
schedulerScanner.pauseClaim()
schedulerScanner.pauseConsume()
                       }
"all" if(pauseOrResumeAction=="resume"){
                         schedulerScanner.resumeClaim()
                         schedulerScanner.resumeConsume()
                       }
                 }

                "claim" if(pauseOrResumeAction=="pause")->
schedulerScanner.pauseClaim()
"claim" if(pauseOrResumeAction=="resume")->schedulerScanner.resumeClaim()

                "consume" if(pauseOrResumeAction=="pause")-> schedulerScanner.pauseConsume()


                "consume" if(pauseOrResumeAction=="resume")-> schedulerScanner.resumeConsume()

                else -> throw IllegalArgumentException(
                    "Invalid phase '$claimOrConsumeOrAllPhase' (expected: claim, consume, all)",
                )
            }
```

thomas-martin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
vincent-couturier-whoz: It seems easier to maintain only in one place the call to the methods. Perhaps the require is cleaner than the throw but it seems not really more clear.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerEndpoint.kt
vincent-couturier-whoz: In this new way to do we cannot add documentation because it's primarily to expose actuator endpoint. It's for this reason I put the possible values in the arguments name.

PR #1253, opened 2026-08-27
feat: wz-30464: prompt translations 

thomas-martin-whoz:
Prompt Translation
Context
Copilot needs to display AgentOS prompts in the user's language. A prompt has two translatable surfaces: a title (a short user-facing label shown on a button to start a case) and content (the instruction elements sent to the agent when the case starts). These two surfaces have different semantic natures — one is user-facing and action-oriented, the other is directive and agent-facing — which matters for translation quality.

The feature needed to be additive with zero regression risk on existing prompts, and the translation infrastructure needed to live in AgentOS rather than Copilot, since AgentOS owns the prompt entity and is the natural place to cache translations.

Engineering Decisions
Caching at the entity level. Translations are stored directly on the Prompt node (translatedTitles, translatedContent) rather than in a separate entity. This avoids a join on every read and keeps the translation lifecycle tied to the prompt. Cached translations are invalidated automatically on update() when title, content, or sourceLanguage changes — each surface independently, so a content change does not wipe title translations and vice versa.

Per-element LLM calls for content. Each content element is translated in a separate LLM call rather than batching all elements in one call. This guarantees index alignment between source and translated lists regardless of how the model formats a multi-item response. The cost is more LLM calls on first translation, but subsequent calls are cache hits.

Source language short-circuit. When the requested language matches sourceLanguage, the originals are returned immediately without any LLM call. This is the common case for the prompt's authoring language and avoids unnecessary latency.

Distinct framing per surface kind. The LLM prompt uses different context paragraphs for title vs content elements. Early iterations used a single generic description, but the two surfaces are semantically different enough that giving the model accurate context produces better translations. The TextKind enum encodes this distinction internally without exposing it in the API.

sourceLanguage added to the translation prompt. The original Copilot template did not include the source language because it was not accessible at that call site. AgentOS has it on the entity, so it was added explicitly. Short strings are ambiguous across languages and the model benefits from the hint.

Trade-offs & Alternatives
The translation prompt instructions were adapted from an existing Copilot template. The original framing used Whoz product vocabulary ("conversation starter", "clickable button"). These were replaced with AgentOS vocabulary (prompt, title, content, agent) while preserving the structure and guidelines verbatim. The decision was to keep the proven instructions and only change the framing, not to author new instructions from scratch.

A separate Translation entity was considered but rejected. The entity-level cache is simpler, avoids a join, and the translation lifecycle is already fully determined by the prompt's own fields.

Assumptions
sourceLanguage has no enforced default at the domain level. The implementation assumes callers set it meaningfully. Prompts created without it will have null, which the translation endpoint will pass to the LLM as-is.

Limitations / Follow-up
The endpoint is not called from anywhere in AgentOS today — it exists solely for Copilot to consume. Real-world validation of translation quality and caching behaviour will happen when Copilot integrates against it. The languageCode path parameter is a free-form string with no validation against a fixed list; AgentOS uses it as both the cache key and the target language injected into the LLM prompt.


agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptDto.kt
```diff
@@ -50,6 +67,33 @@ data class PromptDto(
...
+    )
+    val sourceLanguage: String = "en",
+    @field:Schema(
+        description = "On-demand translations of title, keyed by BCP-47 language code. " +
+            "Each value is a single translated string. " +
+            "Null when title is null or no title translations have been generated. " +
+            "Cleared automatically when title or sourceLanguage is updated.",
+        nullable = true,
+    )
+    val translatedTitles: Map<String, String>? = null,
```
mathieu-dearmey: translatedTitles + translatedContent

is Copilot supposed to read translations directly from GET, or always use the dedicated endpoint?  If it's always the endpoint, why expose directly in DTO?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptServiceImpl.kt
```diff
@@ -55,7 +56,8 @@ class PromptServiceImpl(
             .findByTriple(entity.namespaceId, entity.userId, entity.name)
             ?.takeIf { it.id != entity.id }
             ?.let { throw ConflictException(conflictMessage(entity)) }
-        return saveOrConflict(entity)
+        val existing = repository.findByIds(listOf(entity.id)).firstOrNull()
```
mathieu-dearmey: update() calls findByIds twice, once in `rejectIfFilesystemBacked` and once for `val existing`. You should be able to use the same call for both usages

agentos/agentos-service/src/main/resources/application.yml
```diff
@@ -103,13 +103,16 @@ management:
 # datadog/agent:latest
 
 agentos:
+  encryption:
+    key: ${AGENTOS_ENCRYPTION_KEY}
+    salt: ${AGENTOS_ENCRYPTION_SALT}
```
mathieu-dearmey, line 108: encryption keys without default values will break startup if env vars aren't set. Can you document what needs to be configured to run locally?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptDto.kt
thomasmartin-whoz: I think that is a good point, we should always use the translate API except for the original content, I will remove this and have them only in translation endpoint return DTO.

agentos/agentos-service/src/main/resources/application.yml
thomasmartin-whoz: ${AGENTOS_ENCRYPTION_KEY:NONE} and salt: ${AGENTOS_ENCRYPTION_SALT:NONE} defaults to application.yml. FieldEncryptorConfiguration already treats NONE/NONE as an explicit opt-out to NoOpFieldEncryptor with a warning — the default just makes that the out-of-the-box behaviour for local dev without requiring any env var setup.

mathieu-de-armey-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationService.kt
```diff
@@ -0,0 +1,48 @@
...
+    ): List<String>
+
+    /**
+     * Translates a single [title] string from [sourceLanguage] to [targetLanguage].
+     */
+    fun translateTitle(
+        title: String,
+        sourceLanguage: String,
+        targetLanguage: String,
+        namespaceId: UUID?,
```
vincent-couturier-whoz, line 45: question: why namespaceId and externalId are present in a low level method ? How could you resolve externalId before to have only namespaceId without any doubt ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
```diff
@@ -0,0 +1,161 @@
...
+        val resolvedNamespaceId = resolveNamespaceId(namespaceId, namespaceExternalId)
+        val model = aiModelService.findAiModel(resolvedNamespaceId)
+            ?: throw BadRequestException(
+                "No default AI model configured for namespace $resolvedNamespaceId -- cannot translate prompt",
+            )
+        val provider = aiProviderService.getById(model.aiProviderId)
+        return model to provider
+    }
+
+    private fun resolveNamespaceId(namespaceId: UUID?, namespaceExternalId: String?): UUID =
```
vincent-couturier-whoz, line 86: suggestion: It shouldn't be the responsibility of this service to resolve namespace it should managed in the caller to avoid mixed responsibility.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Distinguishes the two kinds of text a [Prompt] exposes, so the LLM receives
+     * accurate context about what it is translating.
+     *
+     * - [TITLE] -- a short user-facing label shown on a button that starts a case.
+     *   It summarises the prompt's purpose for the user, not for the agent.
+     * - [CONTENT] -- one instruction element sent to the agent when the case starts.
+     *   It is directive in tone and may be technical or detailed.
+     */
+    private enum class TextKind { TITLE, CONTENT }
```
vincent-couturier-whoz, line 71: suggestion: we could define a method in this class to avoid when in the translateText

agentos/agentos-service/src/main/resources/application.yml
```diff
@@ -103,13 +103,16 @@ management:
 # datadog/agent:latest
 
 agentos:
+  encryption:
+    key: ${AGENTOS_ENCRYPTION_KEY}
```
vincent-couturier-whoz, line 107: issue: it makes no sens to have those env variable here because it's a spring convention which already managed by spring itself

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/prompt/PromptServiceImplSpec.kt
```diff
@@ -558,6 +570,340 @@ class PromptServiceImplSpec : StringSpec() {
...
+            )
+            val updated = service.update(saved.copy(description = "touched"))
+            updated.translatedTitles shouldBe existingTitles
+        }
+
+        // -------------------------------------------------------------------------
+        // clearTranslationsIfStale — independence of the two maps
+        // -------------------------------------------------------------------------
+
+        "update clears only translatedContent when content changes but title is unchanged" {
```
vincent-couturier-whoz, line 722: suggestion: use data driven test to simplify reading of tests

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
```diff
@@ -0,0 +1,161 @@
...
+        targetLanguage: String,
+        namespaceId: UUID?,
+        namespaceExternalId: String?,
+    ): List<String> {
+        val (model, provider) = resolveModelAndProvider(namespaceId, namespaceExternalId)
+        val chatClient = chatClientProvider.getChatClient(model, provider)
+        return content.map { element -> translateText(element, sourceLanguage, targetLanguage, chatClient, TextKind.CONTENT) }
+    }
+
+    override fun translateTitle(
```
vincent-couturier-whoz, line 46: suggestion: we could have a cache by splitting the method into two call with just a text, a lang, a model and a provider to avoid calling LLM in case of previous "translated" text to avoid calling the same translation of all federations.

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
vincent-audibert-whoz: We will need soon to outsource the default chat client (through namespace then default model) to a service because it is also used in the auto-name feature and will be needed for every technical LLM call.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationService.kt
thomas-martin-whoz: you are right, it should be done in Controller

agentos/agentos-service/src/main/resources/application.yml
thomas-martin-whoz: well I had to add it to start the service so it definitely has a purpose

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
thomas-martin-whoz: I tend to avoid caching usually but there it can lower our cost of tokens so I think this is a great idea.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
thomas-martin-whoz: there is ExternalIdentifierResolver that already exists for this, I will use it.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptDto.kt
```diff
@@ -50,6 +61,18 @@ data class PromptDto(
...
     val externalMetadata: Map<String, Any?>? = null,
+    @field:Schema(
+        description = "Optional user-facing display label (e.g. starter button text). " +
+            "Distinct from content, which is the message sent to the LLM. " +
+            "Null when no explicit title has been set.",
+        nullable = true,
+    )
+    val title: String? = null,
+    @field:Schema(
+        description = "BCP-47 language code of the authored content and title. Defaults to 'en'.",
```
thibault-drillien-whoz, line 72: BCP-47, c'est bien le bon format ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptController.kt
```diff
@@ -302,8 +309,41 @@ class PromptController(
...
+    )
+    @PostMapping("/{id}/translations/{languageCode}", consumes = [MediaType.APPLICATION_JSON_VALUE])
+    @PreAuthorize("hasPermission(#id, 'Prompt', 'READ')")
+    @HideOnAccessDenied
+    override fun translate(
+        @PathVariable id: UUID,
+        @PathVariable languageCode: String,
+        @Valid @RequestBody request: PromptTranslateRequest,
+    ): PromptTranslationDto {
+        val namespaceId = externalIdentifierResolver.resolveNamespaceId(
```
thibault-drillien-whoz: On a pas de vérification de cohérence entre le namespace du promptId et celui founit dans la request. Et au final si on doit réaliser ce contrôle, on pourrait ne pas fournir le parametre request et déduire le namespace à partir de l'id du prompt ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
```diff
@@ -0,0 +1,177 @@
...
+    private fun translateText(
+        text: String,
+        sourceLanguage: String,
+        targetLanguage: String,
+        chatClient: org.springframework.ai.chat.client.ChatClient,
+        kind: TextKind,
+        modelId: UUID,
+    ): String {
+        val key = CacheKey(text, sourceLanguage, targetLanguage, modelId, kind)
+        translationCache.getIfPresent(key)?.let { return it }
```
thibault-drillien-whoz: Plutot que de gérer manuellement le cache, proposition:
return translationCache.get(key,{ key ->...
Et supprimer le  translationCache.put(key, translated)

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/prompt/PromptDto.kt
thomas-martin-whoz: BCP-47 is the correct standard for language tags (en, fr, zh-Hant, etc.)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
```diff
@@ -0,0 +1,173 @@
...
+    /** Cache key: model-agnostic translation unit. */
+    private data class CacheKey(
+        val text: String,
+        val sourceLanguage: String,
+        val targetLanguage: String,
+        val modelId: UUID,
+        val kind: TextKind,
+    )
+
+    private val translationCache: Cache<CacheKey, String> =
```
vincent-couturier-whoz, line 50: suggestion: in another PR we could/should managed cache to be able to manage the cache using springboot admin

agentos/agentos-service/src/main/resources/application.yml
vincent-couturier-whoz: suggestion: have a default value will solve this issue but this configuration seems weird

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/prompt/PromptServiceImplSpec.kt
```diff
@@ -558,6 +593,322 @@ class PromptServiceImplSpec : StringSpec() {
             effective.shouldBeEmpty()
         }
 
+        // -------------------------------------------------------------------------
+        // title
+        // -------------------------------------------------------------------------
+
+        "create persists title when set" {
```
vincent-couturier-whoz, line 600: suggestion: have a data driven spec to simplify the check of the test coherency. Otherwise it takes more times to read all scenarios

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/prompt/PromptServiceImplSpec.kt
```diff
@@ -558,6 +593,322 @@ class PromptServiceImplSpec : StringSpec() {
             effective.shouldBeEmpty()
         }
 
+        // -------------------------------------------------------------------------
```
vincent-couturier-whoz, line 596: question: What's the purpose of those tests ?

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImplSpec.kt
```diff
@@ -0,0 +1,205 @@
...
+        // -------------------------------------------------------------------------
+
+        "translateTitle falls back to original on bad LLM response" {
+            table(
+                headers("scenario", "llmResponse"),
+                row("blank",          ""),
+                row("whitespace only", "   "),
+            ).forAll { _, llmResponse ->
+                every { chatClientProvider.getChatClient(model, provider) } returns chatClientReturning(llmResponse)
+                service.translateTitle("Review profile", "en", "fr", namespaceId) shouldBe "Review profile"
```
vincent-couturier-whoz, line 141: suggestion: it would be useful to know the behavior of the cache in case of transient error from LLM. it would be useful to avoid caching the "no result" of the translation to allow ability to have the translation

vincent-couturier-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/prompt/PromptTranslationServiceImpl.kt
thomas-martin-whoz: yes, ticket created https://github.com/whoz-oss/coday/issues/1274

frederic-delsert-whoz: suggestion: add a TODO ?

Switch to AsyncCache<CacheKey, String> (Caffeine.buildAsync()) when Spring AI exposes a non-blocking call API.
Today it blocks a Servlet thread, regardless, so AsyncCache would only shift the blocked thread to a different pool, no real gain.
Concurrent identical translation requests are also rare in practice, making the trade-off unjustified for now.
This should be revisited when Spring AI exposes a non-blocking API.

PR #1254, opened 2026-08-27
feat: wz-34012: identify scheduler that initiated a case

thomas-martin-whoz:
Context
The scheduler creates Case entities on behalf of users when a ScheduledPrompt fires. There was no way for downstream consumers — other services, analytics, or future agent logic — to distinguish a scheduler-initiated case from one opened by a human or a delegation tool. The Case entity carried no provenance information beyond its title, which is a fragile and informal signal.

Engineering Decisions
scheduledPromptId as a nullable UUID, not a boolean flag. A boolean scheduledBySystem: Boolean would answer "was this automated?" but nothing more. Storing the actual ScheduledPrompt id preserves the link back to the originating configuration, which is more useful for filtering, auditing, and potential future joins without requiring a schema change later. Null means user- or delegation-initiated; a non-null value means scheduler-initiated and identifies which prompt was responsible.

Field placed on Case, not on CaseEvent. The prompt injection manifests as a MessageEvent, but provenance is a property of the conversation itself, not of any individual event. Placing it on Case means any consumer that fetches the case gets the information without having to scan the event log.

Propagated through all layers without special-casing. The field flows from the domain (Case) through persistence (CaseNode) to the API contract (CaseDto), following the existing round-trip pattern used by parentCaseId. No layer is aware that the field originates from the scheduler; each layer simply carries what it is given. This keeps the scheduler as the sole writer and avoids coupling the persistence or HTTP layers to scheduler-specific logic.

Set in ScheduledPromptExecutor.createAndInjectCase, not in CaseService. The service has no knowledge of scheduled prompts and should not acquire it. The executor is the only caller that knows the originating ScheduledPrompt id at case-creation time, so it is the correct place to set the field. The UserRunContext data class was extended to carry scheduledPromptId so the value is resolved once during context resolution and flows cleanly into case creation.

Trade-offs & Alternatives
A source: CaseSource enum (USER, SCHEDULER, DELEGATION) was briefly considered. It would be more self-documenting at the cost of requiring a schema migration whenever a new source is added. The nullable foreign-key approach is more flexible and sufficient for the current use case.

Storing the ScheduledPromptRun id instead of the ScheduledPrompt id was also considered. The run id would identify a specific execution slot, but the prompt id is the more durable reference — it points to the configuration that drove the execution, which is what consumers are likely to need for filtering or display.

Assumptions
The field is intentionally write-once: once a case is created with a scheduledPromptId, nothing updates it. The PUT /api/cases/{id} endpoint already guards against mass assignment of status and namespaceId; scheduledPromptId is not exposed as an updatable field in the controller, so no additional guard was needed.

Limitations / Follow-up
The field is currently informational only — no query, index, or repository method filters by scheduledPromptId. If listing all cases for a given scheduled prompt becomes a product requirement, a dedicated findByScheduledPromptId query and a Neo4j index on the property will be needed.

### DIFF
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/case/CaseDto.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/case/CaseDto.kt
@@ -41,10 +41,6 @@ data class CaseDto(
     val status: CaseStatus = CaseStatus.PENDING,
     val title: String? = null,
     val parentCaseId: UUID? = null,
-    /**
-     * Id of the ScheduledPrompt that triggered this case.
-     * Null for cases started by a human user or a delegation tool.
-     */
     val scheduledPromptId: UUID? = null,
     val created: Instant? = null,
     val modified: Instant? = null,
--- a/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/case/CaseDto.kt
+++ b/agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/case/CaseDto.kt
@@ -29,6 +29,8 @@ import java.util.UUID
  *   `listByUserExternalId` where the target user may differ from the caller. Defaults to false on create.
  * @property role The caller's direct relation on this case (`ADMIN` or `MEMBER`), or null when the
  *   caller has no direct edge (e.g. transitive namespace-admin access). Same population rules as `favorite`.
+ * @property scheduledPromptId Id of the ScheduledPrompt that triggered this case, or null when the
+ *   case was started by a human user or a delegation tool.
  */
 @Schema(name = "Case")
 @JsonIgnoreProperties(ignoreUnknown = true)
@@ -39,6 +41,11 @@ data class CaseDto(
     val status: CaseStatus = CaseStatus.PENDING,
     val title: String? = null,
     val parentCaseId: UUID? = null,
+    /**
+     * Id of the ScheduledPrompt that triggered this case.
+     * Null for cases started by a human user or a delegation tool.
+     */
+    val scheduledPromptId: UUID? = null,
     val created: Instant? = null,
     val modified: Instant? = null,
     /**
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNode.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNode.kt
@@ -21,6 +21,8 @@ data class CaseNode(
     val status: String,
     val title: String,
     val parentCaseId: String? = null,
+    /** Id of the ScheduledPrompt that triggered this case. Null for user-initiated cases. */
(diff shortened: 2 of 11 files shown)


agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/case/CaseDto.kt
```diff
@@ -39,6 +41,11 @@ data class CaseDto(
     val status: CaseStatus = CaseStatus.PENDING,
     val title: String? = null,
     val parentCaseId: UUID? = null,
+    /**
+     * Id of the ScheduledPrompt that triggered this case.
+     * Null for cases started by a human user or a delegation tool.
+     */
```
mathieu-dearmey: déjà documenté +haut dans `@property`

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -371,6 +378,7 @@ class ScheduledPromptExecutor(
         val caseTitle: String,
         val actor: Actor,
         val message: String,
+        val scheduledPromptId: UUID,
```
mathieu-dearmey, line 381: non nullable. Pourrait poser pbl pour case non schedulé (?)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNode.kt
```diff
@@ -44,6 +46,7 @@ data class CaseNode(
             status = CaseStatus.valueOf(status),
             title = title,
             parentCaseId = parentCaseId?.let { UUID.fromString(it) },
+            scheduledPromptId = scheduledPromptId?.let { UUID.fromString(it) },
```
mathieu-dearmey, line 48: prévoir test comme pour parentCaseId pour la conversion string -- uuid

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
thomasmartin-whoz: on est justement dans la classe qui exécute les scheduler, donc c'est normal

vincent-audibert-whoz (dismissed):

mathieu-de-armey-whoz (dismissed):

vincent-audibert-whoz (dismissed):

pnpm-lock.yaml
vincent-couturier-whoz, line 1: question: why this change ? it's strange to have a change on pnpm lock when you have no change on package.json. What's your version of pnpm ?

vincent-couturier-whoz (dismissed):

pnpm-lock.yaml
thomas-martin-whoz: reverted

vincent-audibert-whoz (approved):

PR #1269, opened 2026-09-01
feat: #34318 sync filesystem agents into Neo4j for addressability

vincent-audibert-whoz:
## Problem

File-defined agents (loaded from YAML under a namespace `configPath`) had no persistent Neo4j node, which prevented them from being referenced by `ScheduledPrompt` and `UserGroup`. The scheduler would immediately disable any `ScheduledPrompt` pointing at a file-origin agent (its `findById` guard returned null), and the `UserGroup.addAgents` Cypher query silently produced no edge (the `MATCH (a:AgentConfig {id: agentId})` found nothing).

The previous `isFilesystemOnly` property — inferred from `metadata.version == null` — was a fragile proxy for this state and leaked implementation details of Spring Data Neo4j into business logic.

## Solution

**Lazy Neo4j sync driven by the existing TTL cache cycle.**

After each cache reload in `FilesystemAgentConfigRepository` (at most once per TTL window per namespace), a new `FilesystemAgentConfigSyncService` is called with the full list of currently-live filesystem agents for that namespace. It:

1. **Upserts** each live agent into Neo4j via `Neo4jAgentConfigRepository` (injected directly, not the composite decorator, to avoid a feedback loop), preserving the deterministic UUID derived from the agent name.
2. **Soft-deletes** any `fileOrigin = true` node whose name no longer appears in the live set — covering file deletion and renaming cleanly without leaving orphaned nodes that would keep a `ScheduledPrompt` or `UserGroup` pointing at a gone agent.

API-managed nodes (`fileOrigin = false`) are never touched.

**`fileOrigin: Boolean = false`** is added as an explicit field on both `AgentConfigNode` and `AgentConfig`, replacing the implicit `metadata.version == null` check everywhere. Filesystem agents produced by `parseYamlFile` carry `fileOrigin = true` from construction.

The guards in `PromptServiceImpl` and `ScheduledPromptServiceImpl` that rejected file-origin agents are removed — now that these agents have a real Neo4j node, they are valid targets for both entity types.

## Changes

- `AgentConfig` / `AgentConfigNode`: add `fileOrigin` field, remove `isFilesystemOnly`
- `FilesystemAgentConfigRepository`: accept optional `syncCallback`, invoke it after cache reload
- `FilesystemAgentConfigSyncService`: new `@Service` implementing the upsert + purge logic
- `Neo4jPersistenceConfiguration`: wire `FilesystemAgentConfigSyncService` as `syncCallback`; split the agent config bean into a delegate + primary composite (mirrors the existing `IntegrationConfig` and `Prompt` pattern)
- `PromptServiceImpl` / `ScheduledPromptServiceImpl`: remove `fileOrigin` guards
- `FilesystemAgentConfigRepositoryUnitSpec`: add `fileOrigin = true` assertion


vincent-audibert-whoz:
## Code Review — `feat: #34318 sync filesystem agents into Neo4j for addressability`

This PR is architecturally sound in its overall approach. Several correctness and robustness issues need addressing before merge.

---

### 🔴 Critical — Sync callback not triggered from `findByParent`

The KDoc on `FilesystemAgentConfigRepository` states the sync fires after each cache reload in `findByParent`, but looking at the code, `findByParent` calls the **private** `filesystemAgents()` method directly, not `retrieveFilesystemAgentConfigs`. The sync callback (`syncCallback?.sync(...)`) only lives in `retrieveFilesystemAgentConfigs`, which is exclusively called from `findDeployedByNamespaceIdAndUserIdAndName`.

```kotlin
// findByParent calls this — NO sync triggered
private fun filesystemAgents(parentId: UUID, excludeNames: Set<String> = emptySet()): List<AgentConfig> { ... }

// findDeployedByNamespaceIdAndUserIdAndName calls this — sync IS triggered
private fun retrieveFilesystemAgentConfigs(namespaceId: UUID): List<AgentConfig> { ... syncCallback?.sync(...) }
```

This means a ScheduledPrompt or UserGroup linking to a filesystem agent can only be created after someone first calls the deployed-agents endpoint. A cold start that only calls `findByParent` (e.g. the admin agent listing page) will never populate Neo4j. The sync must also be invoked from `filesystemAgents()`, or `filesystemAgents()` must be merged with `retrieveFilesystemAgentConfigs`.

---

### 🟡 Significant — `sync()` is non-transactional with 3 sequential Neo4j round-trips

`FilesystemAgentConfigSyncService.sync()` makes three Neo4j calls in sequence:
1. `neo4jRepository.findByParent(namespaceId)` — fetch existing file-origin nodes
2. `neo4jRepository.findByIds(liveIds)` — check for API-managed collision
3. N × `neo4jRepository.save(toSave)` + M × `neo4jRepository.delete(stale.metadata.id)`

There is no `@Transactional` annotation. A failure mid-way (e.g. during purge of stale nodes after some upserts succeed) leaves Neo4j in a partially-updated state — some stale nodes deleted, some not. `sync()` should be annotated `@Transactional` so the entire operation is atomic. Note: for `@Transactional` to work on a `@Service` bean, the existing `open class` declaration on `Neo4jAgentConfigRepository` is already there, but `FilesystemAgentConfigSyncService` is a plain class and needs to be `open` (or the annotation placed on an interface method) for Spring AOP to proxy it.

---

### 🟡 Significant — `@Version` carry-over has a TOCTOU window

```kotlin
val existingFileOriginById = existingFileOriginNodes.associateBy { it.metadata.id }
// ... later ...
val existingVersion = existingFileOriginById[agent.metadata.id]?.metadata?.version
val toSave = agent.copy(metadata = agent.metadata.copy(version = existingVersion))
neo4jRepository.save(toSave)
```

Between the initial `findByParent` and the eventual `save`, another concurrent sync cycle (e.g. two requests racing past the TTL boundary) could increment the version. The `save` would then throw an uncaught `OptimisticLockingFailureException` that propagates up to the original read-path caller, turning a read into a 500. This should be caught and handled gracefully (e.g. log a warning and skip the upsert for that cycle, since the concurrent write already did the work).

---

### 🟡 Significant — Bean wiring relies on undocumented declaration order

`FilesystemAgentConfigSyncService` is a `@Service` that Spring component-scans and auto-wires with `Neo4jAgentConfigRepository` (the delegate, an `open class`). `Neo4jPersistenceConfiguration` also declares both beans explicitly. Spring resolves this correctly today because `neo4jAgentConfigRepositoryDelegate` is declared before `neo4jAgentConfigRepository`, but:

- The `@Service` on `FilesystemAgentConfigSyncService` means the bean exists in two conceptual paths (component scan + explicit use in `@Bean` method), which is confusing.
- If another `Neo4jAgentConfigRepository` bean were introduced, or if bean declaration order changed, injection could silently pick the wrong bean.

Consider removing `@Service` from `FilesystemAgentConfigSyncService` and declaring it as an explicit `@Bean` in `Neo4jPersistenceConfiguration` (after the delegate bean), making the dependency chain explicit and order-independent.

---

### 🟡 Significant — `fileOrigin` flag not backfilled for existing nodes

Existing Neo4j nodes that were previously upserted by an earlier version of the code (before this PR) will have `fileOrigin = false` by default (SDN reads missing properties as the field default). On first sync after deployment, `existingFileOriginNodes` will be empty because all existing file-origin nodes still carry `fileOrigin = false`. The purge logic won't fire for them, and the upsert will skip them if they accidentally share an ID with a live filesystem agent (the `existingApiManagedIds` check). A migration or startup backfill for `fileOrigin` on existing nodes is needed, or at minimum a comment explaining why this is safe to ignore (e.g. "first deployment: no pre-existing file-origin nodes").

---

### 🟡 Minor — Duplicated KDoc on `rejectIfFilesystemBacked` in `PromptServiceImpl`

There are two consecutive `/** ... */` KDoc blocks for the same private method:

```kotlin
/**
 * Rejects [update] / [delete] when [id] resolves to a filesystem-backed prompt.
 * ...
 * defeating the collision rule documented on [FilesystemPromptRepository].
 */
/**
 * Rejects [update] / [delete] when [id] resolves to a filesystem-backed prompt.
 * ...
 * [FilesystemPromptRepository].
 */
private fun rejectIfFilesystemBacked(existing: Prompt, action: String) {
```

The first KDoc (which mentions `isFilesystemOnly`) is a leftover from the previous iteration and contradicts the second. Only the second (accurate) KDoc should remain.

---

### 🟡 Minor — Sync callback missing from `findByIds` filesystem path

`findByIds` in `FilesystemAgentConfigRepository` also calls `filesystemAgents()` when resolving IDs unknown to Neo4j. This path also bypasses the sync callback. While less critical than `findByParent` (since `findByIds` is typically called with a known ID, implying the sync already ran), it creates another inconsistency with the stated design.

---

### 🔵 Test coverage gap — `FilesystemAgentConfigSyncService` has no unit tests

The new `FilesystemAgentConfigSyncService` implements the core write logic of this feature (upsert + purge + version carry-over + collision rule), but has zero test coverage. The existing `FilesystemAgentConfigRepositoryUnitSpec` tests the `fileOrigin=true` assertion on the domain object (good) but does not exercise the sync service at all. Given this is AgentOS production-grade code, dedicated unit tests for the following scenarios are required:
- Upsert of new file-origin agents
- Skip of agents whose ID resolves to an API-managed node (`fileOrigin = false`)
- Soft-delete of stale file-origin nodes (name no longer in live set)
- Version carry-over prevents `OptimisticLockingFailureException` on second sync
- Callback invocation from `findByParent` (once the critical bug above is fixed)

PR #1270, opened 2026-09-02
feat: #34228 add WhatsNext tool to REDIRECT integration

vincent-audibert-whoz:
## What's Next Tool — WZ-34228

Adds a `WhatsNextTool` to the REDIRECT integration, enabling agents to consult a process guideline at the end of their turn and decide whether to hand off to another agent.

### How it works

When a `guideline` field is present in an `IntegrationConfig` of type `REDIRECT`, `RedirectToolPlugin.provideTools` now returns two tools instead of one:

- **`RedirectTool`** — unchanged, handles the actual agent handoff
- **`WhatsNextTool`** — takes no arguments; returns the guideline verbatim when called

The agent calls `whatsNext` when it has finished its task, receives the process guideline as the tool response, and then reasons over it using its existing `redirect` tool if a handoff is warranted. No extra LLM call inside the tool — the current agent does the reasoning itself.

The guideline is delivered at tool-call time only, keeping it out of the system prompt and preserving context efficiency.

### Configuration

Add a `guideline` field to an existing REDIRECT `IntegrationConfig`:

```json
{
  "agents": ["*"],
  "guideline": "When DemandBuilder finishes creating a talent request, redirect to TRSharing..."
}
```

No `guideline` field → behaviour identical to before (only `RedirectTool` produced). Fully backward-compatible.

### Changes

- `WhatsNextTool.kt` — new tool, no-arg, returns guideline verbatim
- `RedirectToolPlugin.kt` — produces `WhatsNextTool` when guideline is present; `guideline` field added to `configSchema`
- `WhatsNextToolSpec.kt` — unit tests for name, execute (with/without input)
- `RedirectToolPluginSpec.kt` — tests for guideline present/absent/blank, configName prefix, execute round-trip

### Last additions

- `x-ui-widget` field to indicate preference for a `textarea`

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/WhatsNextTool.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/WhatsNextTool.kt
@@ -34,9 +34,11 @@ class WhatsNextTool(
     override val name: String = configName?.let { "${it}__whatsNext" } ?: "whatsNext"
 
     override val description: String =
-        "Call this tool when you have finished helping the user with their current request. " +
-            "It returns the process guideline that tells you whether another agent should take over " +
-            "and, if so, which one. Always call this tool before ending your turn."
+        "Call this tool when you have finished the part of the work that falls within your scope. " +
+            "It returns a guideline telling you what the workflow expects next: either that another agent should " +
+            "take over (and which one), or that your step was the last one. If it names another agent, don't " +
+            "attempt that agent's work and switch to it.\n\n" +
+            "Call it once, at completion, including when the request turns out to be outside your scope and you've done what you can."
 
     override val inputSchema: String =
         """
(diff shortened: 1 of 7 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/redirect/WhatsNextTool.kt
```diff
@@ -0,0 +1,69 @@
...
+class WhatsNextTool(
+    configName: String?,
+    private val guideline: String,
+) : StandardTool<Nothing> {
+    override val name: String = configName?.let { "${it}__whatsNext" } ?: "whatsNext"
+
+    override val description: String =
+        "Call this tool when you have finished helping the user with their current request. " +
+            "It returns the process guideline that tells you whether another agent should take over " +
+            "and, if so, which one. Always call this tool before ending your turn."
```
yannick-bride-whoz: Suggestion:
Call this tool when you have finished the part of the work that falls within your scope. It returns a guideline telling you what the workflow expects next: either that another agent should take over (and which one), or that your step was the last one. If it names another agent, don't attempt that agent's work and switch to it.

Call it once, at completion, including when the request turns out to be outside your scope and you've done what you can.

yannick-bride-whoz (dismissed):

yannick-bride-whoz (approved):

PR #1272, opened 2026-09-02
refactor: WZ-34007 scheduler processes UserRuns continuously instead of chunk-by-chunk

frederic-delsert-whoz:
Replace chunk-based tick with a producer/channel/worker pool.
Producer claims batches (size=25) into a buffered channel (cap=50).
Five workers process UserRuns concurrently as soon as they are available, instead of waiting for the entire chunk to complete.
Run parent status (DONE/FAILED) updated by tickClaim scan (60s) instead of inline after each UserRun, avoiding N concurrent writes on the same runId.
Backoff on producer DB errors: min(2s * 2^(attempt-1), 60s).

```
CoroutineScope (SupervisorJob)
│
├── Producer (coroutine)        Channel              Workers (×5 coroutines)
│   ────────────────────        ───────              ───────────────────────
│
│   while(isActive)             ┌────────────┐◄──    Worker 1: for(userRun in channel) ──► createCase ──► awaitLaunch (30s)
│   claimBatch(25)              │            │◄──    Worker 2: for(userRun in channel) ──► createCase ──► awaitLaunch (30s)
│        │                      │  cap = 50  │◄──    Worker 3: for(userRun in channel) ──► createCase ──► awaitLaunch (30s)
│        └─────────────────────►│            │◄──    Worker 4: for(userRun in channel) ──► createCase ──► awaitLaunch (30s)
│                               └────────────┘◄──    Worker 5: for(userRun in channel) ──► createCase ──► awaitLaunch (30s)
│
└── cancel() ──► CancellationException propagated to all ──► channel.close() (finally)
```

Tunable via AGENTOS_PROMPT_SCHEDULER_*:
- WORKER_COUNT (default 5)
- BATCH_SIZE (default 25): items claimed per producer iteration
- CHANNEL_CAPACITY (default 50): channel buffer, ~2x batch size
- TICK_INTERVAL_MS (default 60000): claim + Run completion scan interval
- EMPTY_POLL_DELAY_MS (default 5000): producer idle delay
- PAUSED_POLL_DELAY_MS (default 10000): producer delay when consume is paused
- LEASE_MINUTES (default 30): UserRun lease duration
- LAUNCH_TIMEOUT_SECONDS (default 30): health-check window per UserRun

---

**At-least-once delivery & future effectively-once improvement**

If `markTerminal()` fails (DB unavailable) after a Case has been created, the UserRun stays `RUNNING` until its lease expires and is reclaimed by the producer, potentially creating a second Case for the same user on the next execution.

This is a known and accepted trade-off of the current at-least-once model, predating this PR.
A future improvement would store the `userRunId` on the `Case` node and check for an existing Case before creating a new one, making execution effectively-once even under partial failures.

---

**Why Channel rather than Flow?**

Only the producer would map cleanly onto a Flow; the fan-out would not, flatMapMerge is still @FlowPreview, and the recommended alternative (produceIn + N consumers) is exactly what this code already does explicitly. Three implicit behaviours would also creep in: retryWhen's attempt counter never resets (backoff would measure lifetime failures, not a consecutive-error burst), produceIn closes the channel with the terminating exception so workers die in error instead of draining cleanly, and flowOn inserts a hidden 64-element buffer that would silently override [SchedulerProperties.channelCapacity].
Worth revisiting if the polling policy ever becomes composable (prioritisation, per-namespace quotas, adaptive batching).

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
@@ -22,6 +22,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.isActive
@@ -35,7 +36,6 @@ import java.time.Clock
 import java.time.Duration
 import java.time.Instant
 import java.util.UUID
-import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Executes [ScheduledPromptRun]s in two phases.
@@ -125,13 +125,6 @@ class ScheduledPromptExecutor(
     /** When true, the producer skips claimBatch and delays instead. */
     private val consumePaused = AtomicBoolean(false)
 
-    // No synchronisation needed: start() writes scope then immediately launches the coroutine
-    // that reads it — the launch itself establishes a happens-before edge so runConsumerLoop()
-    // always sees the written value. stop() is called by Spring @PreDestroy, sequentially after
-    // start() has returned, never concurrently with it. If this invariant ever changes (e.g. an
-    // admin endpoint calling start/stop concurrently), replace with AtomicReference + compareAndSet.
-    private var scope: CoroutineScope? = null
-
     fun isConsumePaused(): Boolean = consumePaused.get()
 
     fun pauseConsume() {
@@ -143,6 +136,13 @@ class ScheduledPromptExecutor(
         consumePaused.set(false)
         logger.warn { "[Executor] consume RESUMED by operator" }
     }
+
+    // No synchronisation needed: start() writes scope then immediately launches the coroutine
+    // that reads it — the launch itself establishes a happens-before edge so runConsumerLoop()
+    // always sees the written value. stop() is called by Spring @PreDestroy, sequentially after
+    // start() has returned, never concurrently with it. If this invariant ever changes (e.g. an
+    // admin endpoint calling start/stop concurrently), replace with AtomicReference + compareAndSet.
+    private var scope: CoroutineScope? = null
 
     // -------------------------------------------------------------------------
     // Lifecycle
@@ -190,7 +190,7 @@ class ScheduledPromptExecutor(
             try {
                 while (isActive) {
                     when {
-                        consumePaused.get() -> delay(2_000L)
+                        consumePaused.get() -> delay(properties.pausedPollDelayMs)
                         else -> {
                             try {
                                 val batch = userRunRepository.claimBatch(leaseDuration, properties.batchSize)
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
@@ -125,6 +125,17 @@ class ScheduledPromptExecutor(
     /** When true, the producer skips claimBatch and delays instead. */
     private val consumePaused = AtomicBoolean(false)
 
+
+
+    /** True if the consumer loop is active. Returns false if start() was never called or if
+     * stop() was called. The watchdog uses this to detect unexpected coroutine death. */
+    fun isRunning(): Boolean = scope?.isActive == true
+
+    fun restart() {
+        stop()
+        start()
+    }
+
     fun isConsumePaused(): Boolean = consumePaused.get()
 
     fun pauseConsume() {
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
@@ -54,22 +54,27 @@ import java.util.UUID
  *
  * ### Phase B — Continuous consumption (producer + channel + worker pool)
  *
- * This bean implements [SmartLifecycle]. On [start], a [CoroutineScope] is created and two
- * structured children are launched inside it:
+ * The lifecycle is driven by `@PostConstruct` / `@PreDestroy` (Jakarta annotations).
+ * On [start], a [CoroutineScope] is created with a [SupervisorJob] and [runConsumerLoop]
+ * is launched inside it on the [dispatcher]. [stop] cancels the scope. [restart] (stop
+ * then start) is called by the watchdog [SchedulerScanner.tickWatchdog] whenever
+ * [isRunning] returns false, recovering from unexpected coroutine death.
  *
- * **Producer** (`Dispatchers.IO`): loops continuously, calling
- * [ScheduledPromptUserRunRepository.claimBatch] and sending each claimed [ScheduledPromptUserRun]
- * into the channel. When the batch is empty the producer delays [SchedulerProperties.emptyPollDelayMs]
- * before polling again, avoiding a busy-loop. On database errors it applies exponential
- * backoff (capped at 60 s). The producer respects [consumePaused]: when paused it delays
- * 2 s per iteration without touching the database.
+ * **Producer** (on the [dispatcher] injected, `Dispatchers.IO` by default): loops
+ * continuously, calling [ScheduledPromptUserRunRepository.claimBatch] and sending each
+ * claimed [ScheduledPromptUserRun] into the channel. When the batch is empty the producer
+ * delays [SchedulerProperties.emptyPollDelayMs] before polling again, avoiding a busy-loop.
+ * On database errors it applies exponential backoff (capped at 60 s). The producer respects
+ * [consumePaused]: when paused it delays [SchedulerProperties.pausedPollDelayMs] per
+ * iteration without touching the database.
  * The channel is closed in the producer's `finally` block, guaranteeing that workers
  * exit their `for (userRun in channel)` loop cleanly regardless of how the producer stops.
  *
- * **Worker pool** ([SchedulerProperties.workerCount] coroutines, `Dispatchers.IO`): each worker
- * receives [ScheduledPromptUserRun]s from the channel and calls [processUserRun] followed by
- * [checkCompletion]. A failure in one worker does not affect its siblings — exceptions are
- * caught per-item; [CancellationException] is always re-thrown to honour cooperative cancellation.
+ * **Worker pool** ([SchedulerProperties.workerCount] coroutines, on the [dispatcher]
+ * injected, `Dispatchers.IO` by default): each worker receives [ScheduledPromptUserRun]s
+ * from the channel and calls [processUserRun]. A failure in one worker does not affect its
+ * siblings — exceptions are caught per-item; [CancellationException] is always re-thrown
+ * to honour cooperative cancellation.
  *
  * **Channel capacity**: [SchedulerProperties.channelCapacity] (default 50 = 2 x batchSize).
  * Double-buffering: the producer can fill a second batch into the channel while workers drain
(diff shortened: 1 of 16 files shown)


frederic-delsert-whoz:
## Review — PR #1272: WZ-34007 Scheduler pull-model refactoring

**Title:** `refactor: scheduler processes UserRuns continuously instead of chunk-by-chunk`  
**Author:** fdelsert-whoz | **State:** Draft | +493 / -627 across 10 files

Good architectural move overall. The shift from a tick-driven batch model to a continuous **producer/channel/worker-pool** is the right direction for latency and throughput. The implementation is clean and the test suite has been carefully adapted. A few points worth addressing before merging.

---

### 🔴 Issues

**1. Channel is never closed — shutdown contract is implicit**

`runConsumerLoop()` creates the channel but never calls `channel.close()`. Workers iterate `for (userRun in channel)` which only terminates when the channel is *closed* or when `CancellationException` propagates through `channel.receive()`. The latter does happen on scope cancellation, but it relies on implicit behaviour rather than an explicit contract. If the producer exits for any reason other than cancellation (e.g. an unhandled exception escapes the `try/catch`), workers will block on `channel.receive()` forever.

**Recommended fix:** Close the channel in the producer's `finally` block, or use the `produce {}` coroutine builder which closes the channel automatically:
```kotlin
currentScope.launch(Dispatchers.IO) {
    try {
        while (isActive) { /* ... */ }
    } finally {
        channel.close()
    }
}
```

**2. `scope` is a mutable nullable var accessed from multiple coroutines without synchronisation**

```kotlin
private var scope: CoroutineScope? = null
```

`start()` writes it, `stop()` nulls it, and `runConsumerLoop()` reads it via `checkNotNull(scope)`. These can run on different threads concurrently. Consider `@Volatile` or `AtomicReference<CoroutineScope?>`.

**3. KDoc inconsistency on channel capacity**

The class-level KDoc states:
> **Channel capacity**: `workerCount`. At most one item waiting per worker…

But `runConsumerLoop()` uses `properties.channelCapacity` (default **50**), and its own KDoc says "capacity is `workerCount * 2`". The actual intent (2 × batchSize for double-buffering) is correctly documented in `SchedulerProperties` but the class-level doc is wrong and internally inconsistent.

---

### 🟡 Design observations

**4. Run completion latency: near-immediate → up to 60 s**

The old model checked `checkCompletion` inline after each group settled. The new model delegates to `recoverOrphanedRunningRuns` on `tickClaim` (every 60 s by default). This is a deliberate and documented trade-off to avoid N concurrent writes on the same `runId` — good reasoning. Worth confirming downstream consumers of Run status (dashboards, webhooks, monitoring alerts) tolerate a ~60 s lag.

**5. `processUserRun` is `internal` for testability**

Pragmatic choice, clearly motivated in the KDoc. Adding a `// @VisibleForTesting` comment would make the intent explicit and signal this is not part of the public contract of the bean.

---

### 🟢 Positives

- `exponentialBackoffMs` is clean, well-tested, and correctly handles edge cases (`attempt ≤ 0`, overflow guarded by `coerceIn(0, 29)`).
- Removing `tickConsume` from `SchedulerScanner` simplifies the scanner's single responsibility significantly.
- Moving `consumePaused` from `SchedulerScanner` to `ScheduledPromptExecutor` is the right SRP fix — the executor owns the loop, it should own the pause flag.
- Tests are now deterministic without coroutine infrastructure by calling `processUserRun` directly — good call.
- The two-`tickClaim` pattern in scenario tests accurately mirrors the production flow.
- Removing the `kotlinx.coroutines.delay(500)` hacks from unit tests is a meaningful quality improvement.

---

**Summary:** The three 🔴 points (channel never closed, `scope` thread-safety, KDoc inconsistency) should be addressed before merging. Items 4 and 5 are informational. Overall a solid refactoring that simplifies the architecture and improves throughput.

frederic-delsert-whoz:
All 3 issues raised by Coday have been addressed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -94,8 +126,135 @@ class ScheduledPromptExecutor(
...
+                                logger.error(e) {
+                                    "[Executor] producer error (attempt=$consecutiveErrors), backing off ${backoffMs}ms"
+                                }
+                                delay(backoffMs)
+                            }
+                        }
+                    }
+                }
+            } finally {
+                channel.close()
```
vincent-couturier-whoz, line 242: question: why does the consumer close the channel ? logically it should be closed by the producer

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -94,8 +126,135 @@ class ScheduledPromptExecutor(
...
+    private suspend fun runConsumerLoop() {
+        val channel = Channel<ScheduledPromptUserRun>(capacity = properties.channelCapacity)
+        val currentScope = checkNotNull(scope)
+
+        // Producer
+        currentScope.launch(dispatcher) {
+            val leaseDuration = Duration.ofMinutes(properties.leaseMinutes)
+            var consecutiveErrors = 0
+            try {
+                while (isActive) {
```
vincent-couturier-whoz, line 217: question: what's the origine of isActive ? if it's the scope isActive seems to be the same as true...

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -94,8 +126,135 @@ class ScheduledPromptExecutor(
...
+     * The channel is closed in the producer's `finally` block, guaranteeing that workers
+     * iterating `for (userRun in channel)` exit their loop cleanly whether the producer
+     * stops normally, is cancelled, or throws an unexpected exception.
+     *
+     * The producer and all workers share the same scope. Cancelling the scope (on [stop])
+     * propagates [CancellationException] to all of them at their next suspension point.
+     * The producer's `finally` block then closes the channel, which unblocks any worker
+     * suspended on `channel.receive()`.
+     */
+    private suspend fun runConsumerLoop() {
```
vincent-couturier-whoz: suggestion: we could clarify the naming because between the channel consumer and run consumer it's difficult to understand which manages what

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -94,8 +126,135 @@ class ScheduledPromptExecutor(
...
+                    }
+                }
+            } finally {
+                channel.close()
+            }
+        }
+
+        // Worker pool — Run (parent) completion is NOT checked here.
+        // SchedulerScanner.recoverOrphanedRunningRuns handles RUNNING → DONE/FAILED on each tickClaim.
+        repeat(properties.workerCount) { workerId ->
```
vincent-couturier-whoz, line 254: suggestion: I would create two methods. One for the pooling in database which produce the userRun in the channel and another one which manage the workers. For example an initUserRunProcessingProducer and initUserRunProcessingWorkers

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
```diff
@@ -122,14 +111,28 @@ class SchedulerScanner(
      * When [claimPaused] the tick is a no-op — the scheduling thread still fires
      * but [processClaim] is not called.
      */
-    @Scheduled(fixedDelayString = "\${agentos.prompt.scheduler.tick-interval-ms:30000}")
+    @Scheduled(fixedDelayString = "\${agentos.prompt.scheduler.tick-interval-ms:60000}")
```
vincent-couturier-whoz: question: why this change ?

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
frederic-delsert-whoz: The while (`isActive`) loop is here to ensure the producer stops cleanly when the Spring bean is destroyed.
When `@PreDestroy` calls `scope.cancel()`, the scope's Job transitions to Cancelling, which immediately sets isActive to false, causing the loop to exit at its next iteration and guaranteeing the` finally { channel.close() } `block runs to unblock all workers.

!isActive -> cancellation has been initiated

We store a `CoroutineScope` rather than just the `SupervisorJob` for convenience: it lets us call `launch()` directly and use scope.isActive as a natural liveness check.
Note that `scope.cancel()` is just a shortcut for `scope.coroutineContext[Job]?.cancel()`, cancellation always goes through the Job. Same for isActive :
```kotlin
public val CoroutineScope.isActive: Boolean
    get() = coroutineContext[Job]?.isActive ?: true
```

All this is done to minimise shutdown time.
Also cancellation propagates through the Job to all coroutines, which stop cooperatively at their next suspension point rather than being killed abruptly.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
frederic-delsert-whoz: The naming was confusing here.
The poller claims UserRuns from the database and sends them into the channel.
The workers are the consumers.
So the channel is indeed closed by the producer, in its finally block, which is the correct pattern.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
frederic-delsert-whoz: done at commits b61b3f3f58e21d2f305850fefe09a28980f86ed1 and 23e2f0003c05f363486d8d9516fed53ade01fa46

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
frederic-delsert-whoz: done at commits https://github.com/whoz-oss/coday/commit/b61b3f3f58e21d2f305850fefe09a28980f86ed1 and https://github.com/whoz-oss/coday/commit/23e2f0003c05f363486d8d9516fed53ade01fa46

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/SchedulerScanner.kt
frederic-delsert-whoz: The default in the annotation was changed to 60s but it doesn't matter in practice, the `application.yml` explicitly sets it to 30s, which always takes precedence. The 30s value was intentionally chosen to align with Coday's tick interval.
Rolled back to 30s for consistency at commit: 5d8634c2af8400f95cbfb44e63e29b47600e51b7

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -94,8 +127,150 @@ class ScheduledPromptExecutor(
...
+            }
+        }
+    }
+
+    /**
+     * Processes claimed UserRuns — one worker per configured [SchedulerProperties.workerCount].
+     * Run (parent) completion is NOT checked here:
+     * [SchedulerScanner.recoverOrphanedRunningRuns] handles RUNNING -> DONE/FAILED on each tickClaim.
+     */
+    private fun startUserRunWorkers(channel: Channel<ScheduledPromptUserRun>) {
```
vincent-couturier-whoz, line 252: suggestion: use ReceiveChannel to ensure there's no send in the method

vincent-couturier-whoz (approved):

PR #1273, opened 2026-09-02
fix: #1271 Resume multiple answered questions sequentially

benjamin-valdes-whoz:

PR #1277, opened 2026-09-02
feat: #1275 filesystem skill catalog discovery and prompt injection

### LINKED ISSUE #1275: Add filesystem skills with per-agent skill selectors compatible with Claude's Agent Skills convention
## Context

AgentOS needs native filesystem skills compatible with Claude's Agent Skills convention, while allowing specialized Coday agents to limit catalog size and avoid irrelevant skill routing.

`Namespace.configPath` points to the project's `coday` directory, so AgentOS discovers skills under `<configPath>/skills/{name}/SKILL.md` (flat layout).

---

## Goal

Provide Claude-compatible progressive-disclosure skills in AgentOS as a first-class entity, plus optional per-agent catalog filtering through Coday-specific `skillSelectors`.

Agents receive:

1. Lightweight skill metadata (name + description) at initialization
2. Full `SKILL.md` body only when relevant, via a dedicated `readSkill(name)` tool
3. Referenced resources only when needed, via `readSkillResource(name, path)`

---

## Skill contract

Each skill lives in its own directory directly under `<configPath>/skills/{name}/` and uses a file named exactly `SKILL.md`.

Requirements:

- Filename must be exactly `SKILL.md`
- `name` must be nonblank
- `description` must be nonblank
- YAML frontmatter must parse successfully
- Additional frontmatter fields are ignored
- Malformed skills are skipped with diagnostics
- Duplicate names are resolved deterministically (first by path wins)
- Canonical path validation rejects files escaping the skills root
- Only metadata enters initial agent instructions
- Body and adjacent resources load on demand through dedicated tools
- Bundled scripts are never exposed or executed

---

### Level 1: discovery metadata

AgentOS injects a compact catalog containing name and description only — no filesystem paths:

```markdown
## Available Skills

- **spec-writing**: Writes, reviews, or checks ticket specifications.
```

### Level 2: skill instructions

When a task matches a skill description, the agent calls `readSkill(name)` to retrieve the body.

### Level 3: resources

The agent calls `readSkillResource(name, path)` for adjacent references or resources only when the skill body requires them. Scripts are never registered or executed.

---

## Per-agent skill selection

Optional `skillSelectors` on agent definitions.

### Selector semantics

| Configuration | Result |
|---|---|
| Field absent, `null`, or `[]` | No skills (capability grants are opt-in) |
| `["*"]` | All discovered skills |
| `["spec-writing"]` | Exact skill name match (case-insensitive) |

Folder/glob patterns (e.g. `core/**`) are NOT supported and never match. Unknown selectors emit a warning without blocking agent startup. Matching is case-insensitive and deduplicated.

---

## Architecture

Entity-first, Neo4j-primary with filesystem augmentation:

- `Skill` entity + `SkillNode` — doubleKey uniqueness per namespace/platform scope, soft-delete via tombstoned doubleKeys
- `FilesystemSkillRepository` — decorator over the Neo4j repository; reads merge persisted and discovered skills (persisted wins on name collision), writes forward to Neo4j only
- `SkillService` / `SkillServiceImpl` — selector filtering, namespace-over-platform shadowing, read-only guard for filesystem-backed skills (`metadata.version == null`)
- `SkillCatalogRenderer` — renders the instructions catalog block
- `SkillToolGrantService` — builds `readSkill` / `readSkillResource` bound to the run's resolved catalog, granted iff non-empty
- `SkillController` — REST API at `/api/skills` (CRUD + by-parentId + platform); `SkillApi`/`SkillDto` SDK contract
- `EntityType.SKILL` — wired into namespace-transitive and platform-scoped permissions; `SkillAuthorizationService` fallback authorizes filesystem-backed skills against their owning namespace
- `AgentServiceImpl` — resolves the catalog once per agent run, feeds instructions and tool grant

`skillSelectors` propagates through `AgentConfig`, `AgentConfigDto`, `AgentConfigNode` (as `skillSelectorsJson`), `FilesystemAgentConfigRepository`, `AgentConfigController`, and agent YAML import/export.

---

## Security

- Symlink rejection via `NOFOLLOW_LINKS` plus canonical-path containment re-validated on every file read (entry files and resources)
- Sensitive-file detection on resolved targets (shared `SensitiveFileDetector`)
- Bounds: depth-2 discovery walk, max 500 skills, 512 KiB per SKILL.md, 1 MiB per resource, 5-minute filesystem cache TTL

---

### Discovery

```gherkin
Given valid SKILL.md files under `<configPath>/skills/{name}/`
When AgentOS resolves an agent with skill selectors
Then each valid skill is discovered
And its name and description are available for catalog generation
```

### Invalid metadata

```gherkin
Given a SKILL.md with malformed YAML or missing name or description
When discovery runs
Then the invalid skill is skipped with a diagnostic
And remaining valid skills still load
```

### Progressive disclosure

```gherkin
Given a valid discovered skill
When agent instructions are built
Then only skill name and description are injected
And the full SKILL.md body is not injected
```

```gherkin
Given a user request matching a selected skill description
When the agent handles the request
Then the agent calls readSkill with the skill name
And follows the loaded instructions
```

### Opt-in default

```gherkin
Given an agent has no skillSelectors field
When the agent is resolved
Then no skill catalog is injected
And no skill tools are granted
```

### Wildcard selection

```gherkin
Given an agent has `skillSelectors: ["*"]`
When the agent is resolved
Then all discovered skills are included
```

### Name selection

```gherkin
Given an agent has `skillSelectors: ["spec-writing"]`
When the agent is resolved
Then spec-writing is included
And unrelated skills are excluded
```

### Unknown selectors

```gherkin
Given a selector matches no discovered skill
When filtering runs
Then a warning is emitted
And agent startup continues
```

### Path security

```gherkin
Given a discovered file resolves outside the canonical skills root
When discovery runs
Then the escaped file is rejected
```

### Script governance

```gherkin
Given a skill directory contains scripts
When discovery runs
Then scripts are not exposed or executed
And existing AgentOS tool permissions remain authoritative
```

---

## Out of scope

- TypeScript legacy runtime implementation
- Automatic script registration or execution
- Skill management UI
- Skill packaging or remote installation
- Live filesystem watching
- Folder/glob selector patterns

AgentOS remains the canonical implementation.

---

## Follow-up

`docs` remains a separate always-loaded mechanism.

After skill migration:

- Remove obsolete skill indexes from agent `docs`
- Remove direct skill documents from always-loaded `docs`
- Preserve only documents genuinely required for every turn
- Track legacy `mandatoryDocs` cleanup separately where TypeScript compatibility still matters

Next steps: full entity treatment (resources lifecycle, skill execution, unload, metrics) without touching agent-side code.

leo-punsola-whoz:
Introduces Claude-compatible filesystem skills as a first-class AgentOS entity, with
per-agent opt-in catalog selection and dedicated read tools.

## Model

Skills are `SKILL.md` files (YAML frontmatter `name` + `description`, markdown body)
in a flat layout under `<configPath>/skills/{name}/`, with adjacent files loaded as
resources. Storage is Neo4j-primary with filesystem augmentation:

- `Skill` entity + `SkillNode` (doubleKey uniqueness per namespace/platform scope,
  soft-delete via tombstoned doubleKeys)
- `FilesystemSkillRepository` decorates the Neo4j repository: reads merge persisted
  and discovered skills (persisted wins on name collision), writes forward to Neo4j only
- `SkillService` / `SkillServiceImpl`: selector filtering, namespace-over-platform
  shadowing, read-only guard for filesystem-backed skills (`metadata.version == null`)
- `SkillController` at `/api/skills` (CRUD + by-parentId + platform), `SkillApi`/`SkillDto`
  SDK contract, `EntityType.SKILL` wired into namespace-transitive and platform-scoped
  permissions, `SkillAuthorizationService` fallback for filesystem-backed reads

## Agent exposure

- `AgentConfig.skillSelectors`: absent or empty = **no skills** (capability grants are
  opt-in, aligned with `subAgents`); `["*"]` = all; otherwise exact name match
  (case-insensitive). Folder/glob patterns are not supported.
- `SkillCatalogRenderer` injects a name/description-only catalog into instructions —
  no filesystem paths.
- `SkillToolGrantService` grants `readSkill(name)` and `readSkillResource(name, path)`
  bound to the run's resolved catalog, iff non-empty. Bodies and resources load on
  demand; scripts are never executed.

## Security

- Symlink rejection via `NOFOLLOW_LINKS` + canonical-path containment re-validated on
  every file read (entry files and resources)
- Sensitive-file detection on resolved targets (shared `SensitiveFileDetector`,
  extracted from `AgentDocumentResolver`)
- Bounds: depth 2 walk, 500 skills, 512 KiB per SKILL.md, 1 MiB per resource,
  5-minute filesystem cache TTL

## Also in this PR

- `FilesystemYamlCache` gains optional `filePredicate` and `maxDepth` (defaults
  preserve existing behavior for all current callers)
- `subAgents` entries are now trimmed like other list fields (silent behavior fix)

## Test plan

- New suites: skill discovery/parsing, SkillServiceImpl, FilesystemSkillRepository,
  SkillController, SkillAuthorizationService
- Extended: AgentServiceImplUnitSpec (skills block injection, selector forwarding,
  null/empty semantics), FilesystemAgentConfigRepositoryUnitSpec (skillSelectors parsing)
- CI green: validate-jvm, validate-ts

Closes #1275


agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/agentConfig/AgentConfigDto.kt
```diff
@@ -61,4 +61,16 @@ data class AgentConfigDto(
             ),
     )
     val subAgents: List<String>? = null,
+    @ArraySchema(
+        schema =
+            Schema(
+                minLength = 1,
```
la-celton-whoz, line 67: minLength me fait penser que ca peut pas etre null non ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
```diff
@@ -116,6 +116,79 @@ data class AgentConfig(
...
+     *   scalars are resolved by the YAML parser before collapsing, so `description: >` still
+     *   works correctly.
+     * - Discovery walks at most 10 directory levels below the `skills` root and collects at most
+     *   500 `SKILL.md` files per namespace; files exceeding 256 KiB are skipped with a WARN.
+     * - The discovered catalog is cached per namespace for 60 seconds. Skill authors see edits
+     *   reflected within a minute; steady-state agent traffic pays one filesystem walk per
+     *   namespace per minute.
+     *
+     * Tri-state, inverse of [integrations]:
+     * - null: all discovered skills are advertised (default)
```
vincent-audibert-whoz: `null` = all is a nogo, it was the behavior on Coday typescript, handy in the beginning but just a plain design error in the long run. `null` should amount to "no skills", use `"*"` to explicitly expose all skills.

Also, this comment is 10 times too detailed to document this property.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+import java.nio.file.NoSuchFileException
+import java.nio.file.NotDirectoryException
+import java.nio.file.Path
+import java.time.Duration
+import java.time.Instant
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.io.path.name
+import kotlin.io.path.pathString
+
+data class Skill(
```
vincent-audibert-whoz: extract in its own class

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+        const val SKILLS_SUBDIR = "skills"
+        const val SKILL_FILE_NAME = "SKILL.md"
+
+        // Max bytes read per SKILL.md before skipping (256 KiB).
+        // Prevents a single large accidental commit from allocating fully in memory.
+        const val MAX_SKILL_FILE_BYTES = 256 * 1024L
+
+        // Max walk depth below the skills root.
+        // Prevents runaway traversal of deeply nested directory trees.
+        const val MAX_WALK_DEPTH = 10
```
vincent-audibert-whoz: Seems overkill, 3 or 4 at most ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+        // retained strings each — a few hundred KB at most — so the unbounded map is not a
+        // concern in practice. A deleted namespace leaves a stale entry until process restart.
+        //
+        // No stampede guard: discoverSkills is called on every resolveAgentDefinition, i.e. every
+        // agent invocation. Concurrent callers on a cold or just-expired key all pass the freshness
+        // check and all run computeSkills, producing N redundant filesystem walks where one would
+        // suffice. This is accepted because each walk is bounded by MAX_WALK_DEPTH and
+        // MAX_SKILL_COUNT and runs on Dispatchers.IO. A per-key Mutex with a double-check inside
+        // is the fix if this is ever measured to matter.
+        val CACHE_TTL: Duration = Duration.ofSeconds(60)
```
vincent-audibert-whoz: There is already a file system cache utility service to handle that, can it be used or incorporate the needed changes to avoid duplicating config file caching ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
```diff
@@ -201,6 +201,14 @@ class FilesystemAgentConfigRepository(
                             .normalize()
                             .toString() + suffix
                     }?.takeIf { it.isNotEmpty() },
+            // An explicit empty list (`skillSelectors: []`) is a deliberate opt-out and must be
```
vincent-audibert-whoz: Actually, `null` = no skills = `[]` is a preferable heuristic.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+ * introduce a new privilege boundary — anyone who can write `<configPath>/skills/` already holds
+ * write access to the agent YAML files under `<configPath>/agents/`, which carry raw instruction
+ * text. Both are
+ * administrator-level namespace configuration. See [io.whozoss.agentos.agentConfig.AgentConfig.skillSelectors]
+ * for the full trust and upgrade-semantics contract.
+ */
+@Component
+class SkillResolver {
+
+    private val yamlMapper: ObjectMapper =
```
vincent-audibert-whoz: See `ObjectMapperConfiguration.kt` to use the existing mapper (with its `@Qualifier`).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+                return@withContext emptyList()
+            }
+
+            // Warn once if the catalog was truncated.
+            if (totalSeen > MAX_SKILL_COUNT) {
+                logger.warn { "[SkillResolver] Skill catalog truncated to $MAX_SKILL_COUNT entries; $totalSeen SKILL.md files were found under $skillsRoot" }
+            }
+
+            // Derive projectRoot from the already-normalized absolute configPath.
+            val projectRoot = configPath.parent ?: configPath
```
vincent-audibert-whoz: Why back one step from the configPath ?
The configPath is arbitrarily positioned, could be at the root of a coday-project corresponding to an agentos-namespace, deep in it or even elsewhere.
The core assumption to keep is that the `namespace.configPath` points to a folder where we have `./agents`, `./integrations`, `./prompts` and now `./skills`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
```diff
@@ -0,0 +1,396 @@
...
+    ): String? {
+        val allSkills = discoverSkills(namespaceConfigPath)
+        val filteredSkills = filterSkills(allSkills, skillSelectors)
+        if (filteredSkills.isEmpty()) return null
+
+        return buildString {
+            appendLine()
+            appendLine("## Available Skills")
+            appendLine()
+            appendLine("You have access to the following domain skills. When a task matches a skill's description, read its `SKILL.md` using file tools to retrieve instructions and follow its guidelines on demand.")
```
vincent-audibert-whoz: Warning: Relying on file read for skill load implies:
- the agent has file tools
- file tools cover the path to the `skills` folder

This will work currently on most of our internal agents, but that is a fragile build-up as not all agents will have access to the `{configPath}/skills` folder.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/security/declarative/PreAuthorizeLabelConsistencySpec.kt
```diff
@@ -0,0 +1,140 @@
...
+ * (e.g. `'AgentConfig'`, `'Namespace'`), NOT the Kotlin enum constant name (`'AGENT_CONFIG'`).
+ * Actions must be one of the [Action] entries: `'READ'`, `'WRITE'`, `'DELETE'`.
+ *
+ * ## Scope
+ *
+ * Only `hasPermission(...)` calls are validated. Other SpEL expressions (`isAuthenticated()`,
+ * `hasRole(...)`, `permitAll()`) do not pass through [AgentOsPermissionEvaluator] and are not
+ * subject to the same silent-failure mode, so they are intentionally ignored.
+ */
+class PreAuthorizeLabelConsistencySpec :
```
vincent-audibert-whoz, line 43: This is very very nice 😍 , could be in its own PR 👏

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/agentConfig/AgentConfigDto.kt
vincent-audibert-whoz: minLength should be 0 to allow for empty list, no ?

vincent-audibert-whoz (dismissed):
I have mixed feelings about this PR:

😍 nice test on PreAuthorize
👍 correct attachment to `AgentConfig`
🪄  will indeed work out of the box for skills on our internal usage and with agent setup constraints

💥  brittle setup and implicit dependencies for skills to work on agents
📉  starts fleshing out skills for agents but it should be an entity type alike integration, hence it is already drifting from patterns and integration with the rest

In the end, skills should be entities in AgentOS managed in a similar way as integrations:
- hybrid storage with primary neo4j and secondary files with the de-facto composite repository pattern.
- loading (and executing and unloading?) through dedicated tools (no reliance on file integration existence), allowing skill usage monitoring.

Implementing from file-first approach did bring some powerful bias in Coday, the database-entity-first approach seems much cleaner => what are your plans for the next steps ?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/agentConfig/AgentConfigDto.kt
leo-punsola-whoz: minLength = 1 s’applique uniquement aux valeurs présentes dans la liste : chaque sélecteur doit contenir au moins un caractère. Ça ne rend pas skillSelectors obligatoire ni non-null. Le champ reste nullable, et une liste vide reste autorisée puisqu’on ne définit pas minItems.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfig.kt
leo-punsola-whoz: Addressed: null and [] now expose no skills, while ["*"] explicitly exposes all discovered skills. Property KDoc was reduced to selector semantics and essential loading behavio

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: MAX_WALK_DEPTH is now 4. This accepts skill directories at depths 0–3, with SKILL.md occupying file-walk depth 4; deeper trees are not traversed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: Skill was extracted into its own Skill.kt domain entity. Discovery, persistence, catalog rendering, and on-demand reading are now separated into dedicated components.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: removed custom ConcurrentHashMap caching and reused FilesystemYamlCacheRegistry. Shared utility now supports configurable file predicates and walk depth, allowing skill discovery to cache SKILL.md files without duplicating filesystem-cache logic.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
leo-punsola-whoz: Addressed: filesystem parsing now normalizes missing, empty, and blank-only skillSelectors to null. Both null and [] therefore mean no skills; ["*"] remains explicit opt-in for all skills.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: skill filesystem parsing now injects shared ObjectMapper with @Qualifier("yamlMapper"); inline mapper construction was removed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: discovery now resolves skillsRoot directly as <namespace.configPath>/skills. No parent or project-root assumption remains; configPath is treated as folder containing agents, integrations, prompts, and skills.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillResolver.kt
leo-punsola-whoz: Addressed: skill loading no longer relies on generic file tools or their configured roots. Agents receive dedicated readSkill(name) and readSkillResource(name, path) tools bound to their resolved skill catalog. Catalog exposes names only, and skill bodies load directly through dedicated tool.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillController.kt
```diff
@@ -0,0 +1,146 @@
...
+                    namespaceId = resource.namespaceId,
+                    name = resource.name,
+                    description = resource.description,
+                    body = resource.body,
+                )
+            },
+        )
+
+    @GetMapping("/{id}")
+    @PreAuthorize("hasPermission(#id, 'Skill', 'READ')")
```
selimbensenouciep-prog, line 74: `EntityType.SKILL` is missing from `isNamespaceChildEntity` and `PLATFORM_SCOPABLE_ENTITY_TYPES` in `Neo4jPermissionRepository`. In an integration test with Neo4j, a namespace admin can create and list a skill, but GET by ID returns 404 and PUT returns 403. An authenticated user also gets 404 when fetching a platform skill that appears in the platform list.

Could you wire skills into the namespace inheritance and platform-read rules, and cover these flows with non-superadmin integration tests?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
```diff
@@ -0,0 +1,373 @@
...
+        val description = collapseWhitespace(model.description ?: "").truncate(MAX_SKILL_DESCRIPTION_CHARS)
+
+        if (name.isBlank() || description.isBlank()) {
+            logger.debug { "[FilesystemSkillRepository] Skipping $file: blank name or description" }
+            return null
+        }
+
+        return Skill(
+            metadata = EntityMetadata(
+                id = UUID.nameUUIDFromBytes("filesystem-skill:$name".toByteArray(Charsets.UTF_8)),
```
selimbensenouciep-prog, line 246: The UUID is derived only from the skill name, so two namespaces exposing a skill with the same name receive the same ID. `findByIds` scans all namespaces and keeps the first match. I reproduced a lookup returning the first namespace's body when using the ID advertised by the second namespace.

Could you include the namespace ID in the filesystem skill identity and add a regression test where two namespaces have the same skill name but different bodies?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/plugin/filesystem/FilesystemYamlCache.kt
```diff
@@ -62,11 +75,11 @@ class FilesystemYamlCache<T>(
 
         val items =
             Files
-                .walk(directory)
+                .walk(directory, maxDepth)
```
vincent-audibert-whoz, line 78: Will enable arbitrary file organisation inside the `agents|prompts|integrations|skills` folders ❤️

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Skill.kt
```diff
@@ -0,0 +1,43 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class Skill(
+    override val metadata: EntityMetadata = EntityMetadata(),
+    val namespaceId: UUID? = null,
+    val name: String,
+    val description: String,
+    val body: String,
+    /** Relative path of the skill directory under the skills root. Null for DB-persisted skills. */
+    val skillRelativePath: String? = null,
```
vincent-audibert-whoz: Optional note: I am surprised to see these path surface here in the domain, I wonder if it could not stay in the file system cache class as a map ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Reads a file adjacent to a skill, resolved under [Skill.resourceRoot].
+ *
+ * Enforces path containment (no traversal) and rejects sensitive filenames.
+ * Returns an error when [Skill.resourceRoot] is null (DB-stored skills have no
+ * bundled resources in step 1).
+ */
+class SkillReadResourceTool(
```
vincent-audibert-whoz, line 66: So this tool will at first only be used for file-based skills, that seems a weird assymetry, how do you see the future of skills:
- one body and some scripts
- possibly several resources (but what about the risk of scope creep of the skill ?)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        if (!Files.isRegularFile(resolved)) {
+            return ToolExecutionResult.error("'${input.path}' is not a regular file.", errorType = "NOT_A_FILE")
+        }
+
+        val probe = (MAX_RESOURCE_BYTES + 1).coerceIn(1L, Int.MAX_VALUE.toLong()).toInt()
+        val bytes =
+            try {
+                Files.newInputStream(resolved).use { it.readNBytes(probe) }
```
vincent-audibert-whoz: This is duplication with the file plugin capability to read various file types, we will need a refactoring afterwards (maybe have raw file reading lib in agentos-sdk?).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+     * Selectors support:
+     * - Wildcard: a single star entry matches all available skills.
+     * - Recursive folder prefix: prefix ending with slash-star-star matches the prefix itself
+     *   and all recursive descendants (paths starting with prefix slash).
+     * - Single-level folder prefix: prefix ending with slash-star matches only direct child skills
+     *   under the prefix directory (remainder after prefix slash contains no further slash segments).
+     * - Exact match: exact relative path, path ending with slash SKILL.md, or frontmatter name (case-insensitive).
+     * - DB-persisted skills with null skillRelativePath only match wildcard or exact name.
+     */
+    internal fun filterSkills(
```
vincent-audibert-whoz, line 110: Same as what was done on integrations, we should avoid loading the whole entities for the scope to filter them afterwards. That implies some specific queries (and files will be loaded in full, but cache is there to mitigate the impact) => issue or ticket on that 🙏

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,194 @@
...
+        skills: List<Skill>,
+        selectors: List<String>,
+    ): List<Skill> {
+        if (selectors.isEmpty()) return emptyList()
+        if (selectors.contains("*")) return skills
+        val matched = LinkedHashSet<Skill>()
+        for (selector in selectors) {
+            val normalized = selector.trim()
+            val beforeCount = matched.size
+            when {
```
vincent-audibert-whoz: The complexity here would make me reconsider having an arbritrary folder structure in `skills` 🥲 . Is this a feat absolutely necessary, or could we make do with a flat structure like `[...]/skills/{skill name}/SKILL.md` ?

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptBatchScenarioSpec.kt
vincent-audibert-whoz, line 1: This has nothing to do with the skills, missing rebase on or merge with master ?

vincent-audibert-whoz (dismissed):
⚠️  Please wait end of week for merge ⚠️

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptBatchScenarioSpec.kt
vincent-audibert-whoz: Another simpler fix is proposed here: https://github.com/whoz-oss/coday/pull/1273/changes#r3913393667

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillController.kt
leo-punsola-whoz: Done: EntityType.SKILL added to isNamespaceChildEntity; new SkillControllerPermissionIntegrationSpec covers admin/member/platform access matrix against embedded Neo4j.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Skill.kt
leo-punsola-whoz: Done: skillRelativePath and resourceRoot removed from domain entirely; resources are pre-loaded into an in-memory map at discovery time, so no path surfaces in the domain.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
leo-punsola-whoz: Asymmetry resolved: resources now live on the entity (resourcesJson on SkillNode), so DB-stored skills carry resources too. Scope creep risk acknowledged — resources stay read-only text, never executed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
leo-punsola-whoz: Partially done: sensitive-file detection extracted to shared SensitiveFileDetector in agentos-sdk. Broader raw-file-reading lib refactor agreed as follow-up — will track in a dedicated issue.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Done for the named-selector path: findSkills short-circuits * and otherwise uses findByNamespaceIdAndNames batch lookup instead of loading the full catalog. Full scope-filtered queries remain a follow-up issue.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Done: flat skills/{name}/SKILL.md enforced; selector matching is now pure name equality, glob logic deleted.

agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptBatchScenarioSpec.kt
leo-punsola-whoz: Reverted — fix from #1273 applies instead; no longer part of this branch.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
```diff
@@ -0,0 +1,350 @@
...
+                stream
+                    .filter { Files.isRegularFile(it) && it.fileName.toString() != SKILL_FILE_NAME }
+                    .filter { !isJunkOrBinaryResource(skillDir.relativize(it)) }
+                    .forEach { resourceFile ->
+                        val fileName = resourceFile.fileName.toString()
+                        if (!SkillReadResourceTool.isSensitiveFile(fileName)) {
+                            val relPath = skillDir.relativize(resourceFile).toString().replace("\\", "/")
+                            try {
+                                if (Files.size(resourceFile) <= SkillReadResourceTool.MAX_RESOURCE_BYTES) {
+                                    resources[relPath] = Files.readString(resourceFile)
```
selimbensenouciep-prog, line 269: The `SKILL.md` path is checked for containment, but resource files are not. I reproduced this with a `reference.txt` symlink pointing outside the skill directory: its target's contents were loaded into `Skill.resources`. Checking only the link's filename also lets an innocuous name bypass the sensitive-file filter, exposing the content through the skill API and `readSkillResource`.

Could you validate each resource's canonical path against the skill directory before reading it, or reject symlinks, and add a regression test for this case?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    override fun findByNamespaceIdAndNames(
+        namespaceId: UUID,
+        names: Collection<String>,
+    ): List<Skill> {
+        if (names.isEmpty()) return emptyList()
+        val nameSet = names.mapTo(HashSet()) { it.lowercase() }
+        val persisted = delegate.findByNamespaceIdAndNames(namespaceId, names)
+        val persistedNames = persisted.mapTo(HashSet()) { it.name.lowercase() }
```
selimbensenouciep-prog, line 84: `findByNamespaceIdAndNames` also returns platform skills, so `persistedNames` currently excludes a namespace's filesystem skill whenever a platform skill has the same name. With both versions present, selecting `["review"]` returns the platform version, while `["*"]` correctly returns the namespace version.

Could you only exclude filesystem skills when the matching persisted skill belongs to the current namespace, and cover this collision through the filesystem repository?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillController.kt
```diff
@@ -0,0 +1,151 @@
...
+                    name = resource.name,
+                    description = resource.description,
+                    body = resource.body,
+                    resources = resource.resources ?: emptyMap(),
+                )
+            },
+        )
+
+    @GetMapping("/{id}")
+    @PreAuthorize("hasPermission(#id, 'Skill', 'READ')")
```
selimbensenouciep-prog, line 74: Filesystem-backed skills have no Neo4j node, so this permission check still rejects their IDs. With a non-superadmin namespace admin, I reproduced a successful namespace listing containing the skill, followed by a 404 from `GET /api/skills/{id}` and an empty list from `POST /api/skills/by-ids`.

Could you authorize filesystem-backed skills against their owning namespace in both lookup paths, and cover these calls with a real filesystem skill in the permission integration tests?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return skills.filter { it in matched }
+    }
+
+    // -------------------------------------------------------------------------
+    // Helper methods
+    // -------------------------------------------------------------------------
+
+    private fun isFilesystemBacked(skill: Skill): Boolean {
+        val persisted = skillRepository.findByIds(listOf(skill.metadata.id), withRemoved = false)
```
selimbensenouciep-prog, line 135: `skillRepository` is the filesystem decorator, so `findByIds` also returns filesystem-backed skills. That makes `isFilesystemBacked` return false for them and bypasses the read-only guard. I reproduced a superadmin PUT returning 200 and creating a persisted copy that takes precedence over the file.

Could you use a reliable way to distinguish filesystem and persisted skills, and add tests ensuring filesystem skills cannot be updated or deleted through the API?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/agentConfig/AgentConfigDto.kt
```diff
@@ -61,4 +61,16 @@ data class AgentConfigDto(
...
+                implementation = String::class,
+                description =
+                    "Selectors controlling which skills are advertised to this agent. " +
+                        "Null or empty means no skills. Use ['*'] for all, 'core/**' for a folder prefix, " +
```
selimbensenouciep-prog, line 71: This description still advertises folder selectors such as `core/**` and relative paths, but the runtime now matches skill names only. The same outdated examples remain in the OpenAPI spec and `docs/study/agentos-skills.md`, including nested skill directories that discovery no longer visits. Following these examples can leave an agent with an empty catalog, without a warning.

Could you align the SDK description, generated spec and guide with the flat `skills/{name}/SKILL.md` layout and the supported selectors: skill names, `*`, and `null` or `[]` for no skills?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
leo-punsola-whoz: Fixed: loadSkillResources now rejects symlinks (NOFOLLOW_LINKS) and validates each resource's real path against the skill directory's real path before reading. Sensitive-file check runs on the resolved target. Regression test added in FilesystemSkillRepositoryUnitSpec.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
leo-punsola-whoz: Fixed: persistedNames exclusion now filtered to namespaceId == it.namespaceId, so a platform skill no longer shadows a namespace's own filesystem skill. Collision covered via filesystem repository test.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillController.kt
leo-punsola-whoz: "Fixed: new SkillAuthorizationService.canRead resolves any skill (persisted or filesystem) and authorizes against the owning namespace's READ permission; platform skills (namespaceId == null) are readable by all, unknown ids fail closed. Wired into getById and getByIds; permission integration spec extended with a real filesystem skill.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Fixed: filesystem-vs-persisted distinction now uses metadata.version == null (idiom from AgentConfig.isFilesystemOnly) instead of re-querying the composite repository. Tests added asserting filesystem skills cannot be updated or deleted through the API.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/agentConfig/AgentConfigDto.kt
leo-punsola-whoz: Fixed: AgentConfigDto description, regenerated agentos-openapi.yaml, and agentos-skills.md now document the flat skills/{name}/SKILL.md layout with supported selectors only — skill names, *, and null/[] for no skills. Agent YAML configs updated to match.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/util/SensitiveFileDetector.kt
```diff
@@ -0,0 +1,41 @@
+package io.whozoss.agentos.sdk.util
+
+/**
+ * Shared security utility for detecting sensitive file names (secrets, credentials, private keys).
+ *
+ * Used across plugins, file tools, and skill resource decoders.
+ */
+object SensitiveFileDetector {
+
+    val DEFAULT_SENSITIVE_PATTERNS: List<String> = listOf(
```
vincent-couturier-whoz: suggestion: We have already io.whozoss.agentos.plugins.file.SensitiveFilePatterns which could/should be reuse or merge with this file to avoid multiple source of truth about sensitive files

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
```diff
@@ -144,7 +145,8 @@ class AgentConfigController(
                     advancedExecution = resource.advancedExecution ?: false,
                     externalMetadata = resource.externalMetadata,
                     enabled = resource.enabled ?: existing.enabled,
-                    subAgents = resource.subAgents?.filter { it.isNotBlank() }?.takeIf { it.isNotEmpty() },
+                    subAgents = resource.subAgents?.filter { it.isNotBlank() }?.map { it.trim() }?.takeIf { it.isNotEmpty() },
```
vincent-couturier-whoz: suggestion: have a common function instead of a copy/past on two consecutive lines + two other consecutive lines

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/util/SensitiveFileDetector.kt
vincent-couturier-whoz, line 1: praise: good refactoring to have only one source of truth ❤️

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/util/SensitiveFileDetector.kt
vincent-couturier-whoz, line 10: suggestion: you could also use this constant in BoundaryPathResolver and remove the second list we had with exactly the same name

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
```diff
@@ -201,6 +201,11 @@ class FilesystemAgentConfigRepository(
                             .normalize()
                             .toString() + suffix
                     }?.takeIf { it.isNotEmpty() },
+            skillSelectors =
+                model.skillSelectors
+                    ?.filter { it.isNotBlank() }
```
vincent-couturier-whoz: suggestion: use the same function as the one we should have created to have null or a list of not blank items (something like `model.skillSelectors.nullOrNotBlankItems()`)

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
```diff
@@ -288,6 +293,39 @@ class Neo4jPersistenceConfiguration {
...
+    fun neo4jSkillRepositoryDelegate(
+        skillNodeNeo4jRepository: SkillNodeNeo4jRepository,
+        childLinkService: Neo4jChildLinkService,
+    ): Neo4jSkillRepository = Neo4jSkillRepository(skillNodeNeo4jRepository, childLinkService)
+
+    @Bean
+    @Primary
+    fun neo4jSkillRepository(
+        neo4jSkillRepositoryDelegate: Neo4jSkillRepository,
+        namespaceRepository: NamespaceRepository,
```
vincent-couturier-whoz, line 318: question: why did you mixed repositories ? we should avoid this and use the service layer to avoid mixing entities at the repository level.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }.toDomain()
+            .also {
+                logger.debug {
+                    "[Neo4jSkillRepository] Saved skill ${it.id} ('${entity.name}') " +
+                        "scope=(namespaceId=${entity.namespaceId})"
+                }
+            }
+
+    override fun findByIds(
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+        ids: Collection<UUID>,
+        withRemoved: Boolean,
+    ): List<Skill> =
+        neo4jRepository
+            .findAllById(ids.map { it.toString() })
+            .filter { withRemoved || it.removed != true }
+            .map { it.toDomain() }
+
+    override fun findByParent(parentId: UUID): List<Skill> = findByNamespaceId(parentId)
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    override fun findByNamespaceId(namespaceId: UUID): List<Skill> =
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+        neo4jRepository
+            .findActiveByNamespaceId(namespaceId.toString())
+            .map { it.toDomain() }
+
+    override fun findPlatform(): List<Skill> =
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+        neo4jRepository
+            .findActivePlatform()
+            .map { it.toDomain() }
+
+    override fun findByNameInNamespace(
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+        namespaceId: UUID?,
+        name: String,
+    ): Skill? =
+        neo4jRepository
+            .findActiveByDoubleKey(SkillNode.computeDoubleKey(namespaceId, name))
+            ?.toDomain()
+
+    override fun findByNamespaceIdAndNames(
```
vincent-couturier-whoz: suggestion: add transactional readonly

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
```diff
@@ -101,6 +102,15 @@ class Neo4jSchemaInitializer(
             logger.info { "[Neo4jSchemaInitializer] Constraint $constraintName ensured" }
         }
 
+        // Skill doubleKey uniqueness constraint: enforces uniqueness on (namespaceId, lowercased name)
```
vincent-couturier-whoz, line 105: issue: add also index on namespaceId because there's neo4j node matching by namespaceId

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
vincent-couturier-whoz, line 1: suggestion: I think there's too much responsibility in this class. It could be clarified by moving in another class the processing of one skill file from this class

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillCatalogRenderer.kt
```diff
@@ -0,0 +1,35 @@
...
+            appendLine("## Available Skills")
+            appendLine()
+            appendLine(
+                "You have access to the following domain skills. " +
+                    "When a task matches a skill's description, call readSkill(\"<name>\") " +
+                    "to retrieve its instructions and follow its guidelines on demand.",
+            )
+            appendLine()
+            skills.forEach { skill ->
+                appendLine("- **${skill.name}**: ${skill.description}")
```
vincent-couturier-whoz, line 26: suggestion: we should perhaps enforce the skill name rules to avoid bad surprises here

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+interface SkillNodeNeo4jRepository : Neo4jRepository<SkillNode, String> {
+    /**
+     * Find all non-removed namespace-scoped skills, ordered by name ASC.
+     *
+     * Traverses the BELONGS_TO edge and filters by the Namespace id.
+     */
+    @Query(
+        $$"""
+            MATCH (s:Skill)-[r:BELONGS_TO]->(ns:Namespace)
```
vincent-couturier-whoz: suggestion: we should decide to filter on a property or relationship  because the queries below use property matching to find skills. It would be better to be consistent and use property matching. Furthermore we should avoid having relationship automatically translated because we will retrieve the namespace even if we don't use it.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
```diff
@@ -0,0 +1,140 @@
...
+            return ToolExecutionResult.error("A resource path is required.", errorType = "MISSING_INPUT")
+        }
+
+        val skill = skills.firstOrNull { it.name.equals(input.name, ignoreCase = true) }
+            ?: return ToolExecutionResult.error(
+                "Skill '${input.name}' not found.",
+                errorType = "NOT_FOUND",
+            )
+
+        val normalizedPath = input.path.trim().trimStart('/').replace("\\", "/")
```
vincent-couturier-whoz: suggestion: have a function to always apply the same pattern everywhere we have a path to a file.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillRepository.kt
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * Uniqueness is enforced per level (namespace or platform).
+     */
+    fun findByNameInNamespace(
+        namespaceId: UUID?,
+        name: String,
+    ): Skill?
+
+    /**
+     * Find skills matching any of [names] (case-insensitive) in [namespaceId] or platform scope.
```
vincent-couturier-whoz: question: with removed or not ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,184 @@
...
+    ): Skill? = withContext(Dispatchers.IO) {
+        skillRepository.findByNameInNamespace(namespaceId, name)
+            ?: skillRepository.findByNameInNamespace(null, name)
+    }
+
+    private suspend fun loadEffectiveSkills(namespaceId: UUID): List<Skill> = withContext(Dispatchers.IO) {
+        val namespaceSkills = skillRepository.findByNamespaceId(namespaceId)
+        val platformSkills = skillRepository.findPlatform()
+
+        val namespaceNames = namespaceSkills.mapTo(HashSet()) { it.name.lowercase() }
```
vincent-couturier-whoz: suggestion: namespaceSkillNames to simplifying reading and avoid ambiguity

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,184 @@
...
+        if (selectors.isNullOrEmpty()) return@withContext emptyList()
+        if (selectors.contains("*")) {
+            loadEffectiveSkills(namespaceId)
+        } else {
+            val names = selectors.map { it.trim().removeSuffix("/SKILL.md").removeSuffix("/").trimStart('/') }
+            val matched = skillRepository.findByNamespaceIdAndNames(namespaceId, names)
+            // Apply case-insensitive shadowing: namespace skills shadow platform skills of same name
+            val namespaceSkills = matched.filter { it.namespaceId != null }
+            val platformSkills = matched.filter { it.namespaceId == null }
+            val namespaceNames = namespaceSkills.mapTo(HashSet()) { it.name.lowercase() }
```
vincent-couturier-whoz: suggestion: namespaceSkillNames

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,184 @@
...
+     * - Exact skill name (case-insensitive).
+     */
+    internal fun filterSkills(
+        skills: List<Skill>,
+        selectors: List<String>,
+    ): List<Skill> {
+        if (selectors.isEmpty()) return emptyList()
+        if (selectors.contains("*")) return skills
+        val matched = LinkedHashSet<Skill>()
+        for (selector in selectors) {
```
vincent-couturier-whoz, line 119: suggestion: not sure to understand this code as matched is not modified explicitly. Perhaps using filterTo but it's not really clear. Not sure why we have a mutable set instead of just set built to a simple mapNotNull on selectors

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,184 @@
...
+     * bypassed.
+     *
+     * The reliable discriminator is [io.whozoss.agentos.sdk.entity.EntityMetadata.version]:
+     * filesystem skills are built in-memory on every read and never go through Spring Data
+     * Neo4j's `save()`, so `metadata.version` — set to a non-null value by SDN on first
+     * persistence — stays null for their entire lifetime. This mirrors the same idiom used by
+     * [io.whozoss.agentos.agentConfig.AgentConfig.isFilesystemOnly] and
+     * [io.whozoss.agentos.prompt.PromptServiceImpl.rejectIfFilesystemBacked].
+     */
+    private fun isFilesystemBacked(skill: Skill): Boolean = skill.metadata.version == null
```
vincent-couturier-whoz, line 161: question: how can we ensure that version is not provided in the skill.md file ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
```diff
@@ -0,0 +1,184 @@
...
+
+    private fun requireUniqueName(
```
vincent-couturier-whoz, line 163: question: why and excludeId ?

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/AgentConfigController.kt
leo-punsola-whoz: Done: extracted shared nullOrNotBlankItems() in StringUtils and applied across controller mapping and DTO conversion.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/util/SensitiveFileDetector.kt
leo-punsola-whoz: Done: BoundaryPathResolver now references SensitiveFileDetector.DEFAULT_SENSITIVE_PATTERNS as the single source of truth.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentConfig/FilesystemAgentConfigRepository.kt
leo-punsola-whoz: Done: using nullOrNotBlankItems() on skillSelectors and subAgents parsing in FilesystemAgentConfigRepository.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
leo-punsola-whoz: The FilesystemSkillRepository is a decorator over the Neo4j repository (composite repository pattern). It needs the namespace configPath to locate the filesystem skills directory during repository lookups. This follows the exact same pattern already established for FilesystemAgentConfigRepository, FilesystemPromptRepository, and FilesystemIntegrationConfigRepository

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
leo-punsola-whoz: Done: added @Transactional(readOnly = true) to all read methods in Neo4jSkillRep

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jSchemaInitializer.kt
leo-punsola-whoz: Done: added CREATE INDEX skill_namespace_id IF NOT EXISTS FOR (s:Skill) ON (s.namespaceId) in Neo4jSchemaInitializer.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
leo-punsola-whoz: Done: added @transactional(readOnly = true) to all read methods in Neo4jSkillRep

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
leo-punsola-whoz: Done: added @transactional(readOnly = true) to all read methods in Neo4jSkillRep

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/Neo4jSkillRepository.kt
leo-punsola-whoz: Done: added @transactional(readOnly = true) to all read methods in Neo4jSkillRep

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/FilesystemSkillRepository.kt
leo-punsola-whoz: Done: extracted SkillFileParser to encapsulate parsing, frontmatter extraction, resource loading, containment validation, and name rules.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillCatalogRenderer.kt
leo-punsola-whoz: Done: added SkillFileParser.isValidSkillName() validating charset (alphanumeric, dot, dash, underscore, space) and length <= 100. Enforced in SkillFileParser (filesystem skip) and SkillServiceImpl.create/update (400 Bad Request).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillNodeNeo4jRepository.kt
leo-punsola-whoz: Done: aligned all queries in SkillNodeNeo4jRepository to match by scalar namespaceId property, avoiding unnecessary relationship traversal and Namespace loading. Removed unused namespace relationship field from SkillNode.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillReadTool.kt
leo-punsola-whoz: Done: extracted normalizeRelativePath in StringUtils and used in SkillReadResourceTool and SkillFileParser.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillRepository.kt
leo-punsola-whoz: Only non-removed skills. The underlying Cypher query filters with (s.removed IS NULL OR s.removed = false).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Done: renamed variable to namespaceSkillNames for clarity.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Done: refactored filterSkills to iterate over selectors cleanly and collect matches into a list preserving discovery order.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: Done: renamed variable to namespaceSkillNames for clarity.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: The filesystem parser explicitly extracts only name and description from the YAML frontmatter into a dedicated SkillFrontmatter DTO. The version property lives on EntityMetadata, which is generated in-memory with version = null. It is impossible for a SKILL.md file to inject a version.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/skill/SkillServiceImpl.kt
leo-punsola-whoz: excludeId allows update() to save an entity with its existing name without triggering a self-conflict during the uniqueness check.

vincent-couturier-whoz (approved):

PR #1282, opened 2026-09-04
feat(agentos): #WZ-34091 restrict scheduler execution to off-peak time windows

frederic-delsert-whoz:
**Context**

The scheduler was running continuously 24/7, competing with live user traffic for LLM capacity. The goal is to restrict execution to configurable time windows so LLM resources are preserved for live users during business hours.

**What was implemented**

Time-window guard on both phases of the scheduler, implemented via environment variable — without relying on the existing start/pause mechanism (`SchedulerEndpoint`). The window is evaluated dynamically from config + current time on every tick, including after a restart (no persisted state). When no windows are configured, the scheduler runs continuously (existing behaviour preserved).

 **New `ExecutionWindowService`** (`@Service`): receives `SchedulerProperties` and a `Clock` (default `Clock.systemUTC()`) injected by Spring, parses each `DAYOFWEEK HH:mm UTC` boundary into a weekly minute offset at startup (`init` block), groups them pairwise into open/close execution windows, and evaluates whether the current `Instant` falls within any of them via `isWithinExecutionWindow()`. The `Clock` injection makes the service fully testable without mocking. Handles wrap-around windows (e.g. `FRIDAY 22:00 → MONDAY 05:00`). Validates parity, boundary syntax, zero-length windows, ordering/overlap, and allows at most one wrap-around window. Fails open on any invalid config so a misconfiguration never silently halts all scheduled executions.

**New config property** `agentos.prompt.scheduler.windows` (`AGENTOS_PROMPT_SCHEDULER_WINDOWS`), bound as a `List<String>` — Spring splits the comma-separated value, so the whole schedule stays in a single environment variable:

```
MONDAY 22:00,FRIDAY 05:00,FRIDAY 22:00,MONDAY 05:00
```

Boundaries alternate open/close: the example above defines two execution windows — `Mon 22:00 → Fri 05:00` and `Fri 22:00 → Mon 05:00`. When unset (empty list), the scheduler runs continuously.

The open/close role is carried by position in the list rather than by a structured type — a deliberate trade-off to keep the configuration in a single environment variable (structured YAML binding would require one env var per boundary in production).

**Behaviour:**

- **Phase A** (`tickClaim`): outside the window, the tick is a no-op — due prompts accumulate in DB and are processed at the next window open, including after a restart
- **Phase B** (DB poller in `ScheduledPromptExecutor`): the poller skips `claimBatch()` and delays when `consumePaused || !isWithinExecutionWindow()`. Workers drain whatever is already buffered in the channel (at most `channelCapacity` UserRuns, default 50), then block on `receive()` — bounded and deliberate (no lease expiry risk, no duplicate execution)

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
@@ -1,0 +1,227 @@
+package io.whozoss.agentos.scheduledPrompt
+
+import mu.KLogging
+import java.time.DayOfWeek
+import java.time.LocalTime
+import java.time.ZoneOffset
+import java.time.ZonedDateTime
+
+/**
+ * Evaluates whether the current time falls within a configured execution window.
+ *
+ * ### Configuration format
+ *
+ * The window list is specified as a comma-separated string of `DayOfWeek HH:mm` pairs.
+ * Pairs are ordered as alternating open/close boundaries: open₁, close₁, open₂, close₂, …
+ *
+ * Each boundary is a `java.time.DayOfWeek` name (case-insensitive) followed by a space
+ * and an `HH:mm` time in UTC.
+ *
+ * Example — nightly Mon–Thu + continuous weekend:
+ * ```
+ * MONDAY 22:00,FRIDAY 05:00,FRIDAY 22:00,MONDAY 05:00
+ * ```
+ * This defines two windows:
+ * - Window 1: Monday 22:00 UTC → Friday 05:00 UTC
+ * - Window 2: Friday 22:00 UTC → Monday 05:00 UTC
+ *
+ * Windows are expressed as weekly offsets (minutes since Monday 00:00 UTC).
+ * A window that crosses the Sunday→Monday boundary is handled via modular arithmetic.
+ *
+ * ### No windows configured
+ *
+ * When [SchedulerProperties.windows] is null or blank, [isWithinWindow] always returns `true`
+ * — the scheduler runs continuously, preserving the existing behaviour.
+ *
+ * ### Validation
+ *
+ * [parseAndValidate] is called at construction time. If the string is malformed, all
+ * errors are collected and logged; [isWithinWindow] then always returns `true` (fail-open)
+ * so that a misconfiguration does not silently halt all scheduled executions.
+ *
+ * Validation rules:
+ * - Even number of entries (open/close pairs).
+ * - Each entry matches `DayOfWeek HH:mm` with a valid day name and time.
+ * - Windows do not overlap (close must be strictly after open in weekly offset).
+ * - Windows are ordered (each open must be strictly after the previous close).
+ */
+class ExecutionWindowService(windows: String?) {
+
+    /**
+     * A parsed window boundary: minutes elapsed since Monday 00:00 UTC within the week.
+     * Range: [0, 7×24×60) = [0, 10080).
+     */
+    private data class WeeklyBoundary(val minuteOfWeek: Int)
+
+    /**
+     * A validated open/close window pair, both expressed as [WeeklyBoundary].
+     * A window where [close] < [open] wraps around the Sunday→Monday boundary.
+     */
+    private data class Window(val open: WeeklyBoundary, val close: WeeklyBoundary)
+
+    /**
+     * Parsed and validated windows.
+     * - `null`  → no windows configured (always-open) OR config was invalid (fail-open).
+     * - non-null list → validated windows to evaluate.
+     */
+    private val parsedWindows: List<Window>?
+
+    init {
+        parsedWindows = parseAndValidate(windows)
+    }
+
+    /**
+     * Returns `true` when [now] falls within any configured execution window.
+     *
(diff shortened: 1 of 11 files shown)


frederic-delsert-whoz:
## Code Review — PR #1282: `feat: wz-34091 restrict scheduler execution to off-peak time windows`

The feature is well-scoped: a clean `ExecutionWindowService` utility, a minimal touch on `SchedulerScanner`, good test coverage across boundary conditions, wrap-around, and invalid configs. Four issues need addressing before merge.

---

### 1. Code-style violation — early `return` in `tickClaim()` (`SchedulerScanner.kt`)

```kotlin
when {
    claimPaused.get() -> logger.debug { "... PAUSED ..." }
    !checkExecutionWindow() -> return          // ← early return mid-method
    else -> processClaim()
}
```

The AgentOS code-style rule explicitly states: "Avoid early returns in the middle of a method". The `when` block currently mixes three different exit semantics (log + fall-through, explicit `return`, and `processClaim()`). Restructure so there is a single exit point at the end of the method.

---

### 2. KDoc inconsistency — `parseAndValidate` (`ExecutionWindowService.kt`, line ~110)

```kotlin
 * Returns an empty list on parse failure (fail-open — [isWithinWindow] returns true).
```

The implementation actually returns `null` on parse failure, not an empty list. The `null` return is the correct choice (it is what `isWithinWindow` checks against), but the KDoc is wrong and will mislead future maintainers.

---

### 3. Ordering validation gap for overlapping wrap-around windows (`ExecutionWindowService.kt`, `parseAndValidate`)

The overlap check is:
```kotlin
if (w.open.minuteOfWeek <= prev.close.minuteOfWeek) { errors += ... }
```

This compares raw `minuteOfWeek` integers. It correctly rejects non-wrap-around overlaps. However, two wrap-around windows (both with `close < open`, e.g. two `SUNDAY xx:00 → MONDAY yy:00` entries) could overlap each other without being detected, because the raw `close` value of the first is small and the `open` of the second is large — the comparison passes even though the windows overlap in the Sunday→Monday zone. Minor edge case in practice, but a correctness gap in the validator.

---

### 4. Misleading UTC label in log message (`SchedulerScanner.kt`, `checkExecutionWindow`)

```kotlin
logger.info { "... (${now.toLocalTime()} UTC)" }
```

`now` is constructed from the injected `clock` whose zone is not guaranteed to be UTC. In tests `clock` is fixed to a UTC instant but nothing enforces this in production. If the clock zone is ever non-UTC, `toLocalTime()` returns the wall-clock time in that zone while the label says "UTC" — misleading. The time should be converted to UTC before extracting the local time for the log string.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
```diff
@@ -0,0 +1,234 @@
...
+        parsedWindows = parseAndValidate(windows)
+    }
+
+    /**
+     * Returns `true` when [now] falls within any configured execution window.
+     *
+     * Returns `true` unconditionally when no windows are configured (always-open behaviour)
+     * or when configuration parsing failed (fail-open).
+     */
+    fun isWithinWindow(now: ZonedDateTime = ZonedDateTime.now(ZoneOffset.UTC)): Boolean {
```
vincent-couturier-whoz: question: Why do you use ZonedDateTime instead of Instant which is by design a UTC datetime ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        // Validate ordering: each open must be strictly after the previous close.
+        // Skipped when a wrap-around pair was already reported — raw offset comparison is
+        // meaningless in that case and would emit a confusing secondary error.
+        if (wrapAroundWindows.size <= 1) {
+            for (i in 1 until windows.size) {
+                val w = windows[i]
+                val prev = windows[i - 1]
+                if (w.open == w.close || prev.open == prev.close) continue // already reported above
```
vincent-couturier-whoz: suggestion: use collection operator to have a more functional programming style and so avoid bug due to wrong index or wrong continue effect.

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
frederic-delsert-whoz: done at commit: ea707d391163b5e9965de6b080a24f79a8b62088

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ExecutionWindowService.kt
frederic-delsert-whoz: done at commit: 94ea36bdddcbf7f74030beb555e014889e65c3aa

vincent-couturier-whoz (approved):

nicolas-remy-whoz (approved):

PR #1289, opened 2026-09-07
feat: add AGENTOS internal tool plugin for AgentConfig CRUD

vincent-audibert-whoz:
Screenshot 2026-09-09 at 22 52 23

## Summary

Introduces the `AGENTOS` integration type — a config-less internal `ToolPlugin` that lets a privileged agent manage other agents in its namespace.

## Tools exposed

| Tool | Permission required |
|---|---|
| `ListAgents` | Namespace READ |
| `GetAgent` | AgentConfig READ |
| `CreateAgent` | Namespace WRITE |
| `UpdateAgent` | AgentConfig WRITE |
| `EnableAgent` | AgentConfig WRITE |
| `DisableAgent` | AgentConfig WRITE |

## Design decisions

**Config-less**: No `IntegrationConfig` parameters — the schema is an empty object. All access control is handled by `PermissionService`.

**Service injection**: Possibly later benefits from the permissions.

**Permission model**: Each lambda calls `PermissionService.hasPermission()` with the same entity type and action as the corresponding `@PreAuthorize` on `AgentConfigController`. No logic is duplicated — both paths call the same underlying service.

**Fail-closed**: `userId == null` (anonymous/system call) is denied on all tools without consulting the permission graph.

**Mass-assignment guard on update**: `UpdateAgent` only applies non-null fields; `namespaceId` is never touched.

## Files

- `agentosPlugin/AgentosToolPlugin.kt` — `ToolPlugin` implementation
- `agentosPlugin/AgentosPluginConfiguration.kt` — Spring `@Configuration` with permission-enforcing lambdas
- `agentosPlugin/ListAgentsTool.kt`
- `agentosPlugin/GetAgentTool.kt`
- `agentosPlugin/CreateAgentTool.kt`
- `agentosPlugin/UpdateAgentTool.kt`
- `agentosPlugin/EnableAgentTool.kt`
- `agentosPlugin/DisableAgentTool.kt`
- `AgentosToolPluginUnitSpec.kt` — plugin unit tests
- `AgentosPluginConfigurationUnitSpec.kt` — configuration lambda tests (permission denied/granted per operation)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosPluginConfiguration.kt
```diff
@@ -0,0 +1,193 @@
...
+ * - **Get** : AgentConfig READ (checked on the id of the found entity)
+ * - **Create** : Namespace WRITE
+ * - **Update / Enable / Disable** : AgentConfig WRITE (checked on the id of the found entity)
+ *
+ * All lambdas are fail-closed: `userId == null` → deny. Anonymous / system calls have
+ * no user identity to check against the permission graph.
+ */
+@Configuration
+class AgentosPluginConfiguration(
+    private val agentConfigRepository: AgentConfigRepository,
```
vincent-audibert-whoz: See to use the AgentConfigService rather than the repository (to benefit later from security layer at service ?).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+@Configuration
+class AgentosAgentsPluginConfiguration(
+    private val agentConfigService: AgentConfigService,
+    private val permissionService: PermissionService,
+) {
+    @Bean
+    fun agentosAgentsToolPlugin(): ToolPlugin =
+        AgentosAgentsToolPlugin(
+            listAgents = { namespaceId, userId, withDisabled ->
+                if (userId == null) {
```
vincent-couturier-whoz: suggestion : both check should/could be exported in a dedicated method with a method name in order to have a consistent check

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+                        Action.READ,
+                    )
+                ) {
+                    logger.debug { "[AgentosPlugin] listAgents denied: user $userId lacks READ on namespace $namespaceId" }
+                    return@AgentosAgentsToolPlugin null
+                }
+                agentConfigService.findByNamespace(namespaceId, withDisabled)
+            },
+            getAgent = { namespaceId, userId, name ->
+                if (userId == null) {
```
vincent-couturier-whoz: suggestion: we should check read access to the namespace to be consistent

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+                        Action.READ,
+                    )
+                ) {
+                    logger.debug { "[AgentosPlugin] getAgent denied: user $userId lacks READ on agent ${agent.id}" }
+                    return@AgentosAgentsToolPlugin null
+                }
+                agent
+            },
+            createAgent = { namespaceId, userId, input ->
+                if (userId == null) {
```
vincent-couturier-whoz: suggestion: permission should be extracted

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+                        modelName = input.modelName ?: agent.modelName,
+                        integrations = input.integrations ?: agent.integrations,
+                        subAgents =
+                            input.subAgents?.filter { it.isNotBlank() }?.takeIf { it.isNotEmpty() }
+                                ?: agent.subAgents,
+                        advancedExecution = input.advancedExecution ?: agent.advancedExecution,
+                    ),
+                )
+            },
+            enableAgent = { namespaceId, userId, name ->
```
vincent-couturier-whoz: suggestion: we should add also ability to deploy to a userGroup or the namespace or ideally only to the user himself because enable is not enough to take advantage of the agent

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+ */
+@Configuration
+class AgentosAgentsPluginConfiguration(
+    private val agentConfigService: AgentConfigService,
+    private val permissionService: PermissionService,
+) {
+    @Bean
+    fun agentosAgentsToolPlugin(): ToolPlugin =
+        AgentosAgentsToolPlugin(
+            listAgents = { namespaceId, userId, withDisabled ->
```
vincent-couturier-whoz: suggestion: We could have a class for each tool to avoid changing the overall class we have to change just one tool implementation

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsToolPlugin.kt
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Exposed tools:
+ * - [ListAgentsTool] — list agents in the namespace (Namespace READ)
+ * - [GetAgentTool] — retrieve a single agent by name (AgentConfig READ)
+ * - [CreateAgentTool] — create a new agent (Namespace WRITE)
+ * - [UpdateAgentTool] — update an existing agent (AgentConfig WRITE)
+ * - [EnableAgentTool] — enable an agent (AgentConfig WRITE)
+ * - [DisableAgentTool] — disable an agent (AgentConfig WRITE)
+ */
+class AgentosAgentsToolPlugin(
```
vincent-couturier-whoz, line 36: question : the description mention a Spring Managed bean so why we don't use a simple `@Service` annotation here ? Or perhaps we could refer here to the AgentosAgentsPluginConfiguration

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/CreateAgentTool.kt
```diff
@@ -0,0 +1,113 @@
...
+                    },
+                    "instructions": {
+                        "type": "string",
+                        "description": "System-level instructions for the agent's behaviour."
+                    },
+                    "modelName": {
+                        "type": "string",
+                        "description": "AI model name or alias to use. If omitted, the namespace default applies."
+                    },
+                    "integrations": {
```
vincent-couturier-whoz, line 87: suggestion: we should provide a list models and list integrations tool to let another agent create an agent easily. Models are easy to specify but integrations are hardest to define cleanly using a prompt.

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
```diff
@@ -0,0 +1,199 @@
...
+                        userId.toString(),
+                        EntityType.NAMESPACE,
+                        namespaceId.toString(),
+                        Action.READ,
+                    )
+                ) {
+                    logger.debug { "[AgentosPlugin] listAgents denied: user $userId lacks READ on namespace $namespaceId" }
+                    return@AgentosAgentsToolPlugin null
+                }
+                agentConfigService.findByNamespace(namespaceId, withDisabled)
```
thomas-martin-whoz: un when serait plus lisible que ces return@

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsToolPlugin.kt
vincent-audibert-whoz: You were right to ask — there was no reason. The `@Configuration` + lambdas pattern had been copied from `CasePluginConfiguration` / `RedirectConfiguration`, where the indirection exists to break a real Spring cycle. Here no such cycle exists: `AgentConfigService` depends only on `AgentConfigRepository`, `PromptRepository` and `UserService`, none of which reach `ToolRegistryService`. The KDoc even admitted it.

`AgentosAgentsToolPlugin` is now a plain `@Service` with a single constructor dependency, collected natively by Spring into `ToolRegistryService`'s `List<ToolPlugin>`. `AgentosAgentsPluginConfiguration` is deleted — 199 lines removed. Verified: no cycle after also injecting `NamespaceService` into the collaborator, Spring context tests are green.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
vincent-audibert-whoz: We implemented this, then reverted it — it turns out to be an access **restriction**, not a consistency fix. Writing up the reasoning since it is not visible from `PermissionService` alone.

The semantics live one level below, in `Neo4jPermissionRepository`: `AGENT_CONFIG` is in `isNamespaceChildEntity(...)` and is **not** in `OWNER_PRIVATE_ENTITY_TYPES` (which only contains `CASE`). So for a READ on an `AgentConfig`, `hasTransitivePermission` calls `hasReadAccessViaNamespace`, with this comment in the code: « Shared entities (AgentConfig, IntegrationConfig, AiProvider, AiModel, Prompt): namespace MEMBERs legitimately inherit READ through the namespace (FR21, FR27, FR32, FR35) ».

In other words `hasPermission(user, AGENT_CONFIG, id, READ)` already grants access on **direct relation on the entity OR transitivity through the namespace**. Checking `NAMESPACE READ` first turns that OR into an AND, and denies a user who holds a direct `MEMBER`/`ADMIN` edge on the `AgentConfig` without being a namespace member.

The asymmetry with `listAgents` is therefore deliberate, not accidental: `listAgents` enumerates at namespace scope and has no single entity id to check against, so `NAMESPACE READ` is the only option there. A `getAgent` KDoc now states this so the suggestion is not re-applied later.

Leaving this thread open in case you disagree — happy to revisit.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
vincent-audibert-whoz: Done in `7b71c3df` / `d4b34a3e`. The guards are now four named private helpers on `AgentAdminOperationsImpl` — `canReadNamespace`, `canWriteNamespace`, `canReadAgent`, `canWriteAgent` — each taking an `operation` label so the debug logs keep their context. The repeated agent lookup (`findByNamespace(…).firstOrNull { name.equals(…, ignoreCase = true) }`), which was duplicated identically in four places, is now a single `findAgentByName` helper.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
vincent-audibert-whoz: Agreed on the problem, with a nuance on where it sat: each tool was already its own class — what was monolithic was the *configuration*, which held six lambdas carrying the actual logic. Changing one tool meant editing that shared class.

`AgentosAgentsPluginConfiguration` is now deleted (see the sibling thread on `AgentosAgentsToolPlugin`). The logic moved to a named collaborator, `AgentAdminOperations` / `AgentAdminOperationsImpl`, and each tool class depends on that interface and calls one named method. Adding or changing a tool no longer touches a shared configuration class.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
vincent-audibert-whoz: Done — same change as the sibling thread above. All permission checks now go through the four named helpers on `AgentAdminOperationsImpl`, so the control model is readable per operation instead of being inlined six times.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/AgentosAgentsPluginConfiguration.kt
vincent-audibert-whoz: You were right, and this was the most important point of the review — it was a functional gap, not a nice-to-have. Confirmed in `AgentConfigNodeNeo4jRepository.findDeployedByNamespaceIdAndUserId`: runtime visibility requires a `DEPLOYED_TO` edge to a `Namespace` or a `UserGroup`. `enabled = true` is not sufficient, except for a super-admin and only on namespace-scoped agents. So `CreateAgent` + `EnableAgent` was producing an agent nobody could use.

Addressed in `5082a370` with a new `SetAgentDeployment` tool, wired to the existing `NamespaceService.deployAgents` / `undeployAgents` (which already validate namespace existence and agent membership, including the rule that a platform agent can be deployed on any namespace). Permission is `NAMESPACE WRITE`, aligned with `NamespaceController.deployAgents` — deployment changes what the namespace exposes to its members, so it is a namespace administration decision.

Two parts of your suggestion are deliberately left out of this PR:

- **UserGroup deployment**: `UserGroupRepository.addAgents` exists, but exposing it also means exposing group discovery, which is a different subject.
- **Deploying to the user only**: there is no primitive for this — the graph only knows `Namespace` and `UserGroup` as `DEPLOYED_TO` targets. It would require a data model change.

Both will go into a follow-up issue.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agentosPlugin/CreateAgentTool.kt
vincent-audibert-whoz: Agreed, and your framing is the right one: models can reasonably be stated in a prompt, integrations cannot — without discovery the `integrations` field of `CreateAgent` is guesswork for an LLM. `IntegrationTypeRegistry.listTypes()` already exposes what is needed, so the cost is low.

Keeping it out of this PR to avoid stretching it further, and grouping it into a follow-up issue with two related gaps found while implementing deployment:

- no way to **read** deployment state — `GetAgent` cannot report whether an agent is deployed, since `findDeployedByNamespaceIdAndUserId` filters a list rather than answering the question for one agent. Writing without reading is an asymmetry worth closing.
- UserGroup deployment and per-user deployment, from the deployment thread.

All three are facets of the same goal: making the plugin usable by an LLM with no prior knowledge of the data model. Will link the issue here.

PR #1296, opened 2026-09-07
fix: update countUnread to use max(MessageEvent.timestamp) instead of c.modified

### LINKED ISSUE #1295: countUnread uses c.modified instead of max(MessageEvent.timestamp)
## Bug: `countUnread` uses `Case.modified` instead of last message timestamp

**Describe the bug**

The `countUnread` Cypher query determines unread status by comparing `Case.modified > readAt`, but `Case.modified` is only updated when case metadata changes (title, status, etc.) — not when new messages are added.

This causes the unread count to be systematically under-reported: cases with new messages are not counted as unread because `Case.modified` hasn't changed since the case was last opened.

**Expected behavior**

A case should be counted as unread when `max(MessageEvent.timestamp) > readAt`, matching the semantic already used by `findLastMessageTimestamps`.

**Current query**
```cypher
OPTIONAL MATCH (:User {id: $userId})-[state:WATCHES]->(c)
WITH c, state.readAt AS readAt
WHERE readAt IS NULL OR c.modified > readAt
RETURN count(c)
```

**Correct signal**
```cypher
OPTIONAL MATCH (msg:MessageEvent {caseId: c.id})
WITH c, max(msg.timestamp) AS lastMessageAt
WHERE lastMessageAt IS NOT NULL
OPTIONAL MATCH (:User {id: $userId})-[state:WATCHES]->(c)
WITH lastMessageAt, state.readAt AS readAt
WHERE readAt IS NULL OR lastMessageAt > readAt
RETURN count(*)
```

**Impact**

Any client consuming `/unread-count` will receive a stale or incorrect count after new messages arrive in existing cases.

mathieu-de-armey-whoz:
`countUnread` compared `c.modified > readAt` but `Case.modified` is never updated on message activity.
Now uses `max(MessageEvent.timestamp)` per case, matching the frontend `lastMessageAt` semantic.
Cases with no messages are excluded. Tests updated accordingly.

Fixes #1295

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
@@ -242,12 +242,12 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
      *
      * A case is unread when:
      * - The user has a direct `[:ADMIN|MEMBER]` edge on the case (access check), AND
-     * - Either no `[:WATCHES]` edge exists (never opened), OR `Case.modified` is after
-     *   the edge's `readAt` (the case changed since the user last read it).
+     * - At least one `MessageEvent` exists on the case (`lastMessageAt` is not null), AND
+     * - Either no `[:WATCHES]` edge exists (`readAt IS NULL`), OR `max(MessageEvent.timestamp)`
+     *   is after the edge's `readAt`.
      *
-     * Uses `Case.modified` rather than scanning `CaseEvent` nodes — O(1) per case
-     * instead of O(events). Requires the `modified` audit field to be kept current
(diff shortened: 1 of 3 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseNodeNeo4jRepository.kt
```diff
@@ -256,10 +256,13 @@ interface CaseNodeNeo4jRepository : Neo4jRepository<CaseNode, String> {
         $$"""MATCH (c:Case)-[:BELONGS_TO]->(ns:Namespace {id: $namespaceId})
             WHERE (c.removed IS NULL OR c.removed = false)
               AND EXISTS { MATCH (:User {id: $userId})-[:ADMIN|MEMBER]->(c) }
+            OPTIONAL MATCH (msg:MessageEvent {caseId: c.id})
```
thomas-martin-whoz, line 259: in the previous PR we went against querying events for performance

PR #1298, opened 2026-09-07
Feature/benjamin/#1292 display the resolved model

benjamin-valdes-whoz:
image

image


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/delegation/DelegationTool.kt
```diff
@@ -213,6 +216,20 @@ class DelegationTool(
...
     ): DelegationResult {
         val agentName = delegation.agentName
+        // On resume, retain the durable event ids that predate this delegation. The
+        // reloaded history can then be restricted to this turn, rather than attributing
+        // a model from an earlier execution of the same sub-case.
+        val baselineEventIds =
+            delegation.subCaseId
+                ?.let { subCaseId ->
+                    runCatching {
+                        withTimeout(eventLoadTimeoutMs) { loadCaseEvents(subCaseId).map { it.id }.toSet() }
```
vincent-audibert-whoz: Loading the subcase events is abusive and feels disproportionate regarding starting the delegation. On the delegating agent side, all we know is the agent name and we should keep it that way. Resolving the delegated agent name into a model and provider is for the sub-case runtime, not the parent one.

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -894,6 +900,22 @@ export class CaseChatComponent implements OnInit, OnDestroy {
     return this.extractText(item.event)
   }
 
+  /** Provider/model reported by the nearest preceding execution event for this agent message. */
+  protected agentModelAttribution(message: CaseMessageEvent): string | null {
```
vincent-audibert-whoz, line 995: Will work but we should still aim to provide the information in the CaseEvent.kt - MessageEvent class for direct use in the front.

vincent-audibert-whoz (commented):
PR does not feel settled, I’d recommend different directions:
- leave delegation aside
- enrich the MessageEvent (and possibly the TextChunkEvent + AgentFinishedEvent) in the backend so front has basically no data processing to do.

Adding to the MessageEvent that was aimed at covering users and agents, might need some finesse or use an alternate field like `subtitle` to add in `Actor.kt` ?

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.html
```diff
@@ -188,54 +191,70 @@
...
-                              <pre>{{ phase.prompt }}</pre>
-                            </div>
-                            <div class="case-chat__enrichment-phase-row">
-                              <span class="case-chat__enrichment-phase-label">LLM output</span>
-                              <pre>{{ phase.llmOutput }}</pre>
-                            </div>
-                            @if (phase.enrichmentContent) {
+                  @if (
+                    item.call.response &&
+                    (item.call.toolName === 'DELEGATE__delegate' ||
```
vincent-audibert-whoz: Do not base display of specific tool name if possible. 
We might lack a specific SubCaseEvent to mark a link to a sub-case here ?

PR #1301, opened 2026-09-07
feat: #WZ-34476 add runCostThreshold to platform, namespace, and case

vincent-audibert-whoz:
## What

Adds a `runCostThreshold` property at three levels:

- **Platform** (`LimitsConfigProperties.runCostThreshold = 10.0`) — the platform-wide default, overridable via `AGENTOS_LIMITS_RUN_COST_THRESHOLD`. Default is null.
- **Namespace** (`Namespace.runCostThreshold: Double?`) — per-team override, nullable (inherits platform default when null)
- **Case** (`Case.runCostThreshold: Double?`) — per-case override, nullable (inherits namespace then platform when null); also the field the future negotiation mechanism will write when a user chooses to continue after a breach

All three levels are persisted (Neo4j nodes updated) and exposed via the existing REST API (PUT endpoints).

## What it is not

This PR contains **no enforcement logic**. No evaluation, no interruption, no user prompt. The property exists and is editable — wiring it into the agent turn lifecycle is the next step.

## Semantics

`runCostThreshold` is an anti-drift guard-rail, **not a budget**. It bounds what AgentOS spends between two human interventions. A user can open multiple cases and spend a multiple of the threshold across them.

Resolution chain (documented, not yet implemented): `Case.runCostThreshold ?: Namespace.runCostThreshold ?: LimitsConfigProperties.runCostThreshold`

`null` on Case or Namespace means "inherit from the level above" — never "zero" and never "unlimited". A case that materialises its own value exits the inherited regime and is unaffected by later namespace or platform changes.

## KDoc correction

`LimitsConfigProperties` previously classified `agentMaxIterations` as a terminal guard-rail alongside `caseMaxIterations`. That was wrong: `AgentAdvanced` emits a `WarnEvent`, breaks the loop, and still produces a final response — graceful degradation, not termination. The class KDoc now documents three distinct behaviours:

- `caseMaxIterations` → terminal (`CaseStatus.ERROR`)
- `agentMaxIterations` → graceful degradation (warn + final response, case not terminated)
- `runCostThreshold` → negotiable (user decision, never an error)


libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -254,6 +254,50 @@ export class CaseChatComponent implements OnInit, OnDestroy {
...
+  protected commitEdit(): void {
+    const c = this.activeCase()
+    if (!c?.id) return
+    const title = this.draftTitle().trim()
+    if (!title) return // empty title not allowed
+    const rawThreshold = this.draftThreshold().trim()
+    const parsedThreshold = rawThreshold === '' ? undefined : parseFloat(rawThreshold)
+    if (rawThreshold !== '' && isNaN(parsedThreshold as number)) return
+    this.isEditing.set(false)
+    this.updateRequested.emit({ id: c.id, title, runCostThreshold: parsedThreshold })
```
selimbensenouciep-prog, line 309: When navigating from case A to case B in the same namespace, `caseId` changes but `activeCase` can still point to A because it only depends on the `cases()` signal. If the list does not change, saving the editor updates A instead of B. This can persist when B has no SSE events.

Could we make the active case ID reactive and add a test that switches from A to B without SSE events, then checks that saving updates B?

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
```diff
@@ -254,6 +254,50 @@ export class CaseChatComponent implements OnInit, OnDestroy {
...
+  protected readonly isEditing = signal(false)
+  protected readonly draftTitle = signal('')
+  protected readonly draftThreshold = signal<string>('')
+
+  protected startEdit(): void {
+    const c = this.activeCase()
+    if (!c) return
+    this.draftTitle.set(c.title ?? '')
+    this.draftThreshold.set(c.runCostThreshold != null ? String(c.runCostThreshold) : '')
+    this.isEditing.set(true)
```
selimbensenouciep-prog, line 286: If a run finishes while this editor is open, the existing focus effect moves focus back to the chat composer without checking `isEditing`. Further typing then goes into the composer, and pressing Enter can send a message instead of saving the edit.

Could we keep focus in the editor while editing? The `isEditing` check should also run inside the queued microtask to cover an edit started after the focus callback was scheduled.

apps/server/src/lib/agentos-runtime.ts
```diff
@@ -279,6 +279,8 @@ export async function startAgentos(
     // Replace with real cryptographic values when secret management is in place.
     env: {
       ...process.env,
+      // Setup default runCostThreshold
+      AGENTOS_LIMITS_RUN_COST_THRESHOLD: '10',
```
selimbensenouciep-prog, line 279: Since this assignment comes after `...process.env`, it replaces an explicitly configured `AGENTOS_LIMITS_RUN_COST_THRESHOLD` with `10`. For example, starting Coday with a threshold of `2` still passes `10` to AgentOS.

Could we use `process.env.AGENTOS_LIMITS_RUN_COST_THRESHOLD ?? '10'` here, as the neighboring settings do, so `10` is only the fallback?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
```diff
@@ -95,9 +95,19 @@ class CaseServiceImpl(
...
-        // .
-        val saved = caseRepository.save(entity)
+        // Materialise runCostThreshold at creation time from the resolution chain:
+        // Case (caller-supplied) ?: Namespace.runCostThreshold ?: platform default.
+        // A non-null value on the incoming entity is an explicit caller override — kept as-is.
+        // Materialising at creation rather than resolving at runtime means the case is
+        // unaffected by later namespace or platform config changes (same principle as
+        // UsageRecord denormalising provider pricing at record time).
+        val resolvedThreshold: Double? =
+            entity.runCostThreshold
```
selimbensenouciep-prog, line 105: A threshold supplied to `POST /api/cases` is currently ignored because `CaseController.create()` does not copy `resource.runCostThreshold` into the `Case` passed to the service. For example, a request with `runCostThreshold: 25` ends up using the namespace or platform default.

Could we include this field in the controller mapping and add a test that checks the threshold supplied in the creation request is preserved?

libs/agentos-ui/src/lib/services/case-state.service.ts
```diff
@@ -114,11 +114,50 @@ export class CaseStateService {
...
+      const existing = this.cases().find((c) => c.id === caseId)
+      if (!existing) {
+        return throwError(() => new Error(`[CaseState] Case ${caseId} is not in the current list`))
+      }
+      const previous = { title: existing.title, runCostThreshold: existing.runCostThreshold }
+      this.patchFields(caseId, patch)
+      // Build the payload: spread existing then override with patch.
+      // undefined fields are omitted by JSON.stringify, so runCostThreshold: undefined
+      // removes the key from the body, which is the intended "clear" semantics.
+      const payload: Case = { ...existing, ...patch }
```
selimbensenouciep-prog, line 152: Clearing the threshold in the editor removes `runCostThreshold` from the JSON payload, but `CaseController.update()` keeps the existing value when the field is missing or null. The optimistic update still removes the threshold from the UI, and the successful response is ignored. A threshold of `10` therefore disappears until the next reload, although the server still stores `10`.

Could we align the editor with the API contract? If resetting the threshold is not supported, clearing the input should not be presented as a successful change.

libs/agentos-ui/src/lib/services/case-state.service.ts
```diff
@@ -114,11 +114,50 @@ export class CaseStateService {
...
+      return this.caseController.updateCase(caseId, payload).pipe(
+        catchError((err) => {
+          this.patchFields(caseId, previous)
```
selimbensenouciep-prog: An older failed request can undo a newer successful edit in the UI. The editor can be reopened while a save is pending, and this handler always restores the original values. Starting with a threshold of `10`, if saving `30` succeeds before an earlier save to `20` fails, the UI goes back to `10` while the server keeps `30`.

Could we prevent overlapping saves for the same case, or make the rollback conditional on that request still being the latest update? A test with responses arriving out of order would cover this.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/LimitsConfigProperties.kt
```diff
@@ -0,0 +1,106 @@
...
+     * agent run, and redirect events. The counter resets to 0 each time an
+     * [io.whozoss.agentos.sdk.caseEvent.AgentFinishedEvent] is processed, so
+     * a case with multiple sequential commands can safely exceed this threshold
+     * across the full session without triggering it within a single turn.
+     *
+     * Exceeding this limit transitions the case to [io.whozoss.agentos.sdk.caseFlow.CaseStatus.ERROR].
+     *
+     * Defaults to 100.
+     */
+    val caseMaxIterations: Int = 100,
```
selimbensenouciep-prog, line 64: These settings are documented as configurable through `AGENTOS_LIMITS_CASE_MAX_ITERATIONS` and `AGENTOS_LIMITS_AGENT_MAX_ITERATIONS`, but neither value is passed to the runtime. `CaseRuntime` still uses `100`, and `AgentAdvanced` is constructed with its default of `20`, so changing the environment variables has no effect.

Could we either pass these values to the existing iteration guards or remove the settings and their documentation until they are supported?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/namespace/NamespaceDto.kt
```diff
@@ -38,4 +41,11 @@ data class NamespaceDto(
                 " When null, messages without an @mention will produce an explicit error.",
     )
     val defaultAgentName: String? = null,
+    @field:Schema(
+        description =
+            "Namespace-level default for the run cost threshold, in the platform currency unit. " +
+                "Null means inherit from the platform default. Applies to all cases in this namespace " +
+                "that do not declare their own threshold.",
+    )
+    val runCostThreshold: Double? = null,
```
selimbensenouciep-prog, line 52: The case and namespace DTOs have no validation for negative thresholds. A negative value supplied to the namespace API is passed through to persistence and then copied onto newly created cases. The inputs specify `min="0"`, but the namespace form only checks `nameControl.invalid` before submitting.

Could we reject negative values in both DTOs, keep `null` allowed, and validate the threshold before submitting from the UI? Tests covering negative, zero and null values would make the contract explicit.

selimbensenouciep-prog (dismissed):

libs/agentos-ui/src/lib/components/namespace-form/namespace-form.component.html
```diff
@@ -70,6 +70,20 @@ <h1 class="namespace-form__title">{{ isEditMode() ? 'Edit namespace' : 'New name
...
+        <input
+          id="ns-run-cost-threshold"
+          class="namespace-form__input"
+          type="number"
+          min="0"
+          step="0.01"
+          placeholder="Optional maximum cost per run (e.g. 0.50)"
+          [formControl]="runCostThresholdControl"
+        />
+        <span class="namespace-form__hint">Leave empty for no limit.</span>
```
selimbensenouciep-prog: An empty value means the namespace inherits the platform threshold. For example, with a platform threshold of `10`, new cases still receive `10` when this field is left empty.

Could we replace this hint with "Leave empty to inherit the platform default." so it matches the actual behavior?

vincent-audibert-whoz: changed

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/namespace/NamespaceDto.kt
vincent-audibert-whoz: validation added and tested

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/LimitsConfigProperties.kt
vincent-audibert-whoz: Settings kept for later PR about cost tracking and infinite loop avoidance.

apps/server/src/lib/agentos-runtime.ts
vincent-audibert-whoz: order reversed: we put our values then let `process.env` override it.

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
vincent-audibert-whoz: check added

libs/agentos-ui/src/lib/services/case-state.service.ts
vincent-audibert-whoz: corrected, clearing does not change the value in the UI after validation

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseFlow/CaseServiceImpl.kt
vincent-audibert-whoz: value taken on create

libs/agentos-ui/src/lib/components/case-chat/case-chat.component.ts
vincent-audibert-whoz: activeCaseId made reactive + could not replay this scenario on manual test.

libs/agentos-ui/src/lib/services/case-state.service.ts
vincent-audibert-whoz: sequence handled + tests added.

selimbensenouciep-prog (dismissed):

apps/server/src/lib/agentos-runtime.ts
```diff
@@ -273,31 +273,26 @@ export async function startAgentos(
...
-    // (temporary — credentials stored in plaintext on disk; AgentOS logs a WARN).
-    // Replace with real cryptographic values when secret management is in place.
+    // Setup default values then take possible existing overrides
     env: {
-      ...process.env,
+      // Setup default runCostThreshold
+      AGENTOS_LIMITS_RUN_COST_THRESHOLD: '10',
       // Fallback to NONE only if not already set — real values from the environment
       // take precedence (production deployments with actual encryption keys).
-      AGENTOS_ENCRYPTION_KEY: process.env.AGENTOS_ENCRYPTION_KEY ?? 'NONE',
```
vincent-couturier-whoz, line 284: question: why this removal ?

apps/server/src/lib/agentos-runtime.ts
```diff
@@ -273,31 +273,26 @@ export async function startAgentos(
...
       // Also update the spring.neo4j.uri to match so SDN connects to the right port.
       // Both env vars respect any existing value; SPRING_NEO4J_URI falls back to
       // whichever Bolt port was ultimately chosen (env override or dynamic).
-      AGENTOS_PERSISTENCE_EMBEDDED_BOLT_PORT:
-        process.env.AGENTOS_PERSISTENCE_EMBEDDED_BOLT_PORT ?? String(neo4jBoltPort),
-      SPRING_NEO4J_URI:
-        process.env.SPRING_NEO4J_URI ??
-        `bolt://localhost:${process.env.AGENTOS_PERSISTENCE_EMBEDDED_BOLT_PORT ?? neo4jBoltPort}`,
+      AGENTOS_PERSISTENCE_EMBEDDED_BOLT_PORT: String(neo4jBoltPort),
+      SPRING_NEO4J_URI: `bolt://localhost:${process.env.AGENTOS_PERSISTENCE_EMBEDDED_BOLT_PORT ?? neo4jBoltPort}`,
```
vincent-couturier-whoz, line 294: question: why the removal of process.env.SPRING_NEO4J_URI  ?

vincent-couturier-whoz (approved):

apps/server/src/lib/agentos-runtime.ts
```diff
@@ -273,31 +273,26 @@ export async function startAgentos(
...
+      ...process.env,
```
vincent-couturier-whoz: issue(blocking): this destructuring instruction seems too dangerous to be applied like that

PR #1306, opened 2026-09-08
feat: wz-30464 - fix migration flag

thomas-martin-whoz:


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
```diff
@@ -407,7 +407,7 @@ class Neo4jPersistenceConfiguration {
                         REMOVE flag.isNew
                         WITH count(flag) AS guard
                         MATCH (u:User)-[:ADMIN|MEMBER]->(c:Case)
-                        WHERE (c.removed IS NULL OR c.removed = false)
+                        WHERE (c.removed IS NULL OR c.removed = false) and guard > 0
```
vincent-couturier-whoz, line 410: question: it's strange that the processing continue even if there's not matching node due to the where clause 🤔

vincent-couturier-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/config/Neo4jPersistenceConfiguration.kt
thomas-martin-whoz: the processing does not continue, that's the point, if this is not the first time we do not want to do the data migration

PR #1309, opened 2026-09-08
feat: introduce UsageRecord entity, persistence, service and controller

vincent-audibert-whoz:
## What

Introduces `UsageRecord` as a first-class analytical entity for cost and token tracking, ported and scoped from `feat/vincent.audibert/wz-34476-track-cost-1` — without the emission wiring into `CaseServiceImpl` (that comes next).

## What's included

**SDK (`agentos-sdk`)**
- `sdk/usage/LlmUsage` — value object for aggregated LLM token usage, with null-cost-contaminating `plus` operator
- `sdk/api/usageRecord/` — `UsageRecordDto`, `UsageAggregateDto`, `UsageAggregateByKeyDto`, `UsageRecordApi` contract interface

**Service (`agentos-service`)**
- `UsageRecord` entity + `UsageSource` / `UsageOutcome` enums
- `UsageAggregate` / `UsageAggregateByKey` aggregation types
- `UsageRecordRepository` / `UsageRecordService` / `UsageRecordServiceImpl`
- `UsageRecordNode` / `UsageRecordNodeNeo4jRepository` / `Neo4jUsageRecordRepository` — full Neo4j persistence with all aggregation queries and null-cost contamination strategy
- `UsageRecordSchemaInitializer` — Neo4j indexes on `caseId`, `namespaceId`, `userId`, `timestamp` (all `IF NOT EXISTS`, idempotent)
- `UsageRecordController` — read and aggregation endpoints under `/api/usage-records`, gated on existing Case/Namespace `READ` permissions
- Bean registration in `Neo4jPersistenceConfiguration` + `usage_record_id_unique` constraint in `Neo4jSchemaInitializer`

**Dedicated query: `sumCostByCaseTreeSince`**
- Traverses the `[:PARENT_OF]` delegation tree from a root case, filters by `timestamp >= since`, returns the summed cost as `Double?`
- `null` = unknown (at least one unpriced record, or no records yet) — never silently zero
- Index coverage: `Case.id` unique constraint for tree resolution, `usage_record_case_id` B-tree for the UsageRecord lookups (≤6 seeks at platform depth limit of 5)

**Tests**
- `InMemoryUsageRecordRepository` — test fixture with in-memory parent-child graph for tree traversal
- `UsageRecordNodeUnitSpec` — domain↔node round-trip and `fromLlmUsage` factory
- `UsageAggregateUnitSpec` — null-cost contamination semantics across all aggregation axes + `sumCostByCaseTreeSince`

## What's NOT included (intentionally)

Emission from `CaseServiceImpl` — the wiring that actually writes `UsageRecord` after each agent run. That's the next step.


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/UsageRecordNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,254 @@
...
+     */
+    @Transactional(readOnly = true)
+    @Query(
+        $$"""MATCH (root:Case {id: $rootCaseId})
+            WITH collect(root.id) + [desc IN [(root)-[:PARENT_OF*1..10]->(d:Case) | d.id] | desc] AS caseIds
+            MATCH (u:UsageRecord)
+            WHERE u.caseId IN caseIds
+              AND u.timestamp >= $since
+              AND (u.removed IS NULL OR u.removed = false)
+            RETURN collect({
```
selimbensenouciep-prog, line 248: This query fails on Neo4j because `count()` and `sum()` are nested inside `collect()`. I reproduced it on Neo4j 2026.02.2 with priced records, records with unknown costs, and no records.

Could you move these calculations into a `WITH` clause before calling `collect()`? A test against embedded Neo4j would also catch this, since the current in-memory tests don't execute the query.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/usageRecord/UsageAggregateDto.kt
```diff
@@ -0,0 +1,30 @@
...
+@Schema(name = "UsageAggregate")
+data class UsageAggregateDto(
+    val recordCount: Long,
+    val inputTokens: Long,
+    val outputTokens: Long,
+    val cacheReadTokens: Long,
+    val cacheWriteTokens: Long,
+    val totalTokens: Long,
+    /** null = at least one record had unknown cost. */
+    val cost: Double?,
```
selimbensenouciep-prog, line 22: The API returns `cost: null` for unknown costs, but the generated client declares `cost?: number`. A caller checking `cost !== undefined` can then call `cost.toFixed(2)`: TypeScript accepts it, but it crashes when the API returns null.

Could you mark `cost` as nullable in the OpenAPI schema and regenerate the client? The same issue applies to `UsageRecordDto.cost`.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/UsageRecordNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,254 @@
...
+            WITH u.agentName AS agentName,
+                 count(u) AS recordCount,
+                 sum(u.inputTokens) AS inputTokens,
+                 sum(u.outputTokens) AS outputTokens,
+                 sum(u.cacheReadTokens) AS cacheReadTokens,
+                 sum(u.cacheWriteTokens) AS cacheWriteTokens,
+                 sum(u.totalTokens) AS totalTokens,
+                 sum(CASE WHEN u.cost IS NOT NULL THEN u.cost ELSE 0 END) AS partialCostSum,
+                 sum(CASE WHEN u.cost IS NULL THEN 1 ELSE 0 END) AS nullCostCount
+            RETURN collect({
```
selimbensenouciep-prog, line 160: The repository contract promises results ordered by total tokens descending, and the in-memory implementation does sort them. Neither Neo4j query guarantees that order.

Could you add `ORDER BY totalTokens DESC` before `RETURN collect(...)` in both `aggregateByAgent` and `aggregateByModel`, and cover the ordering with a test?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/usage/LlmUsage.kt
```diff
@@ -0,0 +1,72 @@
...
+            totalTokens = totalTokens + other.totalTokens,
+            estimatedCostUsd =
+                when {
+                    estimatedCostUsd != null && other.estimatedCostUsd != null -> estimatedCostUsd + other.estimatedCostUsd
+                    else -> null
+                },
+        )
+
+    companion object {
+        val ZERO = LlmUsage()
```
selimbensenouciep-prog: Should `ZERO` have `estimatedCostUsd = 0.0`? Currently, adding it to a usage with a known cost makes the result's cost `null`. A caller using `fold(LlmUsage.ZERO)` would therefore get an unknown total even if every usage is priced.

Setting its cost to `0.0` would make `ZERO` neutral for addition, while usages with an unknown cost would still propagate `null`.

selimbensenouciep-prog (dismissed):

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/usageRecord/UsageAggregateDto.kt
selimbensenouciep-prog: I regenerated the OpenAPI spec after this change, but `cost` is still declared as a non-nullable number in both DTO schemas. The project uses OpenAPI 3.1, where `nullable = true` doesn't express nullability, so the client still gets `cost?: number` even though the API can return `null`.

Could you use `@field:Schema(types = ["number", "null"], format = "double")` on both fields and regenerate the OpenAPI spec and TypeScript client?

selimbensenouciep-prog (dismissed):

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/api/usageRecord/UsageAggregateDto.kt
vincent-audibert-whoz: Good suggestion 👍

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/UsageRecordNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,259 @@
...
+            WHERE u.caseId = $caseId AND (u.removed IS NULL OR u.removed = false)
+            WITH count(u) AS recordCount,
+                 sum(u.inputTokens) AS inputTokens,
+                 sum(u.outputTokens) AS outputTokens,
+                 sum(u.cacheReadTokens) AS cacheReadTokens,
+                 sum(u.cacheWriteTokens) AS cacheWriteTokens,
+                 sum(u.totalTokens) AS totalTokens,
+                 sum(CASE WHEN u.cost IS NOT NULL THEN u.cost ELSE 0 END) AS partialCostSum,
+                 sum(CASE WHEN u.cost IS NULL THEN 1 ELSE 0 END) AS nullCostCount
+            RETURN collect({
```
vincent-couturier-whoz, line 59: Question: why a collect here intead of just the fields as there will be only one or no value ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/UsageRecordNode.kt
```diff
@@ -0,0 +1,117 @@
...
+    val outputTokens: Long = 0L,
+    val cacheReadTokens: Long = 0L,
+    val cacheWriteTokens: Long = 0L,
+    val totalTokens: Long = 0L,
+    val cost: Double? = null,
+    val timestamp: Instant = Instant.now(),
+    // Audit fields from EntityMetadata
+    val created: Instant = Instant.now(),
+    val createdBy: String? = null,
+    val modified: Instant = Instant.now(),
```
vincent-couturier-whoz, line 53: issue: You should use a clock or just not default value

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/Neo4jUsageRecordRepository.kt
```diff
@@ -0,0 +1,167 @@
...
+            UsageAggregateByKey(
+                key = row[keyField] as String,
+                aggregate = UsageAggregate(
+                    recordCount = (row["recordCount"] as Number).toLong(),
+                    inputTokens = (row["inputTokens"] as Number).toLong(),
+                    outputTokens = (row["outputTokens"] as Number).toLong(),
+                    cacheReadTokens = (row["cacheReadTokens"] as Number).toLong(),
+                    cacheWriteTokens = (row["cacheWriteTokens"] as Number).toLong(),
+                    totalTokens = (row["totalTokens"] as Number).toLong(),
+                    cost = if (nullCostCount > 0L) null else partialCostSum,
```
vincent-couturier-whoz: question: why a null on a cost give 0 on the sum ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/usage/UsageRecordNodeNeo4jRepository.kt
```diff
@@ -0,0 +1,258 @@
...
+    //
+    // Aggregations wrap each row in a Cypher map literal and collect the rows into
+    // a single list column. This envelope is intentional: this SDN version tries to
+    // map direct multi-column results through the repository entity converter.
+    //
+    // Null-cost contamination strategy:
+    //   Cypher's sum() silently ignores nulls and would produce a silent undercount.
+    //   We count records with null cost separately (nullCostCount) and return that count
+    //   alongside the partial sum. The Kotlin caller replaces the partial sum with null
+    //   when nullCostCount > 0.
```
vincent-couturier-whoz: suggestion: we should raise the null cost count to have the information but we should keep the sum of costs in order to have the a minimal cost

vincent-couturier-whoz (approved):

PR #1318, opened 2026-09-09
fix: #1317 surface provider error body on streaming path

### LINKED ISSUE #1317: OpenAI 400 Bad Request on gpt-5.6 model family (AgentOS) + error body lost on streaming path
## Symptom

Agent runs against the recent OpenAI `gpt-5.6` family (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`) fail in production:

```
Error during agent execution
org.springframework.web.reactive.function.client.WebClientResponseException$BadRequest: 400 Bad Request from POST https://api.openai.com/v1/chat/completions
	at org.springframework.web.reactive.function.client.WebClientResponseException.create(WebClientResponseException.java:321)
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below:
	Error has been observed at the following site(s):
	*__checkpoint ⇢ 400 BAD_REQUEST from POST https://api.openai.com/v1/chat/completions [DefaultWebClient]
```

## Two distinct problems

This issue covers two defects. The second one is why we cannot yet name the first one with certainty.

### 1. Observability defect — the OpenAI error body is discarded on the streaming path

The exception type is the tell. `WebClientResponseException$BadRequest` reaching our catch block means Spring AI's error translation **did not run**.

- Blocking calls (`chatClient.prompt().call()`) go through `RestClient` with `RetryUtils.DEFAULT_RESPONSE_ERROR_HANDLER`, which reads the HTTP body and wraps it: `throw new NonTransientAiException(status + " - " + body)`. The OpenAI message (`{"error": {"message": "..."}}`) survives.
- Streaming calls (`chatClient.prompt().stream()`) go through `WebClient`. `OpenAiChatModel.internalStream()` has no `onErrorMap` for `WebClientResponseException`, so the raw exception traverses the `Flux`, reaches `asFlow()` in the agent, and lands in the generic `catch (e: Exception)` branch — **not** in `catch (e: NonTransientAiException)`.

Consequence: we log `e.message`, which for a `WebClientResponseException` contains only status + method + URL. The response body — the only part that says *which* parameter OpenAI rejected — is available on `responseBodyAsString` but never read.

`AgentSimple` uses `.stream()` exclusively for the main LLM call, so **every** provider rejection in a normal agent run hits this blind spot.

Affected:
- `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentSimple.kt` — `catch (e: Exception)` around the stream collection
- `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentAdvanced.kt` — same pattern
- `agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterruptHandler.kt` — `emitProviderErrorAndFinishEvents`, currently only reachable from the `NonTransientAiException` branch

### 2. Payload defect — unsupported parameters sent to reasoning models

`ChatModelFactory.createOpenAiModel()` builds the request options unconditionally:

```kotlin
val optionsBuilder = OpenAiChatOptions.builder()
    .temperature(temp)          // temp = temperature ?: DEFAULT_TEMPERATURE (1.0)
    .model(model)
if (maxTokens != null) {
    optionsBuilder.maxCompletionTokens(maxTokens)
}
```

`temperature` is **always** sent. The `gpt-5.6` family are reasoning models, and reasoning models reject any `temperature` other than the default. If the production `AiModel` record for `gpt-5.6-*` carries a `temperature` different from `1.0`, OpenAI answers 400 with `Unsupported value: 'temperature' does not support <x> with this model. Only the default (1) value is supported.`

Supporting evidence: the TypeScript prototype (`libs/handler/src/lib/openai.client.ts`) declares all three `gpt-5.6` models with `temperature: 1` and additionally pins `reasoning_effort: 'none'` with an explicit comment that these models reject reasoning configuration when tools are present. AgentOS never sets `reasoning_effort`, so it silently gets the `medium` default.

Note: the `gpt-5.6-*` model definitions are **not** in `agentos/aimodel/*.yaml` (only `gpt-5-mini.yaml` and `claude-haiku.yaml` are). They are persisted in Neo4j via the REST API, so the offending `temperature` value must be read from the running instance to confirm.

## Underlying architectural gap

`AiModel` (`agentos-sdk/.../sdk/aiProvider/AiModel.kt`) exposes `temperature` and `maxTokens` as flat nullable fields, and `ChatModelFactory` forwards them verbatim. There is no way to express *"this model does not accept this parameter"*.

Provider parameter contracts now vary **per model family within a single provider**, not just per provider. The current `apiType`-based branching (`OpenAI` vs `vLLM`) is too coarse: it already had to special-case `maxCompletionTokens` vs `maxTokens`, and this issue adds `temperature` and `reasoning_effort` to the list. Each new model family will keep bending this code until the parameter policy becomes an explicit, named concept.

Also worth noting: `AgentConfig` carries no `temperature`, so temperature is only ever model-level — there is no per-agent override to blame.

## Test gap

`ChatModelFactoryUnitSpec` only asserts `model.shouldBeInstanceOf<OpenAiChatModel>()`. That is tautological — it restates the factory signature and can never catch a payload defect. There is currently **no test that observes the JSON actually sent to OpenAI**, and **no test covering provider error handling on the streaming path**.

## Proposed direction

1. Map 4xx `WebClientResponseException` to `NonTransientAiException` (including `responseBodyAsString`) on the streaming path, so provider rejections surface a readable cause. Fix first — it is what makes the rest diagnosable.
2. Stop sending parameters a model does not accept, via an explicit parameter policy rather than scattered conditionals.
3. Add a test seam on the outgoing request payload (see comments for options).
4. Secondary: `AgentAdvanced`'s `catch (e: Exception)` emits no `AgentFinishedEvent`, unlike `AgentSimple`. A case can be left hanging on a non-`NonTransientAiException` failure. Should be fixed here or tracked separately.

vincent-audibert-whoz:
Closes #1317 partially — observability only. **The root cause of the 400 is not fixed here** (see [Scope](#scope)).

## Problem

Provider rejections on the streaming path produced no usable diagnostic:

```
Error during agent execution
org.springframework.web.reactive.function.client.WebClientResponseException$BadRequest: 400 Bad Request from POST https://api.openai.com/v1/chat/completions
```

Spring AI 1.1.2 translates provider HTTP errors into `NonTransientAiException` **only on the blocking path**. `RestClient` + `RetryUtils.DEFAULT_RESPONSE_ERROR_HANDLER` reads the HTTP body and embeds it in the message. The streaming path goes through `WebClient`, and `OpenAiChatModel.internalStream()` has no `onErrorMap` — so the raw `WebClientResponseException` traverses the `Flux`, reaches `asFlow()`, and lands in the agents' generic `catch (e: Exception)` branch. We logged `e.message`, which carries only status + method + URL. The response body — the only place the provider names the offending field — sat unread on `responseBodyAsString`.

`AgentSimple` uses `.stream()` exclusively for its main LLM call, so **every** provider rejection in a normal agent run hit this blind spot. Diagnosing #1317 took four separate investigations and produced two hypotheses that were later weakened, purely for lack of the error body.

For contrast, here is what a path that surfaces the body gives you — immediately actionable:

```
Status: [400 BAD_REQUEST], Body:[{"type":"error","error":{"type":"invalid_request_error",
"message":"tools.9.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'"},
"request_id":"req_011CesTekr8YWSqVQjNK32uC"}]
```

## Solution

**`util/WebClientExceptionUtils.kt`** (new) — pure extension function `Throwable.unwrapToProviderAiException()`:
- unwraps the Reactor envelope (`Exceptions.unwrap`) then walks the `cause` chain (bounded depth)
- 4xx → `NonTransientAiException`, 5xx → `TransientAiException`, preserving Spring AI's retry semantics
- returns `null` for anything else, leaving propagation untouched
- short-circuits `CancellationException` on the first line, so coroutine cancellation can never be swallowed
- bounds the body to 4000 chars with an explicit truncation marker

**`agent/AgentInterruptHandler.kt`** — new `handleGenericAgentException()` that dispatches on the result. Both agents now delegate their generic catch to it.

### Why this location

`AgentInterruptHandler` is already the file every agent must use for interrupt and provider-error handling. A future agent copying the existing `catch (e: NonTransientAiException) { … } catch (e: Exception) { … }` pattern inherits the enriched behaviour **without needing to know about `WebClientResponseException`, Reactor, or the 4xx/5xx split**. The knowledge lives in one named place rather than scattered across catch blocks.

A `ChatClient` decorator was considered and rejected: intercepting the `Flux` before its conversion to a coroutine `Flow` would complicate cancellation propagation for no benefit here.

The fix is **provider-agnostic** — the defect is in Spring AI's streaming path, not in the OpenAI client. No per-provider branching was introduced, so Anthropic, Gemini, vLLM and Ollama all benefit.

## Behaviour change worth reviewing

`AgentAdvanced`'s generic `catch` previously emitted **no** `AgentFinishedEvent`, unlike `AgentSimple` — a latent defect that could leave a case hanging. `handleGenericAgentException` emits one in all branches, so this is now fixed as a side effect. It is consistent with `AgentSimple` and with the runtime always expecting an `AgentFinishedEvent` to close a run cleanly, but it is **beyond the strict observability scope** and deserves a reviewer's eye.

## Tests

18 new tests (Kotest + mockk), all green. Full `agentos-service` suite passes with no regression.

`WebClientExceptionUtilsUnitSpec` (14) covers: bare 4xx, **Reactor-wrapped 4xx** (the actual production shape), doubly-wrapped, 5xx → transient, non-HTTP exceptions passing through unchanged, **`CancellationException` never converted**, body truncation, empty body.

`AgentSimpleProviderErrorUnitSpec` (4) asserts the emitted `ErrorEvent` now carries the body, on both the bare and Reactor-wrapped paths, plus a non-regression case for generic exceptions.

One test deliberately **documents** that 429 is currently classified non-retryable (via `is4xxClientError`). That is semantically debatable — a rate limit is retryable — but reclassifying it was out of scope. Pinning it in a test makes any future change intentional rather than accidental.

## Scope

Deliberately **not** in this PR, pending an architectural decision:
- inference parameter handling (`temperature`, `maxTokens`, `reasoningEffort`) in `ChatModelFactory`
- MCP tool name and input-schema sanitisation

Both are symptoms of the same underlying gap: requests are built without ever checking the destination's contract. That contract needs to become an explicit, named concept rather than accumulated conditionals — discussion continues on #1317.

## Validation

`compileKotlin` / `compileTestKotlin` succeed (remaining warnings are pre-existing). Full `agentos-service:test` suite: BUILD SUCCESSFUL, 0 failures.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterruptHandler.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterruptHandler.kt
@@ -6,9 +6,12 @@ import io.whozoss.agentos.sdk.caseEvent.AgentSelectedEvent
 import io.whozoss.agentos.sdk.caseEvent.CaseEvent
 import io.whozoss.agentos.sdk.caseEvent.ErrorEvent
 import io.whozoss.agentos.sdk.caseEvent.QuestionEvent
+import io.whozoss.agentos.sdk.caseEvent.WarnEvent
+import io.whozoss.agentos.util.unwrapToProviderAiException
 import kotlinx.coroutines.flow.FlowCollector
 import mu.KLogger
 import org.springframework.ai.retry.NonTransientAiException
+import org.springframework.ai.retry.TransientAiException
 import java.util.UUID
 
 /**
@@ -41,6 +44,79 @@ suspend fun FlowCollector<CaseEvent>.emitProviderErrorAndFinishEvents(
             llmModel = agent.llmModel,
         ),
     )
+}
+
+/**
+ * Handles any [Exception] that escapes the agent's main loop by first checking whether
+ * it wraps a provider HTTP error on the streaming path (a [WebClientResponseException]
+ * buried under Reactor envelopes).
+ *
+ * If [e] or any cause in its chain is a [WebClientResponseException], it is translated
+ * to the appropriate Spring AI exception ([NonTransientAiException] for 4xx,
+ * [TransientAiException] for 5xx) so the response body surfaces in logs and in the
+ * [ErrorEvent] — exactly as it does on the blocking path.
+ *
+ * If [e] is not a provider HTTP error, the function falls back to the original generic
+ * behaviour: log at ERROR level and emit a [WarnEvent].
+ *
+ * This function is the single point of streaming-path exception enrichment. Every agent
+ * that follows the `catch (e: NonTransientAiException) { … } catch (e: Exception) { … }`
+ * pattern must delegate the generic catch to this function so future agents automatically
+ * benefit without needing to know about [WebClientResponseException].
+ */
+suspend fun FlowCollector<CaseEvent>.handleGenericAgentException(
+    agent: Agent,
+    e: Exception,
+    namespaceId: UUID,
+    caseId: UUID,
+    logger: KLogger,
+) {
+    when (val providerException = e.unwrapToProviderAiException()) {
+        is NonTransientAiException -> emitProviderErrorAndFinishEvents(agent, providerException, namespaceId, caseId, logger)
+        is TransientAiException -> {
+            // 5xx — provider-side failure, potentially transient.
+            // Emit ErrorEvent (same lifecycle termination) but log at WARN since a retry may succeed.
+            logger.warn(e) { "LLM provider returned a transient error for case $caseId" }
+            emit(
+                ErrorEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    message = "The AI provider returned a transient error and the agent cannot continue: ${providerException.message}",
+                ),
+            )
+            emit(
+                AgentFinishedEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    agentId = agent.id,
+                    agentName = agent.name,
+                    llmProvider = agent.llmProvider,
+                    llmModel = agent.llmModel,
+                ),
+            )
+        }
+        else -> {
+            // Not a provider HTTP error — preserve the original generic behaviour.
+            logger.error(e) { "Error during agent execution" }
+            emit(
+                WarnEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    message = "Error during agent execution: ${e.message}",
+                ),
+            )
+            emit(
+                AgentFinishedEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    agentId = agent.id,
+                    agentName = agent.name,
+                    llmProvider = agent.llmProvider,
+                    llmModel = agent.llmModel,
+                ),
+            )
+        }
+    }
 }
 
 /**
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/util/WebClientExceptionUtils.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/util/WebClientExceptionUtils.kt
@@ -1,0 +1,83 @@
+package io.whozoss.agentos.util
+
+import kotlinx.coroutines.CancellationException
+import org.springframework.ai.retry.NonTransientAiException
+import org.springframework.ai.retry.TransientAiException
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import reactor.core.Exceptions
+
+/** Maximum number of characters retained from the HTTP response body in the error message. */
+private const val MAX_BODY_CHARS = 4_000
+
+/** Marker appended when the response body is truncated. */
+private const val TRUNCATION_MARKER = "…[truncated]"
+
+/**
+ * Unwraps the throwable chain (Reactor envelope + cause chain) looking for a
+ * [WebClientResponseException] that signals a provider-side HTTP error on a
+ * streaming call.
+ *
+ * Returns:
+ * - [NonTransientAiException] for 4xx responses (the request is faulty — retrying is useless).
+ * - [TransientAiException] for 5xx responses (provider-side failure — may be retried).
+ * - `null` for anything else, including [CancellationException] (which must never be swallowed).
+ *
+ * Reactor frequently wraps exceptions via [Exceptions.propagate] or composite exceptions.
+ * [Exceptions.unwrap] strips the Reactor envelope (one level). If the root cause is still
+ * not a [WebClientResponseException], the function walks the [Throwable.cause] chain up to
+ * [MAX_CAUSE_DEPTH] levels deep, covering nested wrapping patterns.
+ *
+ * [CancellationException] is never converted: Kotlin coroutine cancellation must propagate
+ * unchanged so the cooperative cancellation protocol is respected.
+ */
+fun Throwable.unwrapToProviderAiException(): Exception? {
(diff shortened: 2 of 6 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/util/WebClientExceptionUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+ * that is otherwise lost on the streaming path.
+ */
+private fun WebClientResponseException.toProviderAiException(): Exception {
+    val body = responseBodyAsString
+        .take(MAX_BODY_CHARS + TRUNCATION_MARKER.length)
+        .let { raw ->
+            if (raw.length > MAX_BODY_CHARS) raw.take(MAX_BODY_CHARS) + TRUNCATION_MARKER else raw
+        }
+        .ifBlank { "<empty body>" }
+    val message = "${statusCode} from ${request?.method} ${request?.uri}: $body"
```
vincent-couturier-whoz, line 78: Suggestion: the boy should be only available in debug log because it could contains sensitive data. Or if you want to keep it we should have an configuration property to activate the add of the boy in case of local test / execution.

vincent-couturier-whoz (dismissed):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterruptHandler.kt
```diff
@@ -43,6 +46,79 @@ suspend fun FlowCollector<CaseEvent>.emitProviderErrorAndFinishEvents(
...
+            // Emit ErrorEvent (same lifecycle termination) but log at WARN since a retry may succeed.
+            logger.warn(e) { "LLM provider returned a transient error for case $caseId" }
+            emit(
+                ErrorEvent(
+                    namespaceId = namespaceId,
+                    caseId = caseId,
+                    message = "The AI provider returned a transient error and the agent cannot continue: ${providerException.message}",
+                ),
+            )
+            emit(
```
vincent-couturier-whoz, line 87: issue: either you should reuse emitProviderErrorAndFinishEvents or you should not emit an AgentFinishedEvent but more an AgentIdleEvent to let the user send a message to retry.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/agent/AgentInterruptHandler.kt
```diff
@@ -43,6 +46,79 @@ suspend fun FlowCollector<CaseEvent>.emitProviderErrorAndFinishEvents(
...
+                    llmProvider = agent.llmProvider,
+                    llmModel = agent.llmModel,
+                ),
+            )
+        }
+        else -> {
+            // Not a provider HTTP error — preserve the original generic behaviour.
+            logger.error(e) { "Error during agent execution" }
+            emit(
+                WarnEvent(
```
vincent-couturier-whoz, line 102: question: Why an unknown error type is logged as just a warn when the effect are the same. It's very strange.

vincent-couturier-whoz (approved):

PR #1323, opened 2026-09-09
fix: #1322 sort tools by name for stable prompt cache prefix

### LINKED ISSUE #1322: Non-deterministic tool ordering invalidates OpenAI prompt cache prefix
## Context

OpenAI prompt caching is **enabled by default** — there is no opt-in, unlike Anthropic which requires explicit `cache_control` markers. From the OpenAI docs: *"Prompt caching is enabled by default for supported OpenAI models."*

This means the only lever we have is **prefix stability**. Cache reuse requires the entire rendered prefix to match, and **tool definitions sit at the front of that prefix**. Per the OpenAI caching guide, any change to tool *names, descriptions, schemas, or ordering* prevents reuse.

## Problem

The order of tools sent to the LLM is not deterministic by contract:

1. `integrationConfigService.findEffective()` returns `List<IntegrationConfig>` from Neo4j **with no `ORDER BY`**.
2. `ToolResolverService.resolveToolsForRun()` iterates that list, then calls `dedupToolsByName()`.
3. `dedupToolsByName()` relies on `.groupBy { it.name }` returning a `LinkedHashMap` — insertion order is preserved, so ordering is *incidentally* stable, but it is never asserted.
4. `AgentServiceImpl` applies a second `dedupToolsByName()` over `baseTools + delegationTools + exchangeTools + queryUserTools`.

The order therefore depends entirely on Neo4j's return order. If that changes — after adding or removing an integration, or a reindex — the tool block at the head of the prefix changes, and **the entire cached prefix is invalidated on every request**.

`AgentSimple` passes `tools.map { createToolCallback(...) }` straight to `prompt.toolCallbacks()`, so ordering directly shapes the prefix. (`AgentAdvanced` resolves tools by name from a generated intention and does not send the tool list, so it is unaffected on this axis.)

## Why it matters now

MCP tool definitions are verbose, and the minimum cacheable prefix is 1,024 tokens for GPT-5.6+. A rich toolset is exactly the case caching repays best: cache reads cost 0.1× the uncached input rate against 1.25× for a write, so one write plus nine reads costs 2.15× versus 10× uncached. Losing the prefix means paying full price on every turn.

## Proposed fix

Sort tools deterministically by name before they reach the LLM — in `dedupToolsByName()` or at the exit of `resolveToolsForRun()`. One line, no functional impact, and it turns an incidental property into a guaranteed one.

A regression test asserting stable ordering across repeated resolutions matters as much as the sort itself: the current stability is accidental, and nothing prevents a future refactor from silently breaking it again.

## Out of scope

- Creating an `OpenAiProperties` mirroring `AnthropicProperties`: it would activate nothing, since OpenAI caching is already on by default.
- `prompt_cache_key` (only relevant above 15 req/min for machine routing) and GPT-5.6 explicit cache breakpoints: both need measurement before they can be judged.

## Related gap

`cached_tokens` (`usage.prompt_tokens_details.cached_tokens` for OpenAI, `cache_read_input_tokens` / `cache_creation_input_tokens` for Anthropic) is **not read, logged or persisted anywhere**. There is no equivalent of the TypeScript prototype's `UsageLogger`, and no per-run cost tracking.

Consequence: we cannot currently answer *"is our cache working?"*, nor verify the effect of this fix. Worth tracking separately — it also applies to Anthropic, whose caching we enabled without ever measuring it.

vincent-audibert-whoz:
Closes #1322

## Problem

OpenAI prompt caching is **enabled by default** — there is no opt-in, unlike Anthropic which requires explicit `cache_control` markers. The only lever available is therefore **prefix stability**, and tool definitions sit at the very front of that prefix. Per the OpenAI caching guide, any change to tool names, descriptions, schemas **or ordering** prevents cache reuse.

Tool ordering was not guaranteed by contract:

- `integrationConfigService.findEffective()` returns `List<IntegrationConfig>` from Neo4j **with no `ORDER BY`**
- `dedupToolsByName()` relied on `.groupBy { it.name }` returning a `LinkedHashMap`, so insertion order was preserved — stability was **accidental and never asserted**

If Neo4j's return order changed (adding or removing an integration, a reindex), the tool block at the head of the prefix would change and the entire cached prefix would be invalidated on every request. MCP tool definitions are verbose, and cache reads cost 0.1× the uncached input rate against 1.25× for a write — one write plus nine reads costs 2.15× versus 10× uncached. Losing the prefix means paying full price every turn.

## Change

One line in `ToolResolverService.dedupToolsByName()`:

```kotlin
.sortedWith(compareBy { it.name })
```

### Why this location

`dedupToolsByName()` is the mandatory convergence point for every tool source: `resolveToolsForRun()` calls it internally, and `AgentServiceImpl.resolveAgentDefinition()` calls it again over `baseTools + delegationTools + exchangeTools + queryUserTools`. Sorting in either caller alone would have missed the other tier; sorting in both would have been redundant and fragile. The existing KDoc already frames this method as the shared reconciliation point.

### Two details worth a reviewer's attention

**The sort is applied after deduplication, never before.** This is the trap in this change: sorting first would alter *which* duplicate is kept — the first alphabetically instead of the first encountered — silently changing selection semantics. Dedup behaviour is strictly unchanged, and the KDoc now documents the trap explicitly.

**The sort is binary, not locale-sensitive.** `compareBy` over `String.compareTo` compares Unicode code points. A JVM running under a Turkish or French locale with a `Collator` would produce a different order — an order varying by deployment environment would cause exactly the instability this PR removes. A dedicated test pins this.

## Verified: no caller depends on existing order

- `AgentAdvanced` resolves tools strictly by name (`firstOrNull { it.name == intention.toolName }`), never by position
- `AgentSimple` maps tools to `toolCallbacks` in list order, but position confers no priority — the model distinguishes tools by name in the schema
- The `baseTools + delegationTools + exchangeTools + queryUserTools` precedence operates at the *duplicate selection* level, which the sort does not touch since it runs afterwards

## Tests

7 new tests in `ToolResolverServiceSpec` (23 total). Full suite green: ~2,400 tests, 0 failures, no regression.

The test that justifies the exercise is **"produces the same ordered list regardless of the input order"**. The others verify behaviour; that one locks the *property*. Current stability rested entirely on `groupBy` preserving insertion order — without this test a future refactor would break it silently, and nobody would notice before reading an invoice.

Also covered: alphabetical ordering, first-encountered-wins on collision (non-regression), empty and single-element lists, and locale-independent binary ordering.

## Scope

Deliberately **not** here:
- No `OpenAiProperties` mirroring `AnthropicProperties` — it would activate nothing, OpenAI caching is already on
- No `prompt_cache_key` or GPT-5.6 explicit breakpoints — both need measurement before they can be judged
- MCP tool name sanitisation and inference parameter handling remain open on #1317

## Known remaining gap

`cached_tokens` (`usage.prompt_tokens_details.cached_tokens` for OpenAI, `cache_read_input_tokens` / `cache_creation_input_tokens` for Anthropic) is **not read, logged or persisted anywhere**. This PR maximises cache reuse but its effect cannot currently be measured — the same blind spot as the provider errors in #1318, except this one is paid in invoices rather than investigation time. It applies equally to Anthropic, whose caching we enabled without ever measuring it. Worth a separate issue.

### DIFF
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
@@ -80,8 +80,7 @@ class ToolResolverService(
                     logger.warn { "[ToolResolver] Tool name conflict: '$name' present ${duplicates.size} times, keeping the first one." }
                 }
                 duplicates.first()
-            }
-            .sortedWith(compareBy { it.name })
+            }.sortedBy { it.name }
 
     private fun extractTools(
         allowedNames: List<String>?,
--- a/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
+++ b/agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
@@ -58,7 +58,18 @@ class ToolResolverService(
     }
 
     /**
-     * De-duplicate tools by name, keeping the first occurrence and warning on each conflict.
+     * De-duplicate tools by name, keeping the first occurrence and warning on each conflict,
+     * then sort the result by name using binary (locale-independent) ordering.
+     *
+     * Sorting after deduplication is intentional: sorting before would change which duplicate
+     * is kept (the first alphabetically rather than the first encountered), breaking the
+     * existing selection semantics.
+     *
+     * The binary sort (compareBy with String.compareTo using ordinal comparison) is independent
+     * of the JVM default locale, ensuring a stable, reproducible order across deployments.
+     * This matters for prompt-cache stability: OpenAI's prefix cache is invalidated by any
+     * change in tool name order, so a deterministic order maximises cache reuse.
+     *
      * Shared so every tool source (resolver, delegation, exchange) reconciles collisions identically.
      */
     fun dedupToolsByName(tools: List<StandardTool<*>>): List<StandardTool<*>> =
@@ -70,6 +81,7 @@ class ToolResolverService(
                 }
                 duplicates.first()
             }
+            .sortedWith(compareBy { it.name })
 
     private fun extractTools(
         allowedNames: List<String>?,
--- a/agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/tool/ToolResolverServiceSpec.kt
+++ b/agentos/agentos-service/src/test/kotlin/io/whozoss/agentos/tool/ToolResolverServiceSpec.kt
@@ -317,6 +317,127 @@ class ToolResolverServiceSpec :
         }
 
         // -------------------------------------------------------------------------
+        // dedupToolsByName — ordering guarantee (prompt-cache stability)
+        // -------------------------------------------------------------------------
+
+        "dedupToolsByName returns tools in alphabetical order by name" {
+            // Tools intentionally given in reverse-alphabetical order to detect sort.
+            val tools =
+                listOf(
+                    makeTool("Zebra"),
+                    makeTool("Mango"),
+                    makeTool("Apple"),
+                )
+
+            val result = buildService().dedupToolsByName(tools)
+
+            result.map { it.name } shouldBe listOf("Apple", "Mango", "Zebra")
+        }
+
+        "dedupToolsByName produces the same ordered list regardless of the input order" {
+            // Core regression guard: two calls with the same logical tool set but different
+            // input orderings must produce identical output. This is the property that
+            // prevents OpenAI prompt-cache invalidation when Neo4j changes its return order.
+            val service = buildService()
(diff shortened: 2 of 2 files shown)


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/tool/ToolResolverService.kt
```diff
@@ -70,6 +81,7 @@ class ToolResolverService(
                 }
                 duplicates.first()
             }
+            .sortedWith(compareBy { it.name })
```
frederic-delsert-whoz: .sortedBy { it.name } ?

vincent-audibert-whoz: ok

vincent-couturier-whoz (approved):

Issue #1290, opened 2026-09-07
feat(agentos-ui): visualize case delegation timeline


selimbensenouciep-prog:
## Goal

Make the progression of a case and its delegated work understandable without reconstructing it from long conversations.

## Scope

- Add a timeline/Gantt view linked to the current case and its delegated sub-cases.
- Show the progression of delegated work and where execution is blocked.
- Make the details of each delegated case accessible, including its input, output, and what happened during execution.
- Build on existing AgentOS orchestration and delegation. Introducing scripted workflow steps or gates is deferred.

## Open questions

- What is the final presentation of the timeline and its execution details?
- Which execution data is already available to support this view?

Acceptance criteria have not yet been agreed.

PR #1324, opened 2026-09-09
feat: #1290 add delegation lifecycle events

benjamin-valdes-whoz:
Adds durable and correlated sub-case lifecycle events for agent delegations. Each delegation now exposes start and completion events, linked through delegationId and the parent toolRequestId, with persistence, structured outcomes, and protection against re-triggering the parent runtime. Includes Kotlin and TypeScript models and focused lifecycle/persistence tests.


agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/caseEvent/CaseEvent.kt
```diff
@@ -455,6 +459,47 @@ data class ConfirmationResolvedEvent(
...
+enum class SubCaseOutcome {
+    SUCCESS,
+    WAITING_USER,
+    ERROR,
+    TIMEOUT,
+    KILLED,
+}
+
+/** Durable parent-case observation emitted after a delegated sub-case starts or resumes. */
+data class SubCaseStartedEvent(
```
vincent-couturier-whoz: question: why another event in addition to the one existing in the sub case ?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolContext.kt
```diff
@@ -36,4 +36,6 @@ data class ToolContext(
     val caseEvents: List<CaseEvent>,
     val agentName: String? = null,
     val credentialProvider: CredentialProvider? = null,
+    /** Identifier of the ToolRequestEvent that caused this invocation, when available. */
+    val toolRequestId: String? = null,
```
vincent-couturier-whoz: question: why this change here ? which relationship with delegation ?

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/caseEvent/CaseEvent.kt
benjamin-valdes-whoz: The sub-case events remain the detailed source of truth for its own execution. These additional events are intentionally durable observations on the parent case: they let parent-case consumers — especially the UI/SSE stream — follow a delegation and correlate its start/end with the toolRequestId, without having to navigate into each sub-case or infer a final state from its internal events.

They do not re-enter the parent runtime: emitParentEvent only persists and broadcasts them as observation events. If we do not want to expose delegation lifecycle at the parent level, then we should indeed remove this layer and have the client or a dedicated read model perform the aggregation instead.

agentos/agentos-sdk/src/main/kotlin/io/whozoss/agentos/sdk/tool/ToolContext.kt
benjamin-valdes-whoz: his is the correlation link between the tool invocation and the delegation. The DelegationTool receives the toolRequestId of the ToolRequestEvent that triggered it, then copies it into SubCaseStartedEvent and SubCaseFinishedEvent. This lets us unambiguously associate a delegation — and its sub-case — with the parent tool request, especially when multiple delegations run in parallel.

I propagate it through both AgentSimple and AgentAdvanced execution paths, including the post-confirmation path, so the correlation is not lost depending on how the tool is executed.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -48,23 +53,26 @@ class CaseEventSseController(
...
+     * `includePreviousEvents` is non-null and defaults explicitly to true: a new or
+     * reconnecting client receives durable history before buffered/live events.
      */
     @Operation(
         tags = ["sse"],
         summary = "Stream case events via SSE",
         description =
-            "Server-Sent Events stream emitting all events generated during case execution. " +
-                "Use the browser EventSource API to consume this endpoint, not a regular HTTP client.",
+            "BREAKING: every SSE frame uses the stable event name 'case-event'. " +
```
vincent-couturier-whoz: question: Why the prefix "BREAKING" here ?

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -89,26 +96,16 @@ class CaseEventSseController(
...
-        // Heartbeat: periodically send an SSE comment frame so that a client
-        // disconnect is detected even when the case is IDLE and emits no events.
-        // Without this, the collector coroutine above stays suspended on collect()
-        // indefinitely, keeping the SharedFlow subscriber alive and blocking eviction.
-        startHeartbeatJob(
-            scope = scope,
-            emitter = emitter,
-            caseId = caseId,
-        )
+        startHeartbeatJob(scope = scope, emitter = emitter, caseId = caseId)
```
vincent-couturier-whoz: suggestion: I often hate comments but here it's really helpful. The same for cancel on line 104.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/caseEvent/CaseEventSseController.kt
```diff
@@ -118,85 +115,111 @@ class CaseEventSseController(
...
+        includePreviousEvents: Boolean,
         caseId: UUID,
         emitter: SseEmitter,
     ) {
+        val activeCase = caseService.findActiveRuntime(caseId)
+        val liveEvents = Channel<CaseEvent>(LIVE_BUFFER_CAPACITY)
+        val invalidated = AtomicBoolean(false)
+
+        fun invalidateForSaturation(cause: Throwable) {
+            if (!invalidated.compareAndSet(false, true)) return
```
vincent-couturier-whoz: suggestion: use if/else structure to avoid early return and simplify the reading of the code and avoid bad surprise due to special return behavior in kotlin

Issue #1133, opened 2026-07-13
feat: add lifecycle cleanup hook to StandardTool for resource management


vincent-audibert-whoz:
## Context

Discovered during review of PR #1119 (MCP HTTP tool provider). The AgentOS framework currently follows a "fire-and-forget" pattern for tools: `ToolResolverService.resolveToolsForRun()` creates them, agents use them during a run, and they are simply dereferenced when the run ends. There is no explicit cleanup mechanism.

This was acceptable when all tools were stateless or pooled (stdio MCP connections are managed by `McpConnectionPool`), but the introduction of `HttpMcpConnection` in #1119 breaks that assumption — HTTP connections hold a JDK `HttpClient` with its internal thread pool and an active MCP session on the remote server. On the success path, nobody calls `close()`, leading to resource leaks under sustained usage.

## Problem

- `StandardTool` (SDK interface) has no lifecycle hook (`close()`, `cleanup()`, `dispose()`)
- `AgentAdvanced.run()` and `AgentSimple.run()` do not call any teardown on tools after the run completes
- `ToolResolverService` creates tools but has no mechanism to track or clean them up
- Any tool that holds resources (HTTP connections, file handles, API sessions) will leak when the agent run ends

Currently affected: `McpTool` wrapping `HttpMcpConnection` — the connection has a `close()` method (via `McpConnectionPort`) but no caller on the success path.

## Proposed approach

1. **Add an optional `close()` method to `StandardTool`** with a default no-op, so existing tools are unaffected:
   ```kotlin
   interface StandardTool<T> {
       // ... existing methods ...
       
       /** Called when the agent run completes. Override to release resources. */
       fun close() {}
   }
   ```

2. **Call `close()` in agent run teardown** — both `AgentAdvanced.run()` and `AgentSimple.run()` should invoke `close()` on all tools in a `finally` block, swallowing individual failures to avoid masking the original result.

3. **Implement `close()` in `McpTool`** to delegate to `McpConnectionPort.close()` for HTTP connections (pooled connections already handle their own lifecycle via the pool).

## Scope

- `agentos/agentos-sdk` — `StandardTool` interface
- `agentos/agentos-service` — `AgentAdvanced`, `AgentSimple` run methods
- `agentos/agentos-mcp-plugin` — `McpTool` implementation

## Related

- PR #1119 — MCP HTTP tool provider (where the gap was identified)
- Review comments by @selimbensenouciep-prog and @vincent-couturier-whoz on connection lifecycle

Issue #1053, opened 2026-06-25
AgentOS HTTP integration plugin


vincent-audibert-whoz:
## Context

Coday currently has a generic HTTP integration that allows agents to make GET requests to arbitrary URLs with configured headers/auth. This is used for read-only integrations like Zendesk, and is a powerful general-purpose tool for knowledge workers.

AgentOS needs an equivalent plugin so that existing Coday HTTP-based integrations can be ported without requiring a dedicated plugin per service.

## Goal

Implement a generic HTTP integration plugin for AgentOS that:
- Allows agents to perform HTTP requests (GET at minimum, POST optional) to configured endpoints
- Supports configurable base URL, headers, and authentication parameters
- Replaces the need for custom plugins for simple read-only integrations (Zendesk, etc.)

## Design notes

From the Coday convergence discussion:
- The HTTP integration is the right path for porting simple read-only integrations (Zendesk, Basecamp read-only) — they become pure configuration once the plugin exists
- Authentication handling (API keys, OAuth) should be designed in coordination with the auth mechanism issue (see linked issue), since MCP remote will define the pattern first
- The generic HTTP plugin should be implemented **after** the auth mechanism is established, to reuse the same patterns

## Acceptance criteria

- [ ] A namespace admin can configure an HTTP integration with a base URL and auth parameters
- [ ] Agents with the HTTP integration can make requests to the configured endpoints
- [ ] Sensitive parameters (API keys, tokens) are masked in the UI and not logged
- [ ] At least one existing Coday HTTP integration (e.g. Zendesk) can be replicated via configuration alone

## Related

- Requires auth mechanism issue (to be linked)
- MCP remote plugin (#1018) defines the auth pattern to follow

## Technical update — 2026-09-07

The HTTP integration concept needs to be reconsidered rather than ported unchanged from TypeScript. The direction to explore is defining integrations from an API specification, such as OpenAPI, describing endpoints, payloads and parameters. This is distinct from general web browsing.

Google Drive access (#344) takes priority: a sufficiently reliable MCP integration is the first option, with HTTP/OpenAPI as an alternative if needed. Related porting ticket: #1073.

### Open questions
- Which specification format and integration approach should be used?

PR #1325, opened 2026-09-09
feat(agentos): #1053 preview the tools of an IntegrationConfig from the admin form

selimbensenouciep-prog:
## Merge order (four related PRs)

1. **#1314** service prerequisites (static Auth Setting credentials delivered to plugins, user-scope policy). Merge first: the three others rely on it functionally.
2. **#1315** MCP_HTTP hardening. Independent code-wise; merge after #1314 so the credentials it starts receiving are mapped with the right `Authorization` scheme.
3. **#1328** `HTTP_API` plugin. Compiles and is tested on master alone, but static credentials reach it end to end only with #1314.
4. **#1325 (this PR)** tool preview. Stacked on the branch of #1314; once #1314 is merged its base moves to `master`, then merge it.

The four branches were merged together in a local integration branch without any conflict, and that build was used for the live Google Drive test recorded in #1328.

## What this PR brings

Until now, the only way to know which tools an integration produces was to bind it to an agent and open the agent definition preview. A **Preview tools** action on the integration edit form now shows the resulting tools directly: their names, descriptions and input schemas, the plugin's summary line, or the error if the document cannot be loaded.

Stacked on #1314: it relies on the static-credential delivery and the scope policy introduced there, so #1314 must be merged first.

## What changed

- `POST /api/integration-configs/{id}/preview-tools`: requires `WRITE` on the config, `READ` on the namespace when `namespaceId` is given, Super Admin for platform rows. Resolves the stored row as is (overlay merge not applied), for the calling user, without an agent or a case. Nothing is persisted. 422 when no plugin is loaded for the type; a plugin failure is returned as an error message with an empty tool list. No credential value is ever returned.
- `CredentialProviderFactory` extracted from `AgentServiceImpl` so the preview resolves credentials exactly like an agent run (OAuth only with a live case, otherwise direct lookup; static secrets through `StaticCredentialFactory`). Agent behaviour is unchanged.
- New `preview-describe-namespace-timeout-ms` property (default 5 s) bounding the plugin's summary line.
- UI: "Preview tools" button and results panel on the integration edit form, with a small state service; hidden for platform rows without a namespace context.
- Second commit: regenerated OpenAPI spec and Angular client. It also catches up client drift that predates this branch (`countUnreadCase`, `markCaseRead`, `Case.readAt` / `scheduledPromptId`, `UnreadCountResponse`, prompt-translate doc strings); review it as generated output.

## How it was verified

- `./gradlew :agentos-service:test`: 187 spec classes, 2878 tests, 0 failures (controller unit and MVC tests cover the permission wiring and the error cases).
- `pnpm exec nx run agentos-service:check-openapi-spec` green on the committed state; `nx run-many --targets=lint` for `agentos-ui` and `agentos-api-client` green; `nx run agentos-ui:test` 344 tests green; `nx run client:build` green.
- Not run: no click-through against a live server.

## Risks

- Each preview of an `MCP_HTTP` config opens a connection that is not closed on the success path, like the existing agent definition preview (#1133).
- The preview shows the stored row, not the merged overlay result.

## Related

- #1314 (base), #1328 (`HTTP_API` plugin whose configs this preview is most useful for), #1315. Issue #1053.

https://claude.ai/code/session_01XNX3YCYNeZU95Fp4eDeGvH


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/auth/CredentialProviderFactory.kt
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * The factory yields `null` for every name when [userId] is null: credentials are always
+     * user-scoped, so a run without a user gets no provider at all.
+     *
+     * [caseId] and [emitEvent] make the run interactive ([OAuthFlowService.resolveOAuthCredential]
+     * emits its `QuestionEvent` into the case); when either is absent, OAuth types resolve through
+     * the direct lookup only. An interactive run must name its agent ([agentName] is a precondition
+     * then); a case-less caller such as the tool preview passes all three as null.
+     */
+    fun forRun(
```
vincent-couturier-whoz, line 48: suggestion (non-blocking): two named entry points would restore the invariant the compiler used to enforce

Before the extraction, `agentName` came from `agentConfig.name` — non-nullable by
construction. The unified signature makes it nullable to accommodate the preview, and
the invariant is re-established at runtime by `require`. That trades a compile-time
guarantee for a runtime check.

The preview call site also does not say *why* the three arguments are null:

    credentialProviderFactory.forRun(namespaceId, user.id, null, null, null)(authSettingName)

Two facades over a shared private builder would fix both:

    fun forAgentRun(
        namespaceId: UUID,
        userId: UUID?,
        caseId: UUID?,
        agentName: String,
        emitEvent: ((CaseEvent) -> CaseEvent)?,
    ): (String) -> CredentialProvider?

    fun forPreview(namespaceId: UUID, userId: UUID): (String) -> CredentialProvider?

`caseId` and `emitEvent` stay nullable on the run side (webhooks, oneshot runs), but
`agentName` does not. The `require` disappears, and `forPreview(namespaceId, user.id)`
reads on its own.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigToolPreviewService.kt
```diff
@@ -0,0 +1,148 @@
...
+                credentialProviderFactory.forRun(
+                    namespaceId = namespaceId,
+                    userId = user.id,
+                    caseId = null,
+                    agentName = null,
+                    emitEvent = null,
+                )(authSettingName)
+            }
+        val toolContext = baseContext.copy(credentialProvider = credentialProvider)
+        val (tools, error) = provideToolsSafely(plugin, config, toolContext)
```
vincent-couturier-whoz, line 74: issue (blocking): `provideTools` has no server-side time bound, unlike `describeNamespace`

The protection is on the wrong call. `describeNamespace` is the cheap one — a cached
catalogue read with no network access for `HTTP_API` — and it is wrapped in
`withTimeoutOrNull`. `provideTools` is the expensive one: for `MCP_HTTP`,
`HttpMcpConnection.connect` performs a full `initialize()` followed by `listTools()`,
and it runs unguarded on the Tomcat request thread.

`timeoutSeconds` and `toolCallTimeoutSeconds` are config fields under admin control,
validated only as `> 0` (`McpConfigParser.validateTimeouts`). Setting
`timeoutSeconds: 600` pins a Tomcat thread for ten minutes per click, on an endpoint
designed to be clicked repeatedly while tuning a config.

Proposal: wrap `provideTools` in `withTimeoutOrNull` with its own property
(`preview-provide-tools-timeout-ms`), independent of whatever the config declares.
Same honest caveat as `describeNamespace`: cooperative cancellation does not reclaim
a thread blocked in I/O, but it does return the HTTP request to the caller, which is
what matters for the UI. It also flattens the timing oracle described in the error
message comment.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigToolPreviewService.kt
```diff
@@ -0,0 +1,148 @@
...
+                "[ToolPreview] describeNamespace failed for integration '${config.name}' " +
+                    "(type '${config.integrationType}')"
+            }
+            null
+        }
+
+    /** `ExceptionClass: message`, shown verbatim by the UI; never renders a `null` literal. */
+    private fun describeFailure(e: Exception): String {
+        val type = e::class.simpleName ?: e::class.java.name
+        val message = e.message ?: "no message"
```
vincent-couturier-whoz, line 143: issue (non-blocking): the underlying exception message is surfaced verbatim in `error`

`HttpMcpConnection` throws:

    McpConnectionException("Failed to initialise MCP HTTP session for '$url': ${e.message}", e)

That inner message passes through `describeFailure` untouched into the DTO's `error`
field and is rendered as-is by the UI. It turns a blind SSRF into a semi-blind one:
the caller learns whether the host answered and how it failed.

Not enough for full read primitives, but enough for internal host and port sweeping —
and it composes with the absence of DNS resolution at parse time (see the comment on
`plugin-system.md`).

Proposal: keep the exception type, bound the message (max length, or an allow-list of
types whose message is considered safe to surface).

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/auth/CredentialProviderFactory.kt
```diff
@@ -0,0 +1,221 @@
...
+            } else {
+                null
+            }
+        return { authSettingName ->
+            logger.debug { "CredentialProvider invoked for '$authSettingName'" }
+            if (userId == null) {
+                logger.debug { "CredentialProvider for '$authSettingName': no userId in context, skipping" }
+                null
+            } else {
+                val scopedAuthService = authServiceFactory.create(namespaceId, userId)
```
vincent-couturier-whoz, line 68: question (non-blocking): is `create` a plain object construction or a request-scoped bean?

The `AuthService` KDoc states "Implementations are request-scoped: created per tool
invocation". Here the call happens at the name-to-provider factory level, i.e. before
the provider is invoked — and the provider itself may run on a `Dispatchers.IO` thread,
outside any HTTP request.

If it is a plain construction, no issue at all. If it is a request-scoped bean, the
behaviour is fragile. Nothing changes with this PR, but the extraction makes the
question visible and it seems worth confirming.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/integrationConfig/IntegrationConfigToolPreviewService.kt
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Resolves the tools an [IntegrationConfig] yields for a user, without an agent and without a case,
+ * so an admin can check a config right after saving it.
+ *
+ * The stored row is used as is: the 4-tier overlay merge that an agent run applies
+ * (`IntegrationConfigService.findEffective`) is NOT applied here, so a user-scoped override of the
+ * same name is not folded in. The plugin is invoked exactly like `ToolResolverService` does for a
+ * run — `provideTools(parameters, name, context)` with a `credentialProvider` when an auth setting
+ * is bound — which means the preview inherits the plugins' connection behaviour: `MCP_HTTP` opens a
```
vincent-couturier-whoz, line 24: thought (non-blocking): the preview *cannot* close what it opens — the SDK is missing the abstraction

The risk is well documented; I want to record why it is not fixable here.
`provideTools` returns a bare `List<StandardTool<*>>` with no lifecycle handle, and
`McpConnectionPort` / `McpTool` live in the PF4J plugin classloader —
`agentos-service` can neither reference them nor `instanceof` them. Only
`io.whozoss.agentos.sdk.*` types cross the boundary.

So this is a missing abstraction rather than an oversight. A direction that would fix
#1133 and every caller at once:

    interface ToolSet : AutoCloseable {
        val tools: List<StandardTool<*>>
        override fun close() {}   // no-op default: resource-free plugins are unaffected
    }

The preview would become `plugin.provideTools(...).use { ... }` and the run manager
would close deterministically. SDK contract change, so out of scope here — but worth a
dedicated issue that subsumes #1133.

Two details worth recording: an unclosed connection also retains an MCP session on the
*remote* server (`Mcp-Session-Id`), so we leak resources at the third party too; and
the JDK `HttpClient` is only GC-reclaimable from Java 21 onwards, while #1328 notes the
service Dockerfile still pins JDK 17.

agentos/docs/plugin-system.md
```diff
@@ -110,6 +110,41 @@ host they control, so those types are refused in both user scopes at the API edg
...
+- **No overlay merge**: the stored row is previewed as is; the 4-tier overlay an agent run applies
+  (`IntegrationConfigService.findEffective`) and the agent allowlist are not applied.
+- **No persistence**: the credential provider is built for the caller without a case, so OAuth types
+  resolve through the direct lookup only and never start an interactive flow. Static types synthesise
+  their credential in memory as during a run.
+- **`POST`, not `GET`**: nothing is persisted, but the call resolves the caller's credential and lets
+  the plugin open outbound connections with it, so it must never be prefetched or cached by a browser
+  or an intermediary the way a safe `GET` such as `GET /{id}/export` may be.
+- **422** when no plugin is loaded for the config's integration type.
+- The plugin is invoked exactly like `ToolResolverService` does for a run, so the preview inherits
```
vincent-couturier-whoz, line 145: issue (non-blocking): `MCP_HTTP` and `HTTP_API` do not offer the same SSRF guarantee

`McpConfigParser.validateUrl` (#1315) is thorough — scheme, userinfo, loopback,
link-local (so `169.254.169.254`), RFC 1918, IPv6 unique-local `fc00::/7`, CGNAT
`100.64.0.0/10`, wildcard, with the IPv6 zone id stripped before classification. But
the code says it itself:

> Only literal IPs are inspected — no DNS resolution happens at parse time.

So `https://metadata.attacker.com` resolving to `169.254.169.254` passes validation.

Meanwhile #1328 states for `HTTP_API`: "checked at parse time and again through a
filtering DNS resolver". The filtering resolver already exists in this stack.

Out of scope for this PR, but the preview makes the gap reachable in one click: either
port the resolver to `MCP_HTTP`, or document the gap as accepted.

Issue #1136, opened 2026-07-13
fix: validate MCP HTTP URL and warn on plaintext Bearer token transmission


vincent-audibert-whoz:
## Context

Raised during review of PR #1119 by @selimbensenouciep-prog. The `McpConfigParser.validateHttp` method only checks that `authToken` is not blank, but never validates the URL itself. A malformed or insecure URL only fails at connect time with a generic error.

More critically, nothing prevents configuring a plain `http://` endpoint. When a Bearer token is configured (static `authToken` today, per-user OAuth tokens once #1095 lands), it is sent in cleartext over the wire.

## Problem

- No URL validation at config parse time — malformed URLs produce opaque errors at connect time
- No scheme enforcement — `http://` endpoints silently transmit credentials in cleartext
- The JSON schema declares `url` as a plain `"type": "string"` with no format constraint
- `McpConfigParser.validateHttp` validates `authToken` but ignores `url` beyond null-checking

## Proposed approach

In `McpConfigParser.validateHttp`:

1. **Parse the URL with `java.net.URI`** and require an absolute URL with `http` or `https` scheme (reject relative paths, `file://`, etc.)
2. **Warn on `http://` with authentication** — when a `authToken` is configured (or once `credentialProvider` is wired), log a clear `WARN` that credentials will be transmitted in plaintext. Rejecting `http://` outright would be safer but may break legitimate local dev setups (`http://localhost`).
3. **Add `"format": "uri"` to the JSON schema** for the `url` property — this helps config UIs provide validation feedback.

## Scope

- `agentos/agentos-mcp-plugin` — `McpConfigParser`, `McpHttpToolProvider` JSON schema

## Related

- PR #1119 — MCP HTTP tool provider
- #1095 — AuthSetting & CredentialProvider wiring (amplifies the risk: per-user OAuth tokens over plaintext HTTP)

Issue #1104, opened 2026-07-07
discuss: user-level IntegrationConfig overrides and the privileged execution problem (MCP_STDIO)


vincent-audibert-whoz:
## Context

PR #837 introduced user-level overlays for `IntegrationConfig`, allowing any namespace member to create their own overrides (`userOnNs` or `userGlobal`) without admin intervention. The permission model was intentionally permissive at the user scope: *"Anyone can CRUD their own overrides"*.

PR #815 then introduced `MCP_STDIO` as an integration type. Its trust boundary documentation (added during review) explicitly states:

> Configuring an `MCP_STDIO` integration is equivalent to arbitrary code execution on the host. No allow-list is enforced at the plugin level. Authorization must be delegated to the integration management layer.

The problem: **the integration management layer currently has no mechanism to enforce that delegation**. A simple namespace member can today create a `userOnNs` override of type `MCP_STDIO` with an arbitrary `command`, bypassing any admin validation.

## Why this is hard

`MCP_STDIO` is categorically different from other integration types:

| Integration type | What a user-level override can do |
|---|---|
| `DATETIME` | Change timezone |
| `LLM` provider | Use a different model / API key |
| `MCP_STDIO` | Spawn an arbitrary child process on the server |

Blocking user overrides entirely would regress the legitimate use case from #837 (users personalizing their tool configs). But allowing `MCP_STDIO` user overrides means any authenticated user can execute arbitrary code on the host — a risk that's hard to contain once the surface is open.

## Questions to explore

- **Integration type classification**: Should integration types carry a `privilegeLevel` (e.g. `SAFE` vs `PRIVILEGED`) that the permission layer can use to gate user-level overrides?
- **Namespace admin gate**: Should `PRIVILEGED` integration types require at minimum namespace-admin role to create at any scope (namespace, userOnNs, userGlobal)?
- **Separate concept**: Is `MCP_STDIO` fundamentally a different kind of entity — closer to a "tool deployment" requiring an explicit admin approval workflow — rather than a plain `IntegrationConfig`?
- **User-global scope**: A `userGlobal` override of `MCP_STDIO` would apply to all namespaces the user has access to, amplifying the risk. Should `userGlobal` overrides be restricted to `SAFE` types regardless of admin status?
- **Audit trail**: Even if we restrict creation, should there be visibility for admins over which `MCP_STDIO` configs exist and who created them?

## Related

- PR #837 — user-level overlays implementation (where the permission model was defined)
[...]

Issue #1073, opened 2026-06-30
Port HTTP integration from typescript to AgentOS plugin


vincent-audibert-whoz:
## Technical update — 2026-09-07

See #1053 for the HTTP/OpenAPI design direction: reconsider the integration concept rather than perform a direct TypeScript port. Google Drive access is the immediate priority, tracked in #344.

Issue #735, opened 2026-04-07
AgentOS: trace security concern on IntegrationConfig.parameters (sensitive data handling)


cetautomatix-coday:
## Context

`IntegrationConfig` stores integration parameters as a `JsonNode?` field, which may include sensitive data such as API keys, tokens, or credentials (e.g. for JIRA, GitHub, Slack integrations).

## Problem

There is currently no special handling for sensitive fields in `parameters`. They are stored as plain JSON, returned as-is by the API, and logged or traced without redaction. This is a security concern flagged during PR review.

## Expected

Analyze and implement appropriate handling for sensitive data in `IntegrationConfig.parameters`:
- Identify which fields are sensitive (via JSON Schema `format: "password"` or a dedicated marker)
- Ensure sensitive values are not returned in plain text by the API (masking or omission)
- Ensure sensitive values are not logged
- Consider encryption at rest if applicable

## References

- PR #729 review comment: https://github.com/whoz-oss/coday/pull/729#discussion_r3047140061

PR #1328, opened 2026-09-09
feat(agentos): #1053 HTTP_API plugin exposing OpenAPI-described APIs as agent tools

selimbensenouciep-prog:
## Merge order (four related PRs)

1. **#1314** service prerequisites (static Auth Setting credentials delivered to plugins, user-scope policy). Merge first: the three others rely on it functionally.
2. **#1315** MCP_HTTP hardening. Independent code-wise; merge after #1314 so the credentials it starts receiving are mapped with the right `Authorization` scheme.
3. **#1328 (this PR)** `HTTP_API` plugin. Compiles and is tested on master alone, but static credentials reach it end to end only with #1314.
4. **#1325** tool preview. Stacked on the branch of #1314; once #1314 is merged its base moves to `master`, then merge it.

The four branches were merged together in a local integration branch without any conflict, and that build was used for the live Google Drive test recorded in #1328.

## What this PR brings

The feature asked for in #1053: an agent can use any service that exposes an HTTP API without a dedicated connector. The admin provides an OpenAPI 3.x document (a published URL, a file, or a few hand-written lines for an API without a spec), a base URL and an Auth Setting; the new `HTTP_API` integration type turns each selected operation into an agent tool.

## What changed

- New PF4J plugin `agentos/agentos-http-plugin`, integration type `HTTP_API`, registered in Gradle, the release workflow and the Coday runtime. Configured through the existing integration form (schema-driven) or a YAML file.
- One `StandardTool` per curated operation (`tags`, path prefixes, `operationId` globs, per-operation overrides, `maxTools` cap), named `<INTEGRATION>__<operation>`. Reads only by default; writes exist only with `allowMutations: true`, are prefixed `[WRITE]` and require confirmation (`EVERY_TIME`, honoured by AgentAdvanced).
- Authentication from the bound Auth Setting, mapped by credential type: `Bearer` for OAuth and bearer tokens, RFC 7617 `Basic` (Zendesk style `email/token`), API keys in a header, a query parameter or a bearer, optionally taken from the document's `securitySchemes`. No usable credential means the call fails fast with `AUTH_MISSING`, never an anonymous call.
- OkHttp runtime owned by the plugin lifecycle: HTTPS only, no loopback / private / link-local / CGNAT targets (checked at parse time and again through a filtering DNS resolver), no redirects, per-call timeout, per-config concurrency limit, bounded body reads. Responses can be filtered (`keepPaths` / `ignorePaths`), rendered as JSON or YAML and capped with a truncation marker. Stable error types (`UNAUTHORIZED`, `RATE_LIMITED`, `URL_POLICY_REJECTED`, ...). Tool metadata carries the path only, never a query string or a header.
- Catalogue cache keyed by a hash of the config (never by credential material), conditional refresh with ETag, stale-on-failure; the plugin's `describeNamespace` line reports the operation count or the last load failure in the agent definition preview.
- Documentation: `agentos/docs/http-api-integration.md` (config keys, auth mapping, curation, URL policy, mutations, Zendesk and hand-written examples). `scripts/http-integration-to-openapi.ts` converts a Coday TypeScript HTTP integration into an inline OpenAPI document.

## How it was verified

- `./gradlew :agentos-http-plugin:test :agentos-http-plugin:jar`: 423 tests, 0 failures. Execution tests run against a local JDK `HttpServer`; a dedicated spec asserts that a known secret never appears in outputs, metadata or logs.
- `./gradlew deployPlugins`: jar deployed and inspected (manifest, `extensions.idx`, no Jackson / OkHttp / Kotlin classes bundled).
- `pnpm exec nx test scripts`: converter tests green.
- Live end-to-end run (2026-09-09, local isolated runtime built from this branch merged with #1314, #1315 and #1325, database copied from a developer install, Claude Sonnet as the model): an `HTTP_API` integration pointed at the public Google Drive v3 OpenAPI document (apis.guru, 244 KB, `includePathPrefixes` `/files` + `/about`, `allowMutations` off) was curated to 3 `GET` tools; the bound `OAUTH_REGISTERED` Auth Setting triggered the interactive Google consent through the standard UI popup and callback; the agent then called `drive_files_list` (200, 2.5 KB, shaped to the `keepPaths` fields, rendered as YAML) and `drive_files_get` (200) and answered. Logs carried only method, host, path, status, duration and size; tool metadata carried the path only. Two operations with multipart bodies were reported as skipped, as designed.
- Not run: the Zendesk (Basic auth) recipe against a real tenant, an operation with a `oneOf` parameter, OpenAI as the provider.

## Risks and follow-ups

- Private-network APIs are refused by design (product decision: public SaaS only).
- Writes under the default orchestrator (`AgentSimple`) execute without confirmation; the per-agent tool allowlist is the documented guard.
- The credential is resolved once per run (same as `MCP_HTTP`); long runs with short-lived OAuth tokens get `UNAUTHORIZED` until #1198.
- Minor items left as noted in the code review: `User-Agent` in `defaultHeaders` is overridden; `spec.maxBytes` is not part of the cache key; a second `start()` without `stop()` would leak the client.
- The plugin needs `deployPlugins` and a service restart; the service Dockerfile still pins JDK 17 against a JDK 25 toolchain.

## Related pull requests

- #1314 (service): delivers admin-entered static secrets to plugins; this plugin compiles and is tested on master alone, but the static-credential path works end to end only once #1314 is merged.
- #1315 (MCP plugin): same credential-type mapping and URL validation for `MCP_HTTP`.
- #1325 (stacked on #1314): preview the tools a config yields before binding an agent.
- Issues: #1053, #1073, #1136, #1133, #735, #1104.

https://claude.ai/code/session_01XNX3YCYNeZU95Fp4eDeGvH


thomas-martin-whoz (approved):
Ok pour la partie commune, pas regardé le plugin HTTP

PR #1337, opened 2026-09-15
chore(release): disable useCommitScope and remove obsolete adjustSemverBumpsForZeroMajorVersion

mathieu-de-armey-whoz:
Nx silently downgrades `feat(agentos):` to a patch bump because `agentos` matches no Nx project name.

Setting `conventionalCommits.useCommitScope: false` disables scope-based bump restriction — safe since all projects are in a single fixed release group where scope restriction has no effect anyway.

Also removes `adjustSemverBumpsForZeroMajorVersion: false` which was a no-op since the project is on `1.x`.


nx.json
```diff
@@ -85,8 +88,7 @@
       "git": {
         "commit": false,
         "tag": false
-      },
-        "adjustSemverBumpsForZeroMajorVersion": false
```
vincent-couturier-whoz, line 89: question: why do you remove this parameter ?

vincent-couturier-whoz (approved):

nx.json
mathieu-de-armey-whoz: before nx 23 (which we migrated to recently) `adjustSemverBumpsForZeroMajorVersion` was `false` by default. Then `true` by default with nx23, which meant if major version was still at 0.x.x, feature commits (`feat:...`) were only treated as patches. I had explicitly set it back to `false` so that our "feat" commits could still be considered as minor releases. But now that we're in v1.x.x, this setting is not of use anymore

PR #1341, opened 2026-09-15
fix(agentos): fix namespace member role changes failing with a 500

selimbensenouciep-prog:
## Symptom

Changing the role of an existing member on a namespace's members page failed with HTTP 500 (`PATCH /api/namespaces/{id}/members` with `[{"userId": "…", "role": "MEMBER"}]`). Adding a brand-new member worked, but every entry targeting a user already on the namespace (role change or removal) failed, and nothing was persisted.

## Root cause

`PermissionNodeNeo4jRepository.findRelationsForUsers` returned `u.id AS userId, type(r) AS relation` projected onto the `UserRelationRow` interface. Spring Data Neo4j materializes an interface projection through the repository's domain type (`UserNode`), which such a row cannot instantiate (`id` is null): every returned row raised a `MappingException`.

`Neo4jPermissionRepository.listRelationsForUsers` swallowed it into an empty map, so `NamespacePermissionServiceImpl.updateMembers` saw existing members as new users. A namespace admin got 403 ("Only a super-admin may add a new user") and a removal got 422, while a super-admin's update went through. The exception had, however, already crossed the repository's transactional proxy and marked the surrounding `@Transactional` `updateMembers` rollback-only, so the commit threw `UnexpectedRollbackException`, hence the 500.

Introduced with #1192: the query had no persistence test (the service specs mock it).

## Fix

**Core fix**
- `findRelationsForUsers` returns a single-column `collect({userId, relation})` as `List<Map<String, Any>>`, the pattern already used and documented in `CaseNodeNeo4jRepository.findDirectRelations`. `UserRelationRow` is deleted.
- `listRelationsForUsers` (repository and service) logs and rethrows instead of returning an empty map, which reads as "no current relation" and silently turns members into new users.

**Membership update path, reachable now that role changes commit** (from a pre-PR review)
- The lookup reports ADMIN for a user holding both edges (legacy data, see `PermissionRelation`) instead of whichever row comes last, and ignores soft-deleted users, whose edges outlive their deletion.
- Anti-lockout guard: soft-deleted users no longer count as remaining ADMINs, and the namespace ADMINs are only loaded when the update takes the role away from a current ADMIN without granting it to anyone.
- `updateMembers` first locks the namespace node (`NamespaceService.lockForUpdate`: a `SET`/`REMOVE` write undone in the same statement, `Propagation.MANDATORY`). Two crossed demotions now run one after the other, and the second one gets 422 instead of leaving the namespace without any ADMIN. #1192 deliberately kept the guard optimistic; that path could not commit until this fix.
- `PermissionServiceImpl` clears the permission cache immediately and again when the surrounding transaction completes, so a permission check running before the commit cannot re-cache the previous grant for the 5-minute TTL.
- A `null` entry in a membership PATCH body (`[null]`) returns 400 instead of a `NullPointerException` (500), for namespaces and cases (shared helper).

## Scope

- `agentos-service` only: permissions, namespace membership, membership request validation.
- No REST contract change, so no OpenAPI regeneration.

## Tests

- `EmbeddedNeo4jNamespaceMembersPersistenceSpec` (embedded Neo4j):
  - lookup per requested user, dual edges in both creation orders, soft-deleted users;
  - demotion by a super-admin and role change by a non-super-admin, both read back from the database;
  - 422 when the only other ADMIN is a soft-deleted user;
  - two crossed concurrent demotions: the second one gets 422 and one ADMIN remains.
- `NamespaceMembershipControllerPermissionIntegrationSpec` (MockMvc, real `@PreAuthorize` → permission evaluator → Neo4j chain): a namespace ADMIN and a super-admin each demote an ADMIN (200, persisted), a namespace MEMBER gets 403 with roles unchanged, `[null]` gets 400.
- `Neo4jPermissionRepositorySpec`, `PermissionServiceImplSpec`: lookup failures are rethrown; the cache is cleared again after transaction completion.
- `NamespacePermissionServiceImplSpec`: the ADMIN scan is skipped when the guard cannot trip; existing tests stub the lock.

## Validation

- Red before the fix: the first three persistence tests reproduced the production failures (`MappingException` then `UnexpectedRollbackException`, `AccessDeniedException` for a namespace admin). Each hardening test was red for its own reason: dual edge read as MEMBER, soft-deleted user counted, both crossed demotions committed, `NullPointerException` on `[null]`, cache cleared once, empty map instead of an exception.
- Green after the fix: 178 tests in 14 specs, 0 failures.

  ```bash
  cd agentos
  ./gradlew :agentos-service:test \
    --tests 'io.whozoss.agentos.persistence.neo4j.EmbeddedNeo4jNamespaceMembersPersistenceSpec' \
    --tests 'io.whozoss.agentos.namespace.NamespaceMembershipControllerPermissionIntegrationSpec' \
    --tests 'io.whozoss.agentos.namespace.NamespacePermissionServiceImplSpec' \
    --tests 'io.whozoss.agentos.permissions.PermissionServiceImplSpec' \
    --tests 'io.whozoss.agentos.permissions.Neo4jPermissionRepositorySpec'
    # plus NamespaceMembershipControllerSpec, NamespaceServiceImplSpec, NamespaceControllerMvcIntegrationSpec,
    # CaseMembershipControllerSpec, CaseMembershipMvcIntegrationSpec, EmbeddedNeo4jWatchesPersistenceSpec,
    # EmbeddedNeo4jUserGroupPersistenceSpec, EmbeddedNeo4jNamespacePersistenceSpec,
    # UserGroupControllerPermissionIntegrationSpec
  ```
- The concurrency test fails when the `lockForUpdate` call is removed, and passed in 4 runs with it.
- Not run locally: the full `agentos-service` suite (left to CI), and no manual check in the UI.
- `NamespacePermissionEndpointsMvcIntegrationSpec` fails 13 of 14 locally (403) on unmodified `master` as well; it is untouched here.

## Out of scope — follow-ups

- `NamespacePermissionEndpointsMvcIntegrationSpec` still assumes the permissive in-memory permission service removed in #838.
- Replacing `findRelationsForUsers` with the existing per-relation listings would also collapse dual edges, but trades the indexed lookup for an extra unindexed scan per PATCH; not done.
- The ADMIN listing and the batch role writes still match the entity without a label (`MATCH (e {id: $entityId})`).
- No shared fixture for the `createUser` / `createNamespace` helpers repeated across persistence specs (per-spec helpers are the current norm).

## Risks and rollback

- Relation lookup failures now surface as errors instead of being masked; soft-deleted users are ignored by membership updates; concurrent member updates of the same namespace are serialized (low-frequency admin operation).
- Rollback: revert the squash commit. No data migration, and the lock leaves no property on the node.


selimbensenouciep-prog:
Superseded by #1342 (narrower fix aligned on the group membership repository).

PR #1346, opened 2026-09-17
fix: #WZ-34801 Cascade agent soft-delete and disable to its scheduled prompts

frederic-delsert-whoz:
When an `AgentConfig` was soft-deleted or disabled, its associated `ScheduledPrompt`s were left orphaned — still active in the database, still picked up by the scheduler scanner, and failing at execution time.

A runtime guard already existed in `SchedulerScanner.claim()`: at each tick, for each due scheduler, the scanner checked whether the referenced agent still existed and was active. If not, it disabled the scheduler at that point. This was a late, reactive fix — the scheduler remained `enabled = true` in the database until the next tick, and a scheduler not yet due could stay active indefinitely.

### What was added

**Cascade on agent soft-delete** (`AgentConfigServiceImpl.delete()`)
- Immediately **soft-deletes** all `ScheduledPrompt`s referencing the deleted agent and their linked Prompt templates, within the same transaction
- Immediately **soft-deletes** all `Prompt`s directly linked to the agent (`agentConfigId` set)
- Rationale: `agentConfigId` is immutable after creation — a scheduler referencing a deleted agent cannot be rerouted and is permanently unusable. Soft-deleting preserves history while removing it from active views

**Cascade on agent disable** (`AgentConfigServiceImpl.disable()`)
- Immediately **disables** (`enabled = false`) all `ScheduledPrompt`s referencing the disabled agent
- Reversible: if the agent is re-enabled, `ScheduledPrompt`s can be manually re-enabled
- `enable(agent)` intentionally does nothing on `ScheduledPrompt`s — since `disable` is mechanical but we cannot distinguish a manual disable from a cascade disable, re-enabling is left as an explicit user action

**New repository methods**
- `PromptRepository.softDeleteByAgentConfigId(agentConfigId: UUID): Int` — bulk soft-delete of Prompts linked to an agent
- `ScheduledPromptRepository.disableByAgentConfigId(agentConfigId: UUID): Int` — bulk disable via a single Cypher `SET sp.enabled = false`
- `ScheduledPromptRepository.softDeleteWithPromptsByAgentConfigId(agentConfigId: UUID): Int` — bulk soft-delete of `ScheduledPrompt`s and their linked Prompt templates in a single atomic Cypher query. Uses `OPTIONAL MATCH` for the Prompt so the `ScheduledPrompt` is always soft-deleted even when its Prompt is already gone — a naive `MATCH…MATCH` would have inner-join semantics and silently skip orphaned `ScheduledPrompt`s
- Implemented in `Neo4jScheduledPromptRepository` and `InMemoryScheduledPromptRepository`

### What was already in place

- `SchedulerScanner.claim()` already disabled a `ScheduledPrompt` at tick time if its agent was missing or disabled — kept as a safety net for edge cases (race conditions, pre-existing inconsistent data, ...)
- `agentConfigId` was already immutable after creation (`toDomainForUpdate()` ignores the field on PUT)
- Soft-delete pattern (`removed = true`) and cascade pattern were already established — this PR follows the same conventions

### Tests

- `AgentConfigServiceImplUnitSpec`: four cases — soft-delete cascade on agent delete, disable cascade on agent disable, and isolation checks (other agents' schedulers unaffected) for both
- `AgentConfigServiceImplIntegrationSpec` (new, embedded Neo4j): cascade to Prompts, cascade to `ScheduledPrompt`s + their Prompt templates, orphaned `ScheduledPrompt` (OPTIONAL MATCH), disable/re-enable

---

### Guard: block deletion of a Prompt referenced by a ScheduledPrompt

Deleting a `Prompt` that is still referenced by a `ScheduledPrompt` (`removed = false`) left the scheduler with a broken `promptTemplateId` — at execution time, `resolveRunContext()` would fail with an `IllegalStateException`, marking the `UserRun` as FAILED. The `ScheduledPrompt` would remain active and keep generating failing runs indefinitely.

**What was added**

`PromptServiceImpl.delete()` now checks `ScheduledPromptRepository.existsActiveByPromptTemplateId(id)` before soft-deleting. If at least one non-removed `ScheduledPrompt` references the Prompt, the deletion is rejected with a `ConflictException` → **409 Conflict**.

The check does not consider `enabled` — a disabled `ScheduledPrompt` can be re-enabled at any time and would immediately inherit the broken reference.

**New repository method**
- `ScheduledPromptRepository.existsActiveByPromptTemplateId(promptTemplateId: UUID): Boolean` — single Cypher `count(sp) > 0`, no full node load
- Implemented in `Neo4jScheduledPromptRepository`, `ScheduledPromptNodeNeo4jRepository`, and `InMemoryScheduledPromptRepository`

**Tests**
- `PromptServiceImplIntegrationSpec` (embedded Neo4j): deletion blocked when referenced by active `ScheduledPrompt`; deletion succeeds when the referencing `ScheduledPrompt` is soft-deleted


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNodeNeo4jRepository.kt
```diff
@@ -148,4 +148,61 @@ interface ScheduledPromptNodeNeo4jRepository : Neo4jRepository<ScheduledPromptNo
...
+     */
+    @Query(
+        $$"""
+            MATCH (sp:ScheduledPrompt)
+            WHERE sp.promptTemplateId = $promptTemplateId
+              AND NOT COALESCE(sp.removed, false)
+            RETURN count(sp) > 0
+            """,
+    )
+    fun existsActiveByPromptTemplateId(promptTemplateId: String): Boolean
```
vincent-couturier-whoz, line 179: suggestion : Transactional readonly

vincent-couturier-whoz (dismissed):
Suggestion: if we have soft delete we could/should have restore in order to be able to recover soft deleted item which was removed by mistake.

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptNodeNeo4jRepository.kt
frederic-delsert-whoz: done at commit: 59e2b8ae4d92cac5dd64faf26cf265edc0608e48

vincent-couturier-whoz (approved):

PR #1373, opened 2026-09-23
fix: #WZ-34777 guard Case creation against users who lost agent access

frederic-delsert-whoz:
Between the materialisation of a `ScheduledPromptUserRun` (Phase A) and its execution (Phase B), a user may lose access to the agent — group deleted, deployment removed, etc. Without a guard, a Case was created for that user regardless.

## Change

Before creating the Case, `ScheduledPromptExecutor` now checks that the user is still reachable via the deployment graph using `AgentConfigService.findDeployedByNamespaceIdAndUserIdAndName` — the same path as the interactive `@mention` flow.

If access has been revoked, the `UserRun` is marked `DONE` (no error, no Case) and execution stops cleanly.

## Semantics

The check is symmetric with the interactive flow: a user can only receive a scheduled prompt if they would also be able to invoke the agent manually via `@mention`.


agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
```diff
@@ -361,6 +361,14 @@ class ScheduledPromptExecutor(
         // check for an existing Case before creating one, making execution effectively-once.
         try {
             val runContext = resolveRunContext(userRun)
+            if (!hasAgentAccess(userRun.userId, runContext.agentName, runContext.namespaceId)) {
+                logger.info {
+                    "[Executor] UserRun=${userRun.id} — user=${userRun.userId} no longer has READ access" +
+                        " to AgentConfig=${runContext.agentConfigId}, marking DONE without creating a Case"
+                }
+                userRunRepository.markTerminal(userRun.id, UserRunStatus.DONE, Instant.now(clock))
```
thomas-martin-whoz: it should be FAILED here right ?

thomas-martin-whoz (approved):

agentos/agentos-service/src/main/kotlin/io/whozoss/agentos/scheduledPrompt/ScheduledPromptExecutor.kt
frederic-delsert-whoz: good point on the naming `DONE` is indeed misleading here since no work was actually done.

That said, marking it `FAILED` would cause the parent `Run` to be flagged as failed too (the scanner derives the `Run` outcome from its `UserRun` statuses), which misrepresents the situation, the user simply lost deployment access in the meantime, which is an expected race, not an error.

A dedicated `SKIPPED`/`REJECTED` status would express the intent more clearly.
WDYT ?

frederic-delsert-whoz: Merging as-is, but agreed, we'll revisit if the information loss becomes a problem.

